PR #14, opened 2015-02-09
Made Evaluator a trait

alexarchambault:
Hi,
I'm thinking about plugging some other evaluators on your repl. For that, this PR tries to abstract away the `Evaluator` API. Would you be ok with keeping a distinct `Evaluator` trait like this? Ideally, with no too low level methods on it, like `evalClass` and the like.

### DIFF
--- a/repl/src/main/scala/ammonite/repl/Repl.scala
+++ b/repl/src/main/scala/ammonite/repl/Repl.scala
@@ -3,8 +3,8 @@ package ammonite.repl
 import java.io.{OutputStream, InputStream}
 import java.net.URLClassLoader
 import ammonite.IvyThing
-import ammonite.repl.eval.{Classpath, Evaluator, Preprocessor, Compiler}
-import ammonite.repl.frontend.{DefaultReplAPI, ReplAPIHolder, ReplAPI}
+import ammonite.repl.eval._
+import ammonite.repl.frontend.{DefaultReplAPI, ReplAPI}
 import acyclic.file
 import org.apache.ivy.Ivy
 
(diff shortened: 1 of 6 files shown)


lihaoyi:
Looks reasonable to me! Comments:

- `sbt "~repl/test-only -- ammonite.repl.EvaluatorTests"` fails with your PR merged into master; it looks like the test-`Checker` now has color output which it doesn't expect
- No need to put `private` everywhere. `val eval: Evaluator = new ScalaEvaluator` should be enough encapsulation for now, all `private` does is add noise. The public interface is already fully defined in `Evaluator`.

lihaoyi:
Also, do you want to talk about what "other evaluators" means for you? It would be helpful to know so we don't work cross purposes. I have my own ideas for what "other evaluators" means =P

alexarchambault:
I made some bigger changes (abstracting `Compiler` and `Preprocessor`, and using them from `Evaluator`), any aspect of it can be discussed.

My goal would be to plug kernels from my project Jove, https://github.com/jove-sh/jove (major refactoring  and modularization since last public commit, although it should work as is - I'll update its repo and make a first public announcement about it very soon). Hopefully, the spark interpreter from it should work straight away.

lihaoyi:
Cool. I'm probably going to have to put off reviewing until tonight when I get back from work.

lihaoyi:
Quick question: Why do you put the `Compiler` and `Preprocessor` into the `Evaluator`? I had very intentionally kept them separate and only cross-injected the things I needed

lihaoyi:
FWIW Splitting out the compiler, preprocessor and evaluator traits are uncontroversial and I can merge that separately if you want

lihaoyi:
I think I know why you wanted to stuff everything into `Evaluator`, and I think https://github.com/lihaoyi/Ammonite/commit/9e0c1ae3fdd9334910aa5c2649cbd700a96f659e should satisfy some of those desires

alexarchambault:
Thanks, I'll have a closer look at the last changes, and will actually try to plug in my own interpreters one way or another, later today.  Let's put this PR on hold.

lihaoyi:
Closing this for now

PR #29, opened 2015-03-06
Java 7 compatibility

alexarchambault:
Would you be ok to be compatible with Java 7? Some people (me included) still have some machines around with it. The changes required are minor.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -10,6 +10,7 @@ val sharedSettings = Seq(
   version := "0.2.4",
   libraryDependencies += "com.lihaoyi" %% "utest" % "0.3.0" % "test",
   testFrameworks += new TestFramework("utest.runner.Framework"),
+  scalacOptions += "-target:jvm-1.7",
   autoCompilerPlugins := true,
   addCompilerPlugin("com.lihaoyi" %% "acyclic" % "0.1.2"),
   libraryDependencies ++= Seq(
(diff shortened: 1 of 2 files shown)


lihaoyi:
Looks good to me. I'm running Java 8 though, so will likely end up breaking things in future accidentally, unless you have CI or something set up

alexarchambault:
Here it is:
https://github.com/lihaoyi/Ammonite/pull/30

Kwestor:
Setting up a Travis build for both 7 and 8 should be trivial.

You do use sbt, so something like this should suffice:

```yaml
language: scala
scala:
  - "2.11.6"
jdk:
  - openjdk7
  - oraclejdk8
```

lihaoyi:
Ok merges the travis CI stuff. Anything else I need to do?

Kwestor:
Yep, go https://travis-ci.org/lihaoyi/Ammonite, log with GitHub and add Ammonite to your builds.

lihaoyi:
Ok, i flipped a switch that looks like it does what we want
screenshot 2015-03-06 19 00 01

alexarchambault:
Yep, these were the right things to do.

alexarchambault:
Actually, you should have done them *before* merging for a build to be launched automatically. I think you should be able to launch one manually though.

PR #31, opened 2015-03-07
Early Scala 2.10 support

alexarchambault:
This PR adds support for Scala 2.10, if it's something you'd like to have in Ammonite.

Don't merge now !!!
Some tests don't pass, for various reasons (some Ivy dependencies not available in 2.10, ...), but I would see clearer in them if they all would pass in 2.11 :P

Basic repl interaction is ok though (results display, autocompletion, Ivy deps, ...)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 language: scala
 scala:
 - 2.11.4
+- 2.10.5
 jdk:
 - oraclejdk7
 - oraclejdk8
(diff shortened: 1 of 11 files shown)


lihaoyi:
Everything works in 2.10.5 when run using `test-only`; `test` fails with weird errors I haven't figured out. @alexarchambault FYI since you said not to merge it =P

alexarchambault:
Cool, I must have the issues I was thinking of then :)

Issue #15, opened 2015-02-14
Exception when trying to use Ammonite REPL


jchayat:
```scala
// build.sbt

name := "test-repl"

version := "1.0"

scalaVersion := "2.11.5"

libraryDependencies += "com.lihaoyi" %% "ammonite" % "0.1.6"

libraryDependencies += "com.lihaoyi" %% "ammonite-repl" % "0.1.6"

initialCommands in console := "ammonite.repl.Repl.main(null)"
```

```bash
#Command Line

jon$ sbt console
[info] Loading project definition from /Users/jon/code/test-repl/project
[info] Set current project to test-repl (in build file:/Users/jon/code/test-repl/)
[info] Starting scala interpreter...
[info] 
java.lang.AbstractMethodError
  at scala.tools.nsc.backend.JavaPlatform$class.$init$(JavaPlatform.scala:22)
  at ammonite.repl.interp.Compiler$$anon$1$$anon$6.<init>(Compiler.scala:70)
  at ammonite.repl.interp.Compiler$$anon$1.platform$lzycompute(Compiler.scala:70)
  at ammonite.repl.interp.Compiler$$anon$1.platform(Compiler.scala:70)
  at scala.tools.nsc.interactive.Global$$anon$8.<init>(Global.scala:401)
  at scala.tools.nsc.interactive.Global.loaders$lzycompute(Global.scala:399)
  at scala.tools.nsc.interactive.Global.loaders(Global.scala:399)
  at scala.tools.nsc.Global$GlobalMirror.rootLoader(Global.scala:65)
  at scala.reflect.internal.Mirrors$Roots$RootClass.<init>(Mirrors.scala:307)
  at scala.reflect.internal.Mirrors$Roots.RootClass$lzycompute(Mirrors.scala:321)
  at scala.reflect.internal.Mirrors$Roots.RootClass(Mirrors.scala:321)
  at scala.reflect.internal.Mirrors$Roots$EmptyPackageClass.<init>(Mirrors.scala:330)
  at scala.reflect.internal.Mirrors$Roots.EmptyPackageClass$lzycompute(Mirrors.scala:336)
  at scala.reflect.internal.Mirrors$Roots.EmptyPackageClass(Mirrors.scala:336)
  at scala.reflect.internal.Mirrors$Roots.EmptyPackageClass(Mirrors.scala:276)
  at scala.reflect.internal.Mirrors$RootsBase.init(Mirrors.scala:250)
[...]

PR #32, opened 2015-03-07
SBT plugin

alexarchambault:
Use it by adding
```scala
addSbtPlugin("com.lihaoyi" %% "ammonite-sbt" % "0.2.4")
```
to `project/plugins.sbt` or globally in `~/.sbt/0.13/plugins/build.sbt`.

At the sbt prompt of a sbt project, type `repl` and an Ammonite repl with the project dependencies and build products in the classpath will be launched.

Requires sbt >= 0.13.5 (for auto plugins).

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -4,19 +4,18 @@ publishArtifact := false
 
 publishTo := Some(Resolver.file("Unused transient repository", file("target/unusedrepo")))
 
-val sharedSettings = Seq(
-  scalaVersion := "2.11.5",
-  organization := "com.lihaoyi",
-  version := "0.2.4",
-  libraryDependencies += "com.lihaoyi" %% "utest" % "0.3.0" % "test",
-  testFrameworks += new TestFramework("utest.runner.Framework"),
-  scalacOptions += "-target:jvm-1.7",
+val acyclicSettings = Seq(
   autoCompilerPlugins := true,
   addCompilerPlugin("com.lihaoyi" %% "acyclic" % "0.1.2"),
   libraryDependencies ++= Seq(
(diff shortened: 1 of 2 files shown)


lihaoyi:
Question: What's the advantage of this over the current "add me to project" snippet? It seems like a whole lot of code that mostly does the same thing, and only really seems to save 1 LOC in a user's project if i'm not mistaken

zarthross:
Seems to me that this would allow a person to easily add this to their global sbt config so it will affect all projects.  I think it is also a bit more idiomatic SBT.

alexarchambault:
Same as @zarthross, you set up ammonite once and it's available in all your projects straight away. It's the same for updates: you have to change the version in one place, and you're done for all your projects.

You can also forget about the details of the one line to add in each project. You just have to remember to type `repl` at the sbt prompt, nothing more.

nightscape:
:+1:
@lihaoyi, do you see any problems with this?
I would really love to add Ammonite-REPL to my global plugins so that I can use it in all projects without hassle.

dwijnand:
:+1: Not only does this work well as a global plugin but I think I also agree with it's standalone config instead of using test.

alexarchambault:
The standalone project is here: https://github.com/alexarchambault/sbt-ammonite.

For now, it uses my cross-published version of Ammonite (https://github.com/alexarchambault/Ammonite), to tackle https://github.com/lihaoyi/Ammonite/issues/15.

alexarchambault:
Closing this as there's now https://github.com/alexarchambault/sbt-ammonite.

PR #38, opened 2015-03-10
More generic interpreter API

alexarchambault:
This PR contains the changes I had to make to re-use Ammonite's `Interpreter` to write a kernel for Jupyter / IPython (https://github.com/alexarchambault/jupyter-scala).

They mainly bring the ability to:
- use a different "bridge" object,
- use different wrapper code,
- wrap the code inside a class rather than an object (required for a Spark)

It also enhances the class loader of `Evaluator` by:
- making it provide the resources corresponding to the classes (resources in `*.class`), which requires having it...
- ...put the generated classes on disk upon demand (this is *on demand*, thus it incurs no cost compared to the current situation if the resources are not requested).

(These are for Spark too.)

These are pretty big changes to the internal API, anything can be discussed. Tests are ok.

### DIFF
--- a/repl/src/main/scala/ammonite/repl/Repl.scala
+++ b/repl/src/main/scala/ammonite/repl/Repl.scala
@@ -28,7 +28,7 @@ class Repl(input: InputStream,
     initialHistory
   )
 
-  val interp: Interpreter = new Interpreter(
+  val interp: Interpreter.Console = Interpreter.console(
     frontEnd.update,
     shellPrompt,
     pprintConfig.copy(maxWidth = frontEnd.width),
--- a/repl/src/main/scala/ammonite/repl/Util.scala
+++ b/repl/src/main/scala/ammonite/repl/Util.scala
@@ -75,8 +75,9 @@ case class Catching(handler: PartialFunction[Throwable, Res.Failing]) {
     try Res.Success(t(())) catch handler
 }
 
-case class Evaluated(wrapper: String,
-                     imports: Seq[ImportData])
+case class Evaluated[T](wrapper: String,
+                        imports: Seq[ImportData],
+                        value: T)
 
 case class ImportData(fromName: String,
                       toName: String,
--- a/repl/src/main/scala/ammonite/repl/frontend/JLineFrontend.scala
+++ b/repl/src/main/scala/ammonite/repl/frontend/JLineFrontend.scala
@@ -18,7 +18,7 @@ trait JLineFrontend{
    */
   def width: Int
   def action(buffered: String): Res[String]
-  def update(buffered: String, r: Res[Evaluated]): Unit
+  def update(buffered: String, r: Res[Evaluated[_]]): Unit
 }
 object JLineFrontend{
   def apply(input: InputStream,
@@ -78,7 +78,7 @@ object JLineFrontend{
 
     } yield buffered + res
 
-    def update(buffered: String, r: Res[Evaluated]) = r match{
+    def update(buffered: String, r: Res[Evaluated[_]]) = r match{
 
       case Res.Buffer(line) =>
         /**
--- a/repl/src/main/scala/ammonite/repl/frontend/ReplAPI.scala
+++ b/repl/src/main/scala/ammonite/repl/frontend/ReplAPI.scala
@@ -5,7 +5,7 @@ import java.io.File
 import scala.reflect.runtime.universe._
 import acyclic.file
 
-import scala.util.control.ControlThrowable
+import ammonite.repl.interp.Evaluator.Exit
 
 
 class ReplAPIHolder {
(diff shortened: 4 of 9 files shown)


alexarchambault:
Travis will complain about the tests, but this will be only because some tests assume that they are run from the `master` branch. (See https://travis-ci.org/alexarchambault/Ammonite/jobs/53846510)

lihaoyi:
Won't have time to look at this today but hopefully will get a chance tomorrow

alexarchambault:
Rethinking about this, it may be split in two: the changes for Scala proper first, and the ones for Spark then (in a different commit, or even PR, once I had put my Spark kernel for Jupyter online).

alexarchambault:
A word about the direction this could push Ammonite in my opinion: splitting ammonite-repl, in
- ammonite-compiler or ammonite-interpreter for the wrappers around nsc on the one hand, and 
- ammonite-repl for the repl proper on the other.

In particular, ammonite-compiler would not depend on ammonite-pprint or on JLine. Only ammonite-repl would.

lihaoyi:
This should probably be 4-5 different PRs; I'm not really sure what's happening =/ 

Some things (like making `Evaluated` have a `value: T`, or switching from object-wrappers to class-wrappers, should be entirely uncontroversial. 

Moving the interpreter and other non-interactive stuff into its own subproject is a separate issue, and also plausible.

Others, like making parametrizing Evaluator on its input or adding more constructor args to `Interpreter`, I don't understand. 

Rather than rebasing this huge thing, could you break it up into individual changes that can be properly discussed?

alexarchambault:
So I rebased and split the commits. (They don't appear in the right order here because of my various rebasing, see https://github.com/alexarchambault/Ammonite/commits/topic/generic for that.)

The first two ("More generic ...") make `Evaluator`, `Preprocessor`, and `Interpreter` more generic.

The other three allow for class-based wrappers. "Class based wrappers" allows for... class based wrappers. "Provide compiled classes..." makes the main `ClassLoader` behave more like JAR class loaders by providing the class definitions as resources.

I'll add tests here if these are ok for you.

lihaoyi:
I haven't looked at this yet but I promise i will

lihaoyi:
- aa40c1d Looks ok
- db77f8e I don't like. I would very much like Ammonite-REPL to not need a filesystem at all. Could we inject this in from outside the `Interpreter`?
- 4579196 You're going to have to explain a bit to me. Why is this useful? If it's useful, why make it optional instead of making everyone use it? Either way, controlling things by boolean flag seems pretty gross and probably will have to be replaced by e.g. injecting functionality via lambdas or a configuration object containing the logic
- b445ca2 What's the point of making the interpreter generic? I'd like to see at least two instantiations of that type to be convinced of its usefulness, even if it's just in the unit tests as "hey you can now do this too!"
- d90b949 Looks ok
- 3402fa9 Looks ok
- 648e62e and 4a9e85c Similarly need justification
- 63db038 Looks ok but needs justification

In general, I'm still missing the big picture why you want to genericize everything. Perhaps unit tests will help make this clear?

alexarchambault:
First, the overall goal is to re-use Ammonite's wrappers around scalac, whose main entry point is `Interpreter`, in other contexts. In particular, I'm doing this for a kernel for Jupyter / IPython 3, https://github.com/alexarchambault/jupyter-scala (mainly in https://github.com/alexarchambault/jupyter-scala/blob/master/kernel/src/main/scala/jupyter/scala/ScalaInterpreter.scala). Among other things, it needs a different repl wrapper as interaction is done though the IPython protocol rather than the console. As it doesn't use the printers from Ammonite, it also uses a slightly different "bridge" object. This is for https://github.com/alexarchambault/Ammonite/commit/63db038a6d941ce6b0c83b5a43fb3a4f444e6cef to https://github.com/alexarchambault/Ammonite/commit/b445ca21876a824f4f75f74ecf3687815009da99. I think this is pretty general and could be useful in other contexts, hence this PR.

I think I'm going to pull out https://github.com/alexarchambault/Ammonite/commit/457919679bac26c41c410e6f857804eeaad8e41c for now. Its main goal is to implement the same feature as the `-Yrepl-class-based` option of scalac in 2.11 (https://github.com/scala/scala/blob/2.11.x/src/compiler/scala/tools/nsc/settings/ScalaSettings.scala#L200). It would be used by the Spark counterpart of https://github.com/alexarchambault/jupyter-scala, that I'd like to test more before uploading it. The main goal of this kind of wrapper is to deal better with Java serialization (which Spark relies upon intensively), see for some explanations https://groups.google.com/forum/#!topic/scala-internals/h27CFLoJXjE. The understanding I have of it is to have the class holding the user definitions implement `Serializable`. This allows the user defined values themselves not to be recomputed when one de-serializes some values or closures that involve these values. It would have been interesting to have this feature in Ammonite (in particular, one would have it in scala 2.10, whereas scalac currently only has it in 2.11). I now feel that I'd like to question a bit the assumptions behind that (couldn't we just have the wrapper object extends `Serializable` instead of defining a side class proper?), hence I'll pull it out for now.

About https://github.com/lihaoyi/Ammonite/commit/db77f8e63d26067c2d78bb1416bdc5e3448b5c40, I can understand your criticism, I'll try to come up with something.

I'll also add unit tests to illustrate the more generic `Interpreter` (https://github.com/alexarchambault/jupyter-scala/blob/master/kernel/src/main/scala/jupyter/scala/ScalaInterpreter.scala is an example of use of it).

alexarchambault:
As briefly discussed on gitter, I'll maintain these on my fork. Feel free to ask if you think some of the changes in my fork could/should land in mainline.

PR #56, opened 2015-04-21
Add suport for predefs in REPL

ummels:
This is my stab at adding support for predefined commands that are automatically executed when running a REPL with a nonempty `predef` argument.

One possible drawback is that the predefined commands are not written to the history, but maybe this is even desired?

### DIFF
--- a/repl/src/main/scala/ammonite/repl/Repl.scala
+++ b/repl/src/main/scala/ammonite/repl/Repl.scala
@@ -37,21 +37,26 @@ class Repl(input: InputStream,
     initialHistory = initialHistory
   )
 
-  def action() = for{
+  def action(input: Res[String], save: Boolean) = for{
     // Condition to short circuit early if `interp` hasn't finished evaluating
-    line <- frontEnd.action(interp.buffered)
+    line <- input
     _ <- Signaller("INT") { interp.mainThread.stop() }
-    out <- interp.processLine(line, (f, x) => {saveHistory(x); f(x)}, _.foreach(print))
+    out <- interp.processLine(line, (f, x) => if (save) {saveHistory(x); f(x)}, _.foreach(print))
   } yield {
     println()
     out
   }
 
-
   def run() = {
-    @tailrec def loop(): Unit = {
-      val res = action()
+    @tailrec
+    def loop(): Unit = {
+      val res = action(frontEnd.action(interp.buffered), true)
       if (interp.handleOutput(res)) loop()
+    }
+
+    for (line <- predef.lines map (_.trim) if !line.isEmpty) {
+      val res = action(Res.Success(line), false)
+      interp.handleOutput(res)
     }
     loop()
   }
--- a/repl/src/main/scala/ammonite/repl/Repl.scala
+++ b/repl/src/main/scala/ammonite/repl/Repl.scala
@@ -48,15 +48,14 @@ class Repl(input: InputStream,
   }
 
   def run() = {
-    @tailrec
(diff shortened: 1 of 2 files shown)


repl/src/main/scala/ammonite/repl/Repl.scala
```diff
@@ -37,22 +37,27 @@ class Repl(input: InputStream,
...
   def run() = {
-    @tailrec def loop(): Unit = {
-      val res = action()
+    @tailrec
+    def loop(): Unit = {
+      val res = action(frontEnd.action(interp.buffered), true)
       if (interp.handleOutput(res)) loop()
     }
+
+    for (line <- predef.lines map (_.trim) if !line.isEmpty) {
```
lihaoyi: Could we move this above the `def loop` above, just to make it clear it's running first?

repl/src/main/scala/ammonite/repl/frontend/JLineFrontend.scala
```diff
@@ -107,7 +107,8 @@ object JLineFrontend{
 
       case Res.Success(ev) =>
         val last = reader.getHistory.size()-1
-        reader.getHistory.set(last, buffered + reader.getHistory.get(last))
+        if (last >= 0)
```
lihaoyi: What's this? I'm sure you had a good reason but it could use a comment!

lihaoyi:
How would someone make use of this ability when using ammonite from SBT? As an API?

Could we add a test for it somehow? I want the code-path to be executed otherwise who knows if it works or not =P

repl/src/main/scala/ammonite/repl/Repl.scala
ummels: Done.

repl/src/main/scala/ammonite/repl/frontend/JLineFrontend.scala
ummels: The problem is that this code is run on the first predefined command, which leads to an exception if the history is initially empty. Maybe a better approach would be to inhibit calling `frontend.update` when running the predefined commands? BTW, what is your take on adding the predefined commands to the history?

ummels:
I use this feature in my code by defining a custom `Repl` object that constructs an instance of `ammonite.repl.Repl` with a custom `predef`argument and then call this object's `main` method in my `build.sbt` file. Alternatively, one could use the first argument of the main method in `ammonite.repl.Repl` for the `predef` argument so that one could write

    val predef = ...
    initialCommands in (Test, console) := "ammonite.repl.Repl.main(Array(\"" + predef + "\"))"

Or do you have a better idea?

I will try to write a test in the coming days.

lihaoyi:
I patched down your diff and after some shuffling around committed it as https://github.com/lihaoyi/Ammonite/commit/d57ffac384f68c927b809c01b1f31c41d3ecb4c7

My current idea for SBT would be 

```scala
initialCommands in (Test, console) := s"""
ammonite.repl.REPL.run(\"\"\"
    ...
\"\"\")
"""
```

But it's sooooo ugly. Maybe it's time I went and took @alexarchambault's SBT plugin, make that obey `initialCommands`, and started publishing that too

ummels:
Cool, thanks. My only complaint is that the initial prompt appears directly after the predefined commands with no linebreak inbetween:

    Loading Ammonite Repl...
    x: Int = 1@

lihaoyi:
Ok cool, closing this for now

Issue #48, opened 2015-04-14
Truncate pretty-printed output so it doesn't take over the console


lihaoyi:
Currently, when something gets echoed in the console, it gets printed no matter how much output it produces, so for example this input:

```scala
@ Seq.fill(100)(100)
```

Leaves your terminal looking like this

```scala
  100,
  100,
  100,
  100,
  100,
  100,
  100,
  100,
  100,
  100,
  100,
  100,
  100,
  100,
  100,
  100,
  100,
  100,
  100,
  100,
  100,
  100
)
@
```

After it's done. It's even worse with data-structures like infinite streams (described in https://github.com/lihaoyi/Ammonite/issues/27) which will keep printing indefinitely.

We should control the output, such that after a certain amount of lines or characters has been printed, the rest is represented as `...`:

```scala
@ Seq.fill(100)(100)
res0: Seq[Int] = List(
  100,
  100,
  100,
  100,
  100,
  100,
  100,
  100,
  100,
  100,
  100,
  100,
  100,
  100,
...
@
```

That prevents your console from being taken over by accidentally echoing large data structures to stdout. Similarly, commands which result in lots of lines, but only after wrapping, should have their output truncated to fit within some number of rows. e.g.

```scala
@ Seq.fill(1000)(100).toString
```

currently leaves your terminal looking like

```scala
100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 
100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 
100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 
100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 
100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 
100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 
100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 
[...]

PR #57, opened 2015-04-21
Truncating long PPrint output

laci37:
This will fix issue #48. PPrinters created with PPrinter.apply are now truncating thier outputs. Values wrapped in the type Full[T] will print normally with all output.

### DIFF
--- a/pprint/src/main/scala/ammonite/pprint/Core.scala
+++ b/pprint/src/main/scala/ammonite/pprint/Core.scala
@@ -23,7 +23,7 @@ case class Config(maxWidth: Int = 100,
                   renames: Map[String, String] = Config.defaultRenames)
   extends GenConfig[Config]{
   def deeper = copy(depth = depth + 1)
-  def long = copy(maxHeight = 0)
+  def full = copy(maxHeight = 0)
   def rename(s: String) = renames.getOrElse(s, s)
   object color{
     def apply(s: String, c: String) = {
--- a/pprint/src/main/scala/ammonite/pprint/Core.scala
+++ b/pprint/src/main/scala/ammonite/pprint/Core.scala
@@ -15,7 +15,7 @@ import acyclic.file
  *                TupleN *
  */
 case class Config(maxWidth: Int = 100,
-                  maxHeight: Int = 15,
+                  maxHeight: Int = 0,
                   depth: Int = 0,
                   indent: Int = 2,
                   literalColor: String = null,
--- a/pprint/src/main/scala/ammonite/pprint/Core.scala
+++ b/pprint/src/main/scala/ammonite/pprint/Core.scala
@@ -13,6 +13,7 @@ import acyclic.file
  *                TupleN *
  */
 case class Config(maxWidth: Int = 100,
+                  maxHeight: Int = 15,
                   depth: Int = 0,
                   indent: Int = 2,
                   literalColor: String = null,
--- a/pprint/src/main/scala/ammonite/pprint/Core.scala
+++ b/pprint/src/main/scala/ammonite/pprint/Core.scala
@@ -7,6 +7,8 @@ import acyclic.file
  *
  * @param maxWidth Controls how far to the right a line will go before
  *                 it tries to wrap
+ * @param maxHeight Controls how many lines can be printed at once. 
+ *                  Will print all lines if set to 0
  * @param depth How much the current item being printed should be indented
  * @param renames A map used to rename things to more common names, e.g.
  *                renamig `WrappedArray` to `Array` or getting rid of
@@ -21,6 +23,7 @@ case class Config(maxWidth: Int = 100,
                   renames: Map[String, String] = Config.defaultRenames)
   extends GenConfig[Config]{
   def deeper = copy(depth = depth + 1)
+  def long = copy(maxHeight = 0)
   def rename(s: String) = renames.getOrElse(s, s)
   object color{
     def apply(s: String, c: String) = {
--- a/pprint/src/main/scala/ammonite/pprint/PPrint.scala
+++ b/pprint/src/main/scala/ammonite/pprint/PPrint.scala
@@ -2,6 +2,7 @@ package ammonite.pprint
 
 import scala.annotation.unchecked.uncheckedVariance
 import scala.language.experimental.macros
+import annotation.tailrec
 import acyclic.file
 
 
@@ -211,9 +212,22 @@ object Internals {
     val coloredName = c.color.prefix(renamed)
     val chunks2 = chunkFunc(c.deeper)
     val indent = "  " * c.depth
-    Iterator(coloredName + "(\n") ++
+    val full = Iterator(coloredName + "(\n") ++
     mkIterator(chunks2.map(Seq("  " + indent) ++ _), Seq(",\n")).flatten ++
     Iterator("\n" + indent + ")")
+    takeFirstLines(c.maxHeight,full)
+  }
+
+  def takeFirstLines(lines: Int, iter: Iterator[String]):Iterator[String]={
+    @tailrec
+    def inner(lines: Int, iter: Iterator[String], begin: Iterator[String]): Iterator[String]={
+      if(lines==0) return begin ++ Iterator("...")
+      if(!iter.hasNext || lines < 0) return begin
+      val head = iter.next()
+      val remainingLines = if(head.contains('\n')) lines-1 else lines
+      inner(remainingLines,iter,begin ++ Iterator(head))
+    }
+    inner(lines,iter,Iterator.empty)
   }
 
   def preMap[T, V: PPrint](f: T => V) = PPrinter[T] {
--- a/pprint/src/main/scala/ammonite/pprint/PPrint.scala
+++ b/pprint/src/main/scala/ammonite/pprint/PPrint.scala
@@ -33,13 +33,15 @@ object PPrint extends Internals.LowPriPPrint{
 case class PPrint[A](a: PPrinter[A], cfg: Config){
   def render(t: A): Iterator[String] = {
     if (t == null) Iterator("null")
-    else {
-      a.render(t, cfg)
-    }
+    else a.render(t, cfg)
   }
   def map(f: String => String) = a.map(f)
 }
 
+/**
+ * Wrapper type for disabling output truncation.
+ * PPrint(Full(value)) will always return the full output.
+ */
 case class Full[A](a: A)
 
 /**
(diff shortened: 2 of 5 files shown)


lihaoyi:
Thanks! I'll take a look tonight when I'm off work

pprint/src/main/scala/ammonite/pprint/PPrint.scala
```diff
@@ -118,6 +131,21 @@ object PPrinter {
...
   }
 
+  private def takeFirstLines(cfg: Config, iter: Iterator[String]):Iterator[String]={
+    @tailrec
+    def inner(lines: Int, iter: Iterator[String], begin: Iterator[String]): Iterator[String]={
+      if(!iter.hasNext || lines < 0) return begin
+      if(lines==0) return begin ++ Iterator(cfg.color.prefix("..."))
+      val head = iter.next()
+      if(lines < head.length/cfg.maxWidth){
+        return begin ++ Iterator(head.substring(0, lines*cfg.maxWidth), cfg.color.prefix("..."))
```
lihaoyi: Don't use `return` unless you really need to; e.g. here you can just have chained `if ... else if ... else if ... else`s

pprint/src/main/scala/ammonite/pprint/PPrint.scala
```diff
@@ -30,11 +33,15 @@ object PPrint extends Internals.LowPriPPrint{
...
     if (t == null) Iterator("null")
-    else a.render(t, cfg)
+    else {
+      a.render(t, cfg)
+    }
   }
   def map(f: String => String) = a.map(f)
 }
 
+case class Full[A](a: A)
```
lihaoyi: A short doccomment saying what this is would be great =)

pprint/src/main/scala/ammonite/pprint/PPrint.scala
```diff
@@ -30,11 +33,15 @@ object PPrint extends Internals.LowPriPPrint{
 case class PPrint[A](a: PPrinter[A], cfg: Config){
   def render(t: A): Iterator[String] = {
     if (t == null) Iterator("null")
-    else a.render(t, cfg)
```
lihaoyi: As a matter of style, I'd leave this on the same line because it's short enough (< 100 chars)

pprint/src/main/scala/ammonite/pprint/PPrint.scala
```diff
@@ -126,6 +154,15 @@ object PPrinter {
   implicit def SeqRepr[T: PPrint] = Internals.collectionRepr[T, Seq[T]]
   implicit def SetRepr[T: PPrint] = Internals.collectionRepr[T, Set[T]]
   implicit def MapRepr[T: PPrint, V: PPrint] = Internals.makeMapRepr[Map, T, V]
+  
+  implicit def fullPPrinter[A: PPrint]: PPrinter[Full[A]] = {
+    new PPrinter[Full[A]]{ 
+      def render(wrapper: Full[A], c: Config) = {
+        val pprint = implicitly[PPrint[A]]
+        Iterator("asd") ++  pprint.copy(cfg = c.full).render(wrapper.a)
```
lihaoyi: What's `asd`?

pprint/src/main/scala/ammonite/pprint/Core.scala
```diff
@@ -7,12 +7,15 @@ import acyclic.file
...
  *                 it tries to wrap
+ * @param maxHeight Controls how many lines can be printed at once. 
+ *                  Will print all lines if set to 0
  * @param depth How much the current item being printed should be indented
  * @param renames A map used to rename things to more common names, e.g.
  *                renamig `WrappedArray` to `Array` or getting rid of
  *                TupleN *
  */
 case class Config(maxWidth: Int = 100,
+                  maxHeight: Int = 15,
```
lihaoyi: I'd make this `0`, so by default it'll print everything, and set it to `15` only for the config used in the REPL. In theory someone using this for non-REPL stuff such as logging would want the whole output

lihaoyi:
Could we add an alias for the REPL? I'd rather type

```scala
@ full(Seq.fill(100)("10" * 100))
```

Than 

```scala
@ ammonite.pprint.Full(Seq.fill(100)("10" * 100))
```

Truncation seems to be working great, but the `Full` implicit doesn't seem to get picked up:

```scala
@ ammonite.pprint.Full(Seq.fill(100)("10" * 100))
res9: ammonite.pprint.Full[Seq[String]] = Full(
  List(
    "10101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010",
    "10101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010",
    "10101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010",
    "10101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010",
    "10101010101010101010101010101010101010101010101010101010101010101010101010101010101010101010101...
@
```

It seems to be falling back to the macro. I'm not sure why, but it doesn't seem to work.

Lastly, we should have tests for all the cases (long-non-truncated, short-truncated, short-lines-truncated, one-long-line-truncated, ...) inside the `pprint/` module. That way we can be certain ^_^

lihaoyi:
You said you were having problems with one of the tests. Could you show me where?

pprint/src/main/scala/ammonite/pprint/PPrint.scala
laci37: Trace for seeing whether this code really executes. Shouldn't have gotten into version control though.

laci37:
I have made all suggested changes except the REPL alias, I'll look into that tomorrow.

lihaoyi:
Don't forget to write tests!

laci37:
I have made the changes you suggested. The failing test is repl.EulerTests.p13 and it fails with

[info]                  utest.AssertionError: failLoudly(assert(processed.isInstanceOf[Res.Buffer]))
[info]                  processed: ammonite.repl.Res[ammonite.repl.Evaluated] = Success(Evaluated(cmd0,List(ImportData(s,s,cmd0,cmd0))))

pprint/src/test/scala/ammonite/pprint/PPrintTests.scala
```diff
@@ -378,5 +378,127 @@ object PPrintTests extends TestSuite{
...
+      }
+
+      'long_line_truncated{
+        implicit val cfg = Config.Defaults.PPrintConfig.copy(
+          maxWidth = 100,
+          maxHeight = 1
+        )
+        check(
+          "a" * 1000,
+          """"aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa..."""
```
lihaoyi: Could we add one more test for `maxWidth` > 1, `maxHeight` > 1, but with a single-line of text that wraps several times before wrapping?

repl/src/test/scala/ammonite/repl/EulerTests.scala
```diff
@@ -317,8 +317,6 @@ object EulerTests extends TestSuite{
                    |53503534226472524250874054075591789781264330331690""".stripMargin.replace("\n", " ")
       check.session(s"""
         @ val s = "$data"
-        s: java.lang.String = "$data"
```
lihaoyi: I think you need to leave a blank line here for the `check.session` function to treat these as separate commands. Otherwise it expects the first command to be incomplete (`Buffer`)

lihaoyi:
Could we add two more tests? One for infinite streams:

```scala
Stream.continually("foobarbaz")
```

To verify that we do not run out of memory materializing the whole thing. One for the "wrap once" case, e.g.

```scala
// maxHeight = 6
// maxWidth = 10

input: 
012345678912
012345678912
012345678912
012345678912
output:
012345679
12
012345679
12
012345679
12
...
```
To verify that the line-wrapping math is correctly treating left-over parts of a line as a full line when calculating the number of lines used

lihaoyi:
That's all I think. Thanks for putting up with me ^_^

pprint/src/main/scala/ammonite/pprint/PPrint.scala
```diff
@@ -118,6 +133,24 @@ object PPrinter {
     s.toString()
   }
 
+  private def takeFirstLines(cfg: Config, iter: Iterator[String]):Iterator[String]={
+    @tailrec
+    def inner(lines: Int, iter: Iterator[String], begin: Iterator[String]): Iterator[String]={
+      if(!iter.hasNext || lines < 0)  begin
```
lihaoyi: Can we just `assert lines > 0` and skip these checks?

laci37:
Last test showed that the line wrapping math was wrong. Now I think it shouldn't be fooled by any kind of input. Two tests (p13 and p27) in EulerTests are still failing.

lihaoyi:
I pushed a bad version, try rebasing on master so tests can run?

laci37:
I rebased to lihaoyi/Ammonite master. EulerTests.p13 fails with Failure('utest.AssertionError: failLoudly(assert(processed.isInstanceOf[Res.Buffer])).

pprint/src/test/scala/ammonite/pprint/PPrintTests.scala
```diff
@@ -378,5 +378,153 @@ object PPrintTests extends TestSuite{
...
+    }
+
+    'wrapped_lines{
+      implicit val cfg = Config.Defaults.PPrintConfig.copy(
+        maxWidth = 8,
+        maxHeight = 5
+      )
+      check(
+        "1234567890\n"*10,
+        "\"\"\"\n1234567890\n1234567890\n..."
```
lihaoyi: Shouldn't this be
```
"\"\"\"\n12345678\n90\n12345678\n..."
```

Since the `maxWidth` is only 8, but there are 10 characters in each line of the input, and it cuts off after 5 lines

lihaoyi: Eh nevermind, I'm dumb. We should not do the wrapping!

lihaoyi:
@laci37 what's the status of this? I'm not sure whether you're just waiting for me to merge, or whether you're still fiddling with it ^_^ If you're waiting on me let me know!

laci37:
I'm waiting for the merge. The only issue remaining is the failing test, but you can probably solve that in minutes.

lihaoyi:
Ok, sorry! Will merge/fix when I have time.

Issue #54, opened 2015-04-20
Unknown Resolver spam is super annoying


lihaoyi:
Seems to pop up the first time you download things from the internet. We're probably doing something wrong but I don't know what
```scala
Loading Ammonite Repl...
@ load.ivy("com.github.seratch" %% "awscala" % "0.5.2")
:: loading settings :: url = jar:file:/Users/haoyi/.ivy2/cache/org.apache.ivy/ivy/jars/ivy-2.4.0.jar!/org/apache/ivy/core/settings/ivysettings.xml
:: resolving dependencies :: com.github.seratch#awscala_2.11-caller;working
	confs: [default]
	found com.github.seratch#awscala_2.11;0.5.2 in central
	found org.scala-lang#scala-library;2.11.6 in central
	[2.11.6] org.scala-lang#scala-library;2.11.6
	found com.amazonaws#aws-java-sdk-iam;1.9.31 in central
	found com.amazonaws#aws-java-sdk-core;1.9.31 in central
	found commons-logging#commons-logging;1.1.3 in central
	found org.apache.httpcomponents#httpclient;4.3.4 in central
	found org.apache.httpcomponents#httpcore;4.3.2 in central
	found commons-codec#commons-codec;1.6 in central
	found com.fasterxml.jackson.core#jackson-databind;2.3.2 in central
	found com.fasterxml.jackson.core#jackson-annotations;2.3.0 in central
	found com.fasterxml.jackson.core#jackson-core;2.3.2 in central
	found com.amazonaws#aws-java-sdk-sts;1.9.31 in central
	found com.amazonaws#aws-java-sdk-ec2;1.9.31 in central
	found com.amazonaws#aws-java-sdk-s3;1.9.31 in central
	found com.amazonaws#aws-java-sdk-kms;1.9.31 in central
	found com.amazonaws#aws-java-sdk-sqs;1.9.31 in central
	found com.amazonaws#aws-java-sdk-emr;1.9.31 in central
	found com.amazonaws#aws-java-sdk-redshift;1.9.31 in central
	found com.amazonaws#aws-java-sdk-dynamodb;1.9.31 in central
	found com.amazonaws#aws-java-sdk-simpledb;1.9.31 in central
	found joda-time#joda-time;2.7 in central
	found org.joda#joda-convert;1.7 in central
	found com.decodified#scala-ssh_2.11;0.7.0 in central
	found net.schmizz#sshj;0.10.0 in central
[...]

PR #65, opened 2015-04-22
Less verbose Ivy

alexarchambault:
This almost shuts Ivy, except in case of error.

It can be made more verbose again by increasing `maxLevel` here:
https://github.com/lihaoyi/Ammonite/commit/fe609776aa4dd33df161dbb09457a7fce3e2cb5b#diff-0c45674fa431da1de377ef0cd36e5af7R35
but at `maxLevel = 2`, it's actually quite similar to what it currently is. There seems to be no middle ground (except by manually filtering the most annoying messages :-)

Fixes https://github.com/lihaoyi/Ammonite/issues/54


alexarchambault:
Don't know how to test that, your `Checker` seems to be fine no matter what the log is.

lihaoyi:
Yeah, because Ivy prints straight to STDOUT, so we don't catch it at all =/

alexarchambault:
Setting your own stdout/stderr can be done, with `System.setOut` / `System.setErr`.

PR #66, opened 2015-04-22
Allowing to add Ivy resolvers

alexarchambault:
This PR demonstrates a way to add one's own repositories from the repl, like
```scala
@ load.resolver("Scalaz Bintray Repo" at "https://dl.bintray.com/scalaz/releases")

@ load.resolver(Resolver.sonatypeRepo("snapshots"))
```

About the implementation, as calling several times `addResolver` on the ivy settings (like here: https://github.com/lihaoyi/Ammonite/blob/master/repl/src/main/scala/ammonite/repl/IvyThing.scala#L49) seems not to work, it re-uses some code from SBT regarding "chained resolvers" (mixes several repo together), which I don't have a deep understanding of :-), but it works for me (and in the tests).

An alternate way of doing that could be to depend on the sbt artifacts themselves - cut-n-pasted like I did seems less worse to me...

Yet another way *could* consist in replacing Ivy by aether, which has a slicker API (I never used aether directly though, only heard of it).

### DIFF
--- a/repl/src/main/scala/ammonite/repl/IvyThing.scala
+++ b/repl/src/main/scala/ammonite/repl/IvyThing.scala
@@ -25,6 +25,10 @@ trait IvyConstructor{
   implicit class ArtifactIdExt(t: (String, String)){
     def %(version: String) = (t._1, t._2, version)
   }
+  implicit class ResolverNameExt(name: String){
+    def at(location: String) = Resolver.mavenResolver(name, location)
+  }
+  val Resolver = ammonite.repl.Resolver
 }
 
 /**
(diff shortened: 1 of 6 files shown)


lihaoyi:
Looks like a huge load of code to paste into the repo; this would literally increase the size of the repl source code by 50%! Do we really need all of it? What are the use cases which called for this? 

If you could provide unit tests that illustrate all the use cases this supports then we can discuss it further. I bet we don't actually need 600 lines of code to implement the bits we care about.

alexarchambault:
I can understand your point about the size. The use case was being able to add snapshot dependencies through `load.ivy`, like
```scala
@ load.ivy("com.github.alexarchambault" %% "argonaut-shapeless_6.1" % "0.1.2-SHAPSHOT")
```
Hard to test, as snapshot artifacts are transient - they disappear soon after releases (but there was a test with scalaz-stream, which needs a resolver from bintray).

The added code could either be aggressively stripped/refactored, or aether could be a solution. I may try that at some point, closing this for now.

Issue #72, opened 2015-04-25
Extensible Tab Completion


rcano:
A key feature of any shell, is extensible completion by the input commands. For example, without the ability to tab-complete pathes, it becomes mandatory to launch a second "native" shell in order to work. In fact, one of the reasons people migrate to zsh is because of its superior completion-hooks support.

Here is a possible idea:

I believe that a simple trait could provide a hook into the system. For example (disregard names):
```scala
triat CompletionHook {
  def suggestions: Seq[String] //there should be a better type here than string.
}
```
and then say you implement that trait in BasePath, then, when the repl reads:

    root / "home" / <TAB>
it would ideally detect the prefix `root / "home"`, typecheck that to be of type BasePath which extends CompletionHook, and then invoke `(root / "home").suggestions`, which would in turn provide suggestions under the home directory

The pros about this solution is that its very simple to enhance your types with it in your repl-libs, as well as already having proper prefix since its an instance method. The cons is that the repl would have to evaluate that code which could go awfully given side effects. Although this is a problem of any shell, and the hook implementor is responsible for doing good code I guess.

An alternative approach would be some form of companion based dispatch like `BasePath.suggestions(linePrefix)` but I'm afraid that kind of interface would be hard to use, since the parsing (of whatever form) of the prefix would have to be done by every hook implementor.

rcano:
I've managed to implement this functionality here https://github.com/rcano/Ammonite/tree/completion-hook
I tried hard to make the change as reduced as possible, though I don't like a lot the way that certain parts need to be passed among them (I needed a reference to the evaluator from the Pressy in order to implement this), but at least shows a working concept of the feature.
I also ended up not needed a base trait, and worked with any type whose shape defined a suggestions method, and as an example I implemented it for Path (only absolute ones).

I'm very concerned with the amount of classes loaded with *each* suggestion made, I think the `newFileDict` should somehow be turned into a weak reference map, and see how class unloading can be leveraged, otherwise, the repl will hog more and more memory, eventually dying (I think this is too an issue with normal input, but suggestions aggravate it).

lihaoyi:
Interesting! Will take a look

lihaoyi:
Do you want to clean up your diff (e.g. remove all the silly whitespace changes that distract from the point) and send in a PR? I can't guarantee that I'll merge it as is, but it's much easier to talk that way.

rcano:
[...]

PR #74, opened 2015-04-25
Completion hook

rcano:
Added extensible completion suggestions to the REPL.

See issue: https://github.com/lihaoyi/Ammonite/issues/72

What this PR misses is a custom evaluator that discards the generated code after the suggestions are obtained.


rcano:
I screwed up and left an incorrect modification to the build.sbt. If you merge this, please correct that.

sschaef:
There are compilation errors in this PR:
```
[error] Unwanted cyclic dependency
[info] 
[info] /home/travis/build/lihaoyi/Ammonite/ops/src/main/scala/ammonite/ops/package.scala:63: 
[info]           val completionsToLoad = ls! root / 'usr / 'share / "bash-completion" / 'completions |? (_.name contains cmdName) | (f => s"source $f")
[info]                                   ^
[info] symbol: object ls
[info] 
[info] /home/travis/build/lihaoyi/Ammonite/ops/src/main/scala/ammonite/ops/FileOps.scala:132: 
[info]     FilterMapExt(ls.rec! from) | copyOne
[info]     ^
[info] symbol: package object ops
```

repl/src/main/scala/ammonite/repl/interp/Pressy.scala
```diff
@@ -97,15 +104,56 @@ object Pressy {
...
+                  try {
+                    val suggestions = cls.getMethod("suggestions").invoke(null).asInstanceOf[Seq[String]]
+                    Some(index -> suggestions.map(s => s -> None).toList)
+                  } catch { case ce: ClassCastException => None }
+                case _ => None
+              }
+          }.headOption.flatten
+        }
+        suggestions getOrElse (index -> ask(index, pressy.askScopeCompletion).map(s => (s.sym.name.decoded, None)))
+      }
```
sschaef: This block of code is more or less unreadable. Bind blocks to defs and vals and give them descriptive names, don't do this function chaining when it consumes more than one line of code.

ops/src/main/scala/ammonite/ops/package.scala
```diff
@@ -41,4 +41,46 @@ package object ops extends Extensions with RelPathStuff{
 
 
   implicit def fileData(p: Path) = stat.full(p)
+  
+  private val bashCompletionFunction = """__print_completions() { for ((i=0;i<${#COMPREPLY[*]};i++)); do echo ${COMPREPLY[i]}; done }"""
+  implicit class BashInterpolator(val sc: StringContext) {
```
sschaef: How is one supposed to call this bash completion functionality?

repl/src/main/scala/ammonite/repl/interp/Pressy.scala
```diff
@@ -18,6 +19,7 @@ trait Pressy{
 }
 object Pressy {
 
+  private val suggestionNumber = new java.util.concurrent.atomic.AtomicInteger
```
sschaef: Scaladoc for these sort of things would be nice

repl/src/main/scala/ammonite/repl/interp/Pressy.scala
```diff
@@ -97,15 +104,56 @@ object Pressy {
...
+        }.sortBy(index - _._1.pos.end).headOption.flatMap {
+          case (operationTarget, operator, argument) =>
+            val completionInvocation = new String(operationTarget.pos.source.content.slice(operationTarget.pos.start, operationTarget.pos.end))
+            val completionIncovationArguments = Seq('"' + operator.decoded + '"', argument.map(a => new String(a.pos.source.content.slice(a.pos.start, a.pos.end))).getOrElse("()")).mkString(", ")
+            val wrapperName = "SuggesterWrapper" + eval.getCurrentLine + suggestionNumber.incrementAndGet
+            val code = s"""$previousImports
+                        |object $wrapperName {
+                        |def suggestions = ($completionInvocation).suggestions($completionIncovationArguments)
+                        |}""".stripMargin
+//            println(s"($completionInvocation).suggestions($completionIncovationArguments)")
```
sschaef: It is not good style to commit println and it is even worse to commit them as comments.

sschaef:
In #72 you talked about adding a completion hook but I can't see one. How is one supposed to implement their own completion hooks?

rcano:
@sschaef 
What I implemented here follows the idea exposed in the issue, that if a type - say Path - provides a method called suggestions which has a specific form (that must be documented), the repl will make a call to that before resuming the standard scopes completion.
So your custom types wouldn't need to extend any particular trait or anything.
I have diverged a bit from what the issue says, because I experimented a lot while writing this, giving more power to the completion hook. Maybe this PR should be taken out.

ops/src/main/scala/ammonite/ops/package.scala
rcano: Well, documentation pending but: in this branch string interpolators support completion so that when you type, for example: sh"some text<TAB>", the REPL tries to invoke StringContext(texts).sh.suggestions(args).
So in this case, if you do for example: sh"git branch <TAB>" it would autocomplete with the branches, or sh"git l<TAB>" yields log.

repl/src/main/scala/ammonite/repl/interp/Pressy.scala
rcano: Absolutely.

repl/src/main/scala/ammonite/repl/interp/Pressy.scala
rcano: I don't know if you are watching the diff in github with its very thin view, it does look good to me, and every line is a val declaration, except for the match and a try catch, there isn't any long chain of calls..
Anyway, I don't mind if someone wants to reformat it, I don't know the style you like so I wouldn't want to do it to avoid needless cycles of "make that look like that". Please reformat it to your heart content.

repl/src/main/scala/ammonite/repl/interp/Pressy.scala
rcano: noted.

lihaoyi:
to set the context, I'm not going to land this as is, so no need to go over too much detail or coding style =P I was more thinking of the how-this-works and the whether-this-will-work-at-all kind of review.

I just wanted this to be here to give us a place to talk about it without having to dig into @rcano's repo every time I wanted to find the diff, and without the comments being lost to ancient history as he makes changes

lihaoyi:
This has been superseded by https://github.com/lihaoyi/Ammonite/commit/b477a423bd2214adf9d69ce350b4924dba3963f6 and the surrounding work

Issue #55, opened 2015-04-20
Running code in object-initializers makes things slow


lihaoyi:
Apparently code directly inside an `object Foo{ ... }` block doesn't get JITed, and the commands that you enter into Ammonite all are run that way. This seems to result in a 10x slowdown in the rare-ish case you're doing intensive work directly in that block.

```scala
@ (() => {val start = System.currentTimeMillis(); var i = 1L; while(i < 1000000000L) i += (i % 2) + 1; System.currentTimeMillis() - start})()
res13: Long = 1065L
@ {val start = System.currentTimeMillis(); var i = 1L; while(i < 1000000000L) i += (i % 2) + 1; System.currentTimeMillis() - start}
res14: Long = 11464L
```

A solution would be to lift all `var`/`val`/`lazy val`/`<expr>` statements automatically into `(() => ...)()` in the `Preprocessor`.

som-snytt:
This is a common problem in typesafe repl, and the reason for `-Yrepl-class-based` or whatever.

lihaoyi:
Fixed, thanks @laci37!

som-snytt:
There's another take on scala/scala to use App, where DelayedInit solves the init problem (which is similar to this fix).

https://github.com/scala/scala/pull/4349

PR #85, opened 2015-05-15
Lifting expressions to anonymous function application

laci37:
This solves #55. All REPL tests are passing. Should I write some tests or benchmarks?

### DIFF
--- a/repl/src/main/scala/ammonite/repl/interp/Preprocessor.scala
+++ b/repl/src/main/scala/ammonite/repl/interp/Preprocessor.scala
@@ -52,8 +52,24 @@ object Preprocessor{
     val TypeDef = DefProc("type"){ case m: G#TypeDef => m.name }
 
     val PatVarDef = Processor { case (name, code, t: G#ValDef) =>
+      //Function to lift lhs expressions into anonymous functions so they will be JITed
+      def wrap(code: String)={
+        import fastparse._
+        import scalaparse.Scala._
+        val par = P( ( `implicit`.? ~ `lazy`.? ~ ( `var` | `val` ) ~! BindPattern.rep1("," ~! Pass) ~ (`:` ~! Type).?).! ~ (`=` ~! StatCtx.Expr.!) )
+        val Result.Success((lhs, rhs), _) = par.parse(code)
+        //Rebuilding definition from parsed data to lift rhs to anon function
+        s"$lhs = { () =>\n $rhs \n}.apply"
+      }
+
       Output(
-        code,
+        //Only wrap rhs in function if it is not a function
+        //Wrapping functions causes type inference errors.
+        t.rhs match {
+          case _: G#Function => code //simple anon function
+          case _: G#Match => code   //anon partial function
+          case _ => wrap(code)
+        },
         // Try to leave out all synthetics; we don't actually have proper
         // synthetic flags right now, because we're dumb-parsing it and not putting
         // it through a full compilation
@@ -70,7 +86,8 @@ object Preprocessor{
     }
 
     val Expr = Processor{
-      case (name, code, tree) => Output(s"val $name = (\n$code\n)", Seq(pprint(name)))
+      //Expressions are lifted to anon function applications so they will be JITed
+      case (name, code, tree) => Output(s"val $name = { () =>\n$code\n}.apply", Seq(pprint(name)))
     }
 
     val decls = Seq[(String, String, G#Tree) => Option[Preprocessor.Output]](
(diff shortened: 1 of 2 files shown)


repl/src/main/scala/ammonite/repl/interp/Preprocessor.scala
```diff
@@ -52,8 +52,30 @@ object Preprocessor{
...
     val PatVarDef = Processor { case (name, code, t: G#ValDef) =>
+      //Function to lift expressions into anonymous functions so they will be JITed
+      def wrap(code: String)={
+        import fastparse._
+        import scalaparse.Scala._
+        val par = P( ( `implicit`.? ~ `lazy`.? ~ ( `var` | `val` ) ~! BindPattern.rep1("," ~! Pass) ~ (`:` ~! Type).?).! ~ (`=` ~! StatCtx.Expr.!) )
+        val Result.Success((lhs, rhs), _) =
+          par.parse(code).asInstanceOf[Result.Success[(String, String)]]
+        //Rebuilding definition from parsed data to lift rhs to anon function
+        val bdr = new StringBuilder
```
lihaoyi: Could we collapse this into a single string interpolation call? `s"$lhs = { () =>\n $rhs \n}.apply"` would make it a lot more concise ^_^

repl/src/main/scala/ammonite/repl/interp/Preprocessor.scala
```diff
@@ -52,8 +52,30 @@ object Preprocessor{
     val TypeDef = DefProc("type"){ case m: G#TypeDef => m.name }
 
     val PatVarDef = Processor { case (name, code, t: G#ValDef) =>
+      //Function to lift expressions into anonymous functions so they will be JITed
+      def wrap(code: String)={
+        import fastparse._
+        import scalaparse.Scala._
+        val par = P( ( `implicit`.? ~ `lazy`.? ~ ( `var` | `val` ) ~! BindPattern.rep1("," ~! Pass) ~ (`:` ~! Type).?).! ~ (`=` ~! StatCtx.Expr.!) )
+        val Result.Success((lhs, rhs), _) =
+          par.parse(code).asInstanceOf[Result.Success[(String, String)]]
```
lihaoyi: No need `asInstanceOf` here: the pattern match should already fail if its a failure

repl/src/main/scala/ammonite/repl/interp/Preprocessor.scala
```diff
@@ -70,7 +92,7 @@ object Preprocessor{
     }
 
     val Expr = Processor{
-      case (name, code, tree) => Output(s"val $name = (\n$code\n)", Seq(pprint(name)))
+      case (name, code, tree) => Output(s"val $name = { () =>\n$code\n}.apply", Seq(pprint(name)))
```
lihaoyi: Wait, if we already do the wrapping above, do we need this here too?

lihaoyi: Nm, I'm an idiot, this is for wrapping expressions. Could you leave a similar comment as the one you left at the `PatVarDef` definition?

laci37: Aren't Expr and PatVarDef separate, so only one of them processes a single line of code?

lihaoyi:
- Could you also just try reproducing the performance degradation in the linked issue before and after your patch, to verify manually that the perf problem has gone away? No need benchmarks; the existing tests passing should be sufficient. 
- Could you squash your commits and force push?

That is all

laci37:
That's all for now then.

repl/src/main/scala/ammonite/repl/interp/Preprocessor.scala
```diff
@@ -52,8 +52,24 @@ object Preprocessor{
...
+        val par = P( ( `implicit`.? ~ `lazy`.? ~ ( `var` | `val` ) ~! BindPattern.rep1("," ~! Pass) ~ (`:` ~! Type).?).! ~ (`=` ~! StatCtx.Expr.!) )
+        val Result.Success((lhs, rhs), _) = par.parse(code)
+        //Rebuilding definition from parsed data to lift rhs to anon function
+        s"$lhs = { () =>\n $rhs \n}.apply"
+      }
+
       Output(
-        code,
+        //Only wrap rhs in function if it is not a function
+        //Wrapping functions causes type inference errors.
```
lihaoyi: Can you give an example of a type inference error this will cause if we leave it out? I'm wondering if it's possible to cause problems in cases other than just function and partial function literals

laci37: This kind of lines:
```scala
val fun: T1 => T2 = { x => f(x) }
val pfun: T1 => T2 = { case x => f(x) }
```
So function and partial function literals with the type specified on the left hand side. The error message said, that the function parameter's type could not be determined.
`val f = {x: Int => 2*x}` compiled fine with the lifting.

lihaoyi:
Looks good, I didn't manage to find a way to make the type inference fail. If someone comes up with a case in the future (not unlikely!) we can sort it out then. Thanks!

PR #92, opened 2015-06-01
Bugfix in SyntaxError

laci37:
SyntaxError.msg throws an `IndexOutOfBoundsException` if the failing position is at the end of the last line.

### DIFF
--- a/repl/src/main/scala/ammonite/repl/Util.scala
+++ b/repl/src/main/scala/ammonite/repl/Util.scala
@@ -184,7 +184,7 @@ object SyntaxError{
   def msg(code: String, p: fastparse.core.Parser[_], idx: Int) = {
     val locationString = {
       val (first, last) = code.splitAt(idx)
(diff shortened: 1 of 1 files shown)


repl/src/main/scala/ammonite/repl/Util.scala
```diff
@@ -184,7 +184,7 @@ object SyntaxError{
   def msg(code: String, p: fastparse.core.Parser[_], idx: Int) = {
     val locationString = {
       val (first, last) = code.splitAt(idx)
-      val lastSnippet = last.split('\n')(0)
+      val lastSnippet = last.split('\n').headOption getOrElse ""
```
lihaoyi: can we use `.getOrElse("")` here? I tend to not use infix notation except for cases where we're chaining lots of things together

lihaoyi:
Thanks for fixing! Juts one nit

Issue #101, opened 2015-06-07
Consolidate persistent data in a sensible way


lihaoyi:
We currently have the `~/.amm` file, which contains the command history in a crazy ad-hoc syntax. Soon we'll be adding cached bytecode (https://github.com/lihaoyi/Ammonite/issues/95), and cached dependency resolution (https://github.com/lihaoyi/Ammonite/issues/100) and no doubt more things in the future.

We need to figure out a more structured way of doing this (make `~./amm` JSON? YAML? A folder with flat files? SQLite?) that won't become a huge mess in future.

laci37:
I'd say make it a `~/.ammonite` folder. I would write out the full name so it is easy to see which application created the folder. Having a directory structure inside won't make a mess I think. Something like this:
```
.ammonite
  cache
    ivy
      jarfile.jar
      jarfile2.jar
    scripts
      block1.class
      block2.class
  history
  user_settings
```
We can can store caches in single files too. If we use a directory structure backwards compatibility wouldn't be hard to solve either.

lihaoyi:
Sounds reasonable. Ivy jars I don't think we should copy; we just need to cache the path to the file since they'll all be cached in the `~/.ivy` folder anyway, maybe with a hash. If the jar disappears, or doesn't match the hash, we can ask Ivy to re-download it.

I think the other plain-text-code data can live in YAML files, because their [block notation](http://en.wikipedia.org/wiki/YAML#Newlines_preserved) would make the text both very easy to human-read (no escaping!), very easy to human/machine edit, and very easy to machine-append (so we don't need to re-write the entire history file or ivy-resolution file each time we add a new entry).

lihaoyi:
@laci37 do you want to pick this up as well and take a first pass at it? You're going to need it as part of https://github.com/lihaoyi/Ammonite/issues/95 anyway to put your class-files somewhere

laci37:
Okay, I'll get started.

lihaoyi:
Thanks to @laci37 we now have a `~/.ammonite/` folder which currently contains:

- `history`
- `ivycache.json`
- `predef.scala`

And lots of room to add more stuff. Even though we don't yet have proper concurrency semantics, I'm going to call this closed

Issue #99, opened 2015-06-07
Reverse order of history


lihaoyi:
We're currently showing it oldest-first, which is problematic:

- The newer things, which you probably care about the most, get truncated when you print it
- To chop off some segment of the newer things to look at, you need to do awkward `history.drop(history.length - 100)`

We could replace the data structure and pprint and other things, but a simpler solution would just to store history in reverse order, or at least treat it as a reverse-ordered-thing when printing/indexing/slicing

lihaoyi:
On second thoughts, having a custom datastructure that prints last-parts-first and uses @laci37's script syntax would be pretty cool in allowing people to C&P parts straight from the terminal, while still being a real `Seq[String]` and thus programmatically accessible. That would require *not* reversing it, though. Requires more thought

lihaoyi:
Closing for now since @laci37 has implemented a custom pprint and it seems to work great. Thanks @laci37 !

PR #108, opened 2015-06-14
History YAML storage and pretty printing

laci37:
This addresses issues #99 and #101. A new class `History` and a new singleton `Storage` is added. `History` has a `PPrinter` that prints it in script syntax. Script syntax was modified, so now not just `@\n` but anything matching the regex `@[\n ]+` is accepted as compilation block separator. `History` also has a `last(lines: Int)` method. `Storage` stores history in the `~/.ammonite/history` file with YAML syntax.

### DIFF
--- a/repl/src/main/scala/ammonite/repl/Storage.scala
+++ b/repl/src/main/scala/ammonite/repl/Storage.scala
@@ -45,7 +45,7 @@ object Storage{
 }
 
 class History extends ArrayBuffer[String]{
-  def last(lines: Int){
+  def last(lines: Int) = {
     drop(length - lines)
   }
 }
--- a/repl/src/main/scala/ammonite/repl/Storage.scala
+++ b/repl/src/main/scala/ammonite/repl/Storage.scala
@@ -5,6 +5,7 @@ import java.io.{File, FileInputStream, IOException, FileWriter}
 import org.yaml.snakeyaml.Yaml
 import scala.collection.mutable.ArrayBuffer
 import scala.collection.JavaConversions.asScalaBuffer
+import ammonite.pprint
 
 trait Storage{
   
@@ -56,4 +57,16 @@ class History extends ArrayBuffer[String]{
   }
 }
 
+object History{
+  import pprint._
+  implicit def historyPPrint(implicit c: Config): PPrint[History] = new PPrint(
+    new PPrinter[History]{
+      def render(t: History, c: Config)={
+        val seq = "\n...\n" +: t.last(c.lines()).flatMap{ code => Seq("@ ", code, "\n") }
+        seq.iterator
+      }
+    },
+    c
+  )
+}
 
--- a/repl/src/main/scala/ammonite/repl/Storage.scala
+++ b/repl/src/main/scala/ammonite/repl/Storage.scala
@@ -62,8 +62,10 @@ object History{
   implicit def historyPPrint(implicit c: Config): PPrint[History] = new PPrint(
     new PPrinter[History]{
       def render(t: History, c: Config)={
-        val seq = "\n...\n" +: t.last(c.lines()).flatMap{ code => Seq("@ ", code, "\n") }
-        seq.iterator
+        val lines = if(c.lines() > 0) c.lines() else t.length
+        val seq = "\n" +: t.last(lines).flatMap{ code => Seq("@ ", code, "\n") }
+        if(t.length > lines) ("\n..." +: seq).iterator
+        else seq.iterator
       }
     },
     c
--- a/repl/src/main/scala/ammonite/repl/Storage.scala
+++ b/repl/src/main/scala/ammonite/repl/Storage.scala
@@ -38,7 +38,7 @@ object Storage{
   def saveHistory(h: History): Unit = {
     val fw = new FileWriter(dir + "/history")
     yaml.synchronized{
-      yaml.dump(h.toArray,fw)
+      yaml.dump(h.toArray, fw)
     }
   }
 
--- a/repl/src/main/scala/ammonite/repl/Storage.scala
+++ b/repl/src/main/scala/ammonite/repl/Storage.scala
@@ -6,38 +6,44 @@ import org.yaml.snakeyaml.Yaml
 import scala.collection.mutable.ArrayBuffer
 import scala.collection.JavaConversions.asScalaBuffer
 
+trait Storage{
+  
+  def loadHistory: History
+
+  def saveHistory(h: History): Unit
+}
+
 object Storage{
-  val dir = new java.io.File(System.getProperty("user.home") + "/.ammonite")    
-  if(dir.exists){
-    if(!dir.isDirectory){
-      dir.delete()
+  val defaultPath = new java.io.File(System.getProperty("user.home") + "/.ammonite") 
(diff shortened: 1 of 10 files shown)


laci37:
There are a few failing tests, I'm going to deal with them tomorrow.

repl/src/main/scala/ammonite/repl/Storage.scala
```diff
@@ -0,0 +1,53 @@
...
+  if(dir.exists){
+    if(!dir.isDirectory){
+      dir.delete()
+      dir.mkdir()
+    }
+  } else {
+    dir.mkdir()
+  }
+
+  lazy val yaml = new Yaml
```
lihaoyi: Is initializing this expensive? If not let's create a new one in every `loadHistory` and `saveHistory` and save ourselves the messiness of caching + lazy-init + locking

repl/src/main/scala/ammonite/repl/Storage.scala
```diff
@@ -0,0 +1,53 @@
+package ammonite.repl
+
+import acyclic.file
+import java.io.{File, FileInputStream, IOException, FileWriter}
+import org.yaml.snakeyaml.Yaml
+import scala.collection.mutable.ArrayBuffer
+import scala.collection.JavaConversions.asScalaBuffer
+
+object Storage{
```
lihaoyi: Could we turn this into a class parametrized on the folder path? That feels to me like it could be useful, e.g. if I'm running scripts on a shared machine and want everyone to share the same history/caching

repl/src/main/scala/ammonite/repl/Storage.scala
```diff
@@ -0,0 +1,53 @@
...
+        case a: java.util.List[String] => res ++= a
+        case _ =>
+      }
+      res
+    } catch {
+      case e: IOException => new History
+    }
+  }
+
+  def saveHistory(h: History): Unit = {
```
lihaoyi: How long does dumping the whole histo ry every time take? Picking some arbitrary metric like 1000 lines of history 100 characters each? 10000 lines? If it turns out to be slow we should think about this more (e.g. append to the file instead of re-writing the whole thing). In either case, the rough benchmarks should live in a comment in the source code

repl/src/main/scala/ammonite/repl/Storage.scala
```diff
@@ -0,0 +1,53 @@
...
+      val list = yaml.synchronized{
+        yaml.load(new FileInputStream(dir + "/history"))
+      }
+      list match {
+        case a: java.util.List[String] => res ++= a
+        case _ =>
+      }
+      res
+    } catch {
+      case e: IOException => new History
```
lihaoyi: Only the initialization of `val list` can throw an exception, so the catch block should only go around that.

lihaoyi:
Now that we're splitting out `Storage` into its own subsystem, I think we should define an explicit `trait Storage` as the interface, and have the `Storage` companion object nest multiple implementations of this trait. 

Apart from the file-based one you've already defined (parametrized on the path to the storage folder), I'd want an in-memory implementation for cases where the REPL process can't write to the filesystem (not unreasonable, e.g. if you're running ammonite-REPL inside some docker container, security-manager or on Heroku) or for unit testing. 

If we share e.g. the yaml serialization/etc. code between the in-memory and folder-based versions, we could ensure that the file formats and reading/writing behave as expected in unit tests without needing to have all the tests touching the filesystem

lihaoyi:
As an example of this trait + implementations pattern, take a look at the `FrontEnd` trait

repl/src/main/scala/ammonite/repl/interp/Interpreter.scala
```diff
@@ -135,8 +135,16 @@ class Interpreter(shellPrompt0: Ref[String],
     def search(target: scala.reflect.runtime.universe.Type) = Interpreter.this.compiler.search(target)
     def compiler = Interpreter.this.compiler.compiler
     def newCompiler() = init()
-    def history = history0.toVector.dropRight(1)
+    def history = history0
     def show[T](a: T, lines: Int = 0) = ammonite.pprint.Show(a, lines)
+
+    implicit val historyPPrinter: pprint.PPrinter[History] =
```
lihaoyi: A unit test for this would be great. Doesn't need to be an entire `check.session` test, just call `PPrint(...).mkString` and check that it returns what you'd expect.

lihaoyi: I'd just chuck it in `UnitTests` with the other random stuff, we can split it out into its own file if it gets too big/complicated

lihaoyi:
What are the concurrency semantics for the file-based approach? Is it "undefined if multiple people are writing at the same time", e.g. if there are multiple REPL processes open on the same machine? 

If we want this to be a reasonable replacement for Bash (I do =D), it'd need some reasonable semantics (e.g. folder-locking). Even if we don't fix it now, a comment in the code mentioning the current semantics would be nice

lihaoyi:
That's all for now

repl/src/main/scala/ammonite/repl/Storage.scala
laci37: I ran a quick benchmark and this is about 5 - 20 us (minus overhead that I added in the benchmark), so it seems fast enough.

repl/src/main/scala/ammonite/repl/Storage.scala
laci37: This one seems to be mostly linear wrt file size.I tested with short lines, 1000 lines gave 1.3ms, 10000 lines:  14 ms 100000 lines: 168ms. I benchmarked the whole `saveHistory` function so YAML serialization is included. I'd say it is acceptable but we will need to make it append to the file some time.

repl/src/main/scala/ammonite/repl/Storage.scala
laci37: The pattern match can probably throw an exception too because of type erasure `list` can't be checked if it contains strings.

repl/src/main/scala/ammonite/repl/interp/Interpreter.scala
laci37: Now I see that this instance doesn't belong in ReplAPI, it should be in the History companion.

laci37: The is not get picked up in the companion (it should be according to [stackoverflow](http://stackoverflow.com/questions/5598085/where-does-scala-look-for-implicits)). Where should I put it? It is a bit hard to access from `UnitTests` if it's in `ReplApi`.

lihaoyi: How do the other special `PPrint`s, e.g. for the `CommandResult` pprint? Presumably we could do the same here

laci37:
I did most changes that you wanted. The history file is rewritten each time, I think it is not a huge performance hit with <10000 lines of history, but I will look into it how to append YAML documents with snakeyaml. There are no proper semantics for concurrency in `Storage`, but I need to do some research on that, Java's `FileLock`'s docs say that how locking works is highly platform dependent.

lihaoyi:
Looks great. Thanks!

PR #110, opened 2015-06-26
Switching to coursier

alexarchambault:
Not necessarily for immediate merge...

Replaces Ivy by [coursier](https://github.com/alexarchambault/coursier), and illustrates how coursier works. Minor glitches remain in its API, I'll fix them soon. Messages during the resolution could be better (progress bars, colors, etc.). This uses the artifact downloader from coursier, but alternatives can be used too.


alexarchambault:
(I'm looking at the travis failure...)

lihaoyi:
This looks great, way faster than ivy even without any caching on our part. What else is missing before `0.1.0`?

alexarchambault:
I'd like to add:
- small enhancements to the API (make the cache policy implicit - I forgot to use it for the metadata here, because of that I guess),
- basic checksum verifying / file locking / error handling,
- and minor other things too (see https://github.com/alexarchambault/coursier/issues?q=is%3Aopen+is%3Aissue+milestone%3A0.1.0).

But the milestone is already usable.

alexarchambault:
Closing this, code now largely deprecated.

PR #114, opened 2015-07-16
Checker regex

laci37:
This PR makes `Checker` in REPL test hadnle the expected output as a vim nomagic mode regex. This way we can handle more possible right outputs with ease. Regex special characters in expected output are handled as literals unless they are prefixed with `\`. Literal `\` itself needs to be escaped.


lihaoyi:
Looks good, thanks!

Issue #95, opened 2015-06-07
Cache script loading for performance


lihaoyi:
A common usecase for scripts is that they're run over and over. If they don't change, then we shouldn't need to re-compile them.

As discussed with @laci37, this would look something like

- Caching will be persistent somehow
- We'll cache the compilation blocks of loaded scripts separately under top-level cache-tags each
- The cache tags be a hash of classpath, previous commands (in-same-script), and the block-code
- Loaded scripts would not have access to loading environment, for encapsulation and ease-of-caching
- We'd implement this by limiting the imports provided inside load.file, and replace the auto-increment wrapper-names with cache-tag-names for uniqueness
- This will mean scripts will be cached more often as we can be 100% sure that non-classpath-changing commands (e.g. 2+2) run before the script will not affect its compilation

lihaoyi:
@laci37 here's another question for you: we've already agreed that scripts will not be able to grab things from the caller's dynamic scope. What about the inverse? If I perform an import in a script, will it be visible to the loader after I finish loading it?

The downside of doing this is that it allows more leakage. The downside of *not* is that without this, there's basically no way to abstract over imports; a lot of people will probably want to load a "predef" script that imports all the stuff they care about.

A possible compromise could be to only allow definitions and imports from the *last* CompilationUnit of a script to be visible to the outside world. That would allow the end-user to customize encapsulation/exporting, and only export the things he wants to export.

What if you don't want to export *everything* in the last compilation unit? Well there's always labelling things `private`. That doesn't work now but could work with some minor tweaks to the parser which would make it do exactly what you'd expect

laci37:
This would be simple to do. However I'd prefer making all compilation blocks behave the same way and do not export anything by default. I have to think about how to do this.

lihaoyi:
Yeah one problem is that Scala doesn't currently have an idea of exports; everything is implicitly public unless turned private =/ If it did the same thing as C# and made public-ness opt-in then we could just use that.

Also, annotations don't work because you can't annotate imports =(

lihaoyi:
Removing this from 0.4.0, since this will probably take some time to land, it's been a while since the last release, and I don't want to delay it. It can get on the next train

lihaoyi:
Actually it seems reasonably soon, so I'll wait

lihaoyi:
Closed by #115

PR #115, opened 2015-07-18
Cache

laci37:
Scripts loaded with `load.module` are cached to avoid recompiling them. Sideeffects in these scripts only happen once in each session. These scripts are limited to importing `ReplApi` stuff from the REPL, nothing else defined in the REPL is accessible.


repl/src/main/scala/ammonite/repl/Storage.scala
```diff
@@ -20,6 +20,8 @@ trait Storage{
 
   val history: Ref[History]
   val ivyCache: Ref[IvyMap]
+  def compileCacheSave(tag: String, classFiles: Traversable[(String, Array[Byte])], imports: Seq[ImportData]): Unit
+  def compileCacheLoad(tag: String): Option[(Traversable[(String, Array[Byte])], Seq[ImportData])]
```
lihaoyi: Could we break out the `Traversable[(String, Array[Byte])]` into a type alias? We use it all over the place so having a short name for it would be nice

repl/src/main/scala/ammonite/repl/Storage.scala
```diff
@@ -57,6 +58,57 @@ object Storage{
       }
     }
 
+    def compileCacheSave(tag: String, classFiles: Traversable[(String, Array[Byte])], imports: Seq[ImportData]): Unit = {
+      val cacheDir = new File(dir + s"/compileCache/$tag")
+      if(!cacheDir.exists){
+        cacheDir.mkdirs()
+        val metadata = upickle.default.write(imports)
+        val metadataFw = new FileWriter(cacheDir + "/metadata.json")
```
lihaoyi: Since you're adding a second case of this, let's break out all this `FileWriter` logic into a helper function so `ivyCache` below can re-use it

lihaoyi: I'd make the helper do all the `FileOutputStream` stuff for us, and just call `toBytes` if we need to write a string

repl/src/main/scala/ammonite/repl/Storage.scala
```diff
@@ -57,6 +58,57 @@ object Storage{
...
+          fos.flush()
+        }
+      }
+    }
+
+    def compileCacheLoad(tag: String): Option[(Traversable[(String, Array[Byte])], Seq[ImportData])] = {
+      val cacheDir = new File(dir + s"/compileCache/$tag")
+      if(!cacheDir.exists) None
+      else for{
+        metadataJson <- try{
```
lihaoyi: We seem to be doing this "try, catch exception, convert to Option" thing a lot; at least three times in this code. Could we try breaking that out into a `catching[T, E](t: T): E` helper?

repl/src/main/scala/ammonite/repl/Storage.scala
```diff
@@ -57,6 +58,57 @@ object Storage{
...
+        }
+        classFiles <- loadClassFiles(cacheDir)
+      } yield (classFiles, metadata)
+    }
+
+    def loadClassFiles(cacheDir: File): Option[Traversable[(String, Array[Byte])]] = {
+      val classFiles = cacheDir.listFiles().filter(_.getName.endsWith(".class"))
+      try{
+        val data = classFiles.map{ file =>
+          val fis = new java.io.FileInputStream(file)
```
lihaoyi: Same as output, I think we've reached the point where this should probably go in a helper function and we should refactor `ivyCache`/`loadPredef` to use that and call `new String(...)` on the byte array

repl/src/main/scala/ammonite/repl/frontend/FrontEnd.scala
```diff
@@ -169,7 +169,7 @@ object FrontEnd{
 
       val term = makeTerm()
       term.init()
-      val reader = new ConsoleReader(???, output, term)
+      val reader = new ConsoleReader(System.in, output, term)
```
lihaoyi: Let's make this take the `input` passed in above, rather than grabbing stdin

repl/src/main/scala/ammonite/repl/interp/Evaluator.scala
```diff
@@ -206,6 +222,74 @@ object Evaluator{
...
+      )
+      
+      _ = cacheSave(wrapperName,(classFiles,importData))
+      cls <- loadClass(wrapperName,classFiles) 
+    } yield (cls, importData)
+
+    def processScriptBlock(code: String, scriptImports: Seq[ImportData]) = {
+      val wrapperName = cacheTag(code, scriptImports)
+      loadCachedClass(wrapperName) match {
+        case Some((cls,importData)) => {
```
lihaoyi: No need curlies after each `case`

repl/src/main/scala/ammonite/repl/interp/Evaluator.scala
```diff
@@ -206,6 +222,74 @@ object Evaluator{
       )
     }
 
+    def compileCacheBlock(wrapperName: String, code: String, scriptImports: Seq[ImportData]): Res[(Class[_], Seq[ImportData])] = for {
+      _ <- Catching{ case e: ThreadDeath => interrupted() }
+      (classFiles, importData) <- compileClass(
+        s"""
```
lihaoyi: Could we extract this and the similar code above in `processLine` into a common function to make sure it stays in sync?

repl/src/main/scala/ammonite/repl/interp/Evaluator.scala
```diff
@@ -206,6 +222,74 @@ object Evaluator{
...
+
+    def processScriptBlock(code: String, scriptImports: Seq[ImportData]) = {
+      val wrapperName = cacheTag(code, scriptImports)
+      loadCachedClass(wrapperName) match {
+        case Some((cls,importData)) => {
+          evalMain(cls)
+          Res.Success(
+            Evaluated(
+              wrapperName,
+              importData.map(id => id.copy(
```
lihaoyi: Let's pull out all three callsites of this pattern into a shared function

repl/src/main/scala/ammonite/repl/interp/Interpreter.scala
```diff
@@ -77,6 +78,31 @@ class Interpreter(shellPrompt0: Ref[String],
       }
     }
   }
+  
+  def processModule(code: String): Unit = {
```
lihaoyi: This fella has a whole lot of duplicate code with `processScript` above; let's pull it out

lihaoyi:
Since you asked about testing, we can look at what the original "spec" (https://github.com/lihaoyi/Ammonite/issues/95) was to determine what we should test. Firstly, we'd need some way to verify that we're actually hitting a cached copy of a script; you'll have to figure out how to do that. But once we have that...

> Caching will be persistent somehow

Could we spawn multiple `Repl`s one after the other, point them to the same `ammoniteHome`, and make sure that the second one loads from cache and doesn't recompile?

> We'll cache the compilation blocks of loaded scripts separately under top-level cache-tags each

Can we load a script with 1, 2, and 3 blocks and verify that 1, 2 and 3 separate cache entries gets generated respectively?

> The cache tags be a hash of classpath, previous commands (in-same-script), and the block-code

Can we load almost-the-same script with different previous-commands, different block-code, and different classpath hash, and check that we get three different cache entries?

> Loaded scripts would not have access to loading environment, for encapsulation and ease-of-caching

This seems easy to test. We should be able to load a script that would pass if it has access to caller-scope definitions, but fails because it can't access them. And we can verify it failing

> We'd implement this by limiting the imports provided inside load.file, and replace the auto-increment wrapper-names with cache-tag-names for uniqueness

We could assert that loading stuff from cache doesn't result in "normally" named `cmd0`, `cmd1`, `cmd2`... top-level-objects being created, e.g. via reflection

> This will mean scripts will be cached more often as we can be 100% sure that non-classpath-changing commands (e.g. 2+2) run before the script will not affect its compilation

We can load a module, run `2+2`, and load it again, and make sure it is only compiled once

lihaoyi:
One thing that is not in that spec is the fact that "load.module scripts would only be loaded once, and side effects would only happen once, while load.exec scripts could execute over and over". We can easily have a script with side effects to verify both of these facts.

repl/src/main/scala/ammonite/repl/interp/Evaluator.scala
```diff
@@ -206,6 +222,74 @@ object Evaluator{
...
+                wrapperName = wrapperName,
+                prefix = if (id.prefix == "") wrapperName else id.prefix
+              ))
+            )
+          }
+        }
+      }
+    }
+
+    def cacheTag(code: String, imports: Seq[ImportData]): String = {
```
lihaoyi: Shouldn't the cacheTag also depend on the hash of the classpath when the code was compiled? The same code with the same imports could have wildly different compilations depending both on previous commands and on the jars we had pre-loaded

lihaoyi:
That's all for now =)

repl/src/main/scala/ammonite/repl/Storage.scala
laci37: I replaced these by using `Try.toOption`

repl/src/main/scala/ammonite/repl/interp/Evaluator.scala
laci37: On the next line I add `evalClassloader.getURLs` to the string being hashed, it's just not taken as a parameter.

repl/src/main/scala/ammonite/repl/interp/Evaluator.scala
laci37: As I see they only have the `Catching` line in common. Should I break up the `evalClass` call in `processLine` to separate `compileClass` and `loadClass` calls?

repl/src/main/scala/ammonite/repl/interp/Evaluator.scala
lihaoyi: Good point, I just didn't notice it. Perhaps we could move `cacheTag` into the companion object and pass this stuff in as arguments in manually? I feel like that would make it clearer what the tagging does and does not depend on, since currently potentially it takes the whole evaluator state as input

repl/src/main/scala/ammonite/repl/interp/Evaluator.scala
lihaoyi: The structure of the triple-quoted `compileClass` string is also 99% the same, and we probably want to maintain it the same. e.g. if we ever move to class-based-wrappers, we should only need to change it in one place. Currently we'd have to remember to update both string literals to keep things in sync

laci37:
Ready for review ;)

lihaoyi:
will take a look!

repl/src/main/scala/ammonite/repl/interp/Evaluator.scala
```diff
@@ -219,6 +296,15 @@ object Evaluator{
   def evaluatorRunPrinter(f: => Unit) = f
 
   /**
+   * This gives our cache tags for compile caching.
+   */
+  def cacheTag(code: String, imports: Seq[ImportData], classpath: String): String = {
```
lihaoyi: A few points here:

- We could pass in a `Seq[String]` here instead of `String`, since that's what you had outside the function.
- We most likely really want a `Seq[Array[Byte]]` because otherwise successive identically-named `-SNAPSHOT` jars will confuse this
- If we're hashing the bytes of each jar, we might want to cache the hashed values somewhere because it could get expensive
- If we're combining multiple hashes, we probably want to do it the "proper" way i.e. http://stackoverflow.com/a/4785309/871202; if it turns out to be slow we could use a weaker hash, but we shouldn't do ad-hoc string to combine the keys

repl/src/test/scala/ammonite/repl/ScriptTests.scala
```diff
@@ -7,92 +10,300 @@ object ScriptTests extends TestSuite{
...
-          error: Compilation Failed
-          Main.scala:\\d\\+: not found: value res2
-          res2 
-          ^
-          """)
+      'caching{
+        'blocks{
+          'one{
+            val storage = new MemoryStorage
+            val interp = new Interpreter(
```
lihaoyi: Tests should be DRY too =D this is a non-trivial section of code to initialize the `Interpreter`, pull it out into a def somewhere

repl/src/main/scala/ammonite/repl/Util.scala
```diff
@@ -150,6 +155,7 @@ object Ex{
 
 object Util{
   type IvyMap = Map[(String, String, String), Set[String]]
+  type CompileCache = (Traversable[(String, Array[Byte])], Seq[ImportData])
```
lihaoyi: - Can we pull `Traversable[(String, Array[Byte])]` out into its own type alias? We use it 9 times in the codebase.

repl/src/main/scala/ammonite/repl/Storage.scala
```diff
@@ -57,10 +59,48 @@ object Storage{
...
+        }
+      }
+    }
+
+    def compileCacheLoad(tag: String): Option[CompileCache] = {
+      val cacheDir = new File(dir + s"/compileCache/$tag")
+      if(!cacheDir.exists) None
+      else for{
+        metadataJson <- Try{
+          io.Source.fromFile(cacheDir + "/metadata.json").mkString
```
lihaoyi: Can we convert this to use `readFile` as well?

repl/src/main/scala/ammonite/repl/interp/Evaluator.scala
```diff
@@ -197,18 +230,62 @@ object Evaluator{
       // Exhaust the printer iterator now, before exiting the `Catching`
       // block, so any exceptions thrown get properly caught and handled
       evaluatorRunPrinter(printer(evalMain(cls).asInstanceOf[Iterator[String]]))
+      evaluationResult(wrapperName, newImports)
+    }
+
+    def compileCacheBlock(wrapperName: String, code: String, scriptImports: Seq[ImportData]): Res[(Class[_], Seq[ImportData])] = for {
```
lihaoyi: Can you talk a bit about how this is different from the "normal" way of doing things in `processLine`? I suspect there's some possible code sharing to be had, but I don't fully understand what you're trying to do here

lihaoyi:
That's all for now...

repl/src/main/scala/ammonite/repl/interp/Evaluator.scala
laci37: `processLine`'s algorithm is roughly compile, load, evaluate. `evalClass` does both the compiling and loading. `compileCacheBlock` compiles, then writes the cache, and then loads the class, and does no evaluation. (AFAIK the initializer runs only if the object is referenced somehow). The common code that made sense to me to pull is already in `loadClass` and `compileClass`.

lihaoyi: That makes sense, but doesn't `compileCacheBlock` only get used in one place where we're going to immediately evaluate it anyway? At least that's the only call site I see in this diff. I don't see any reason that normal command-line input couldn't do the same caching that loading scripts does, and it seems to me that it could simplify this code considerably

laci37: So I should compile-cache all code, not just scripts? It might make sense for predef and stuff like that.

lihaoyi: Yeah let's do that. compile-caching predef would be great, since it will rarely change, and people would pay the cost at every initialization!

lihaoyi: the rest of the code (i.e. basically the manually-entered code) I don't care so much, but if it doesn't cost us anything and lets us eliminate code duplication, no reason we can't cache that too =D

At some point the code cache will grow big enough to be a problem and we'll need to expire things, but we can deal with that when the time comes

laci37:
Under Scala 2.10 the following tests fail, because of some error coming from the reflection API:
* ScriptTests
  * module
    * compilationBlocks
      * annotation
      * loadIvy
      * preserveImports
      * syntax
    * caching
      * persistence
      * tags
* ProjectTests
  * load.ivy
    * standalone
    * complex
  * shapeless
  * scalaz
  * finagle
  * spire

The following tests fail in 2.10 because of PPrint printing the typenames with the wrapper prefix:
* EvaluatorTests
  * types
  * classes
  * multistatement
* AdvancedTests
  * pprint

lihaoyi:
Looks great! Some final things before landing:

- I turned off that Appveyor thing; I'll turn it on again when I can figure out how to get it working
- Can we move the "scala version is not 2.10" into a helper function? We're using it a lot more than we were previously.
- `classpathHash` does not seem to take into account the `newFileDict`, so I would expect it to behave badly if I enter `val x = 1` multiple times in a row into the command-line. It seems to do the right thing though! And properly re-compiles each snippet. Could you explain to me how it works?

This was a pretty big diff and review, thanks for putting up with it =)

lihaoyi:
@laci37 here's an idea for what's causing problems in 2.10; the reflection API is not thread safe in 2.10 and causes things to blow up randomly when run in parallel. 

I bet that's what you're seeing. You could verify this by trying to run these tests individually and see if they still fail; if they don't fail when run serially, and only fail when run in parallel, you're probably bumping into the reflection APIs non-thread-safety.

If that's the case, it could be worth investigating somewhat, so we can re-enable some of these tests =D

laci37:
We don't need `newFileDict` in the hash, because we assume that evaluated won't access wrappers with manual imports, only with generated imports. And generated imports hashed in the cache tag. Also generated imports contain the hash of the code from previous blocks in the wrapper names.

lihaoyi:
> We don't need newFileDict in the hash, because we assume that evaluated won't access wrappers with manual imports, only with generated imports. And generated imports hashed in the cache tag. Also generated imports contain the hash of the code from previous blocks in the wrapper names.

Ok cool!

laci37:
I determined, that the exception coming from the reflection API is not thread safety issue. Here is an example of the problem:
```
[info] -----------------------------------Results-----------------------------------
[info] tags		Failure('java.lang.AssertionError: assertion failed: 
TypeHistory(<no phase>:0,<?>,null)')
[info] Failures:
[info] 1/1     ammonite.repl.ScriptTests.module.caching.tags		
[info]     		java.lang.AssertionError: assertion failed: TypeHistory(<no phase>:0,<?>,null)
[info] scala.reflect.internal.Symbols$TypeHistory.<init>(Symbols.scala:3353)
[info] scala.reflect.internal.Symbols$Symbol.info_$eq(Symbols.scala:1251)
[info] scala.reflect.internal.Symbols$TypeSymbol.info_$eq(Symbols.scala:2816)
[info] scala.reflect.internal.Symbols$Symbol.setInfo(Symbols.scala:1257)
[info] scala.reflect.internal.Mirrors$Roots$RootClass$.<init>(Mirrors.scala:280)
[info] scala.reflect.internal.Mirrors$Roots.RootClass$lzycompute(Mirrors.scala:279)
[info] scala.reflect.internal.Mirrors$Roots.RootClass(Mirrors.scala:279)
[info] scala.reflect.internal.Mirrors$Roots$EmptyPackageClass$.<init>(Mirrors.scala:290)
[info] scala.reflect.internal.Mirrors$Roots.EmptyPackageClass$lzycompute(Mirrors.scala:290)
[info] scala.reflect.internal.Mirrors$Roots.EmptyPackageClass(Mirrors.scala:290)
[info] scala.reflect.internal.Mirrors$Roots.EmptyPackageClass(Mirrors.scala:245)
[info] scala.reflect.internal.Mirrors$RootsBase.init(Mirrors.scala:234)
[info] scala.tools.nsc.Global.rootMirror$lzycompute(Global.scala:59)
[info] scala.tools.nsc.Global.rootMirror(Global.scala:57)
[info] scala.tools.nsc.Global.rootMirror(Global.scala:37)
[info] scala.reflect.api.Trees$Traverser.<init>(Trees.scala:2822)
[info] scala.reflect.internal.Positions$DefaultPosAssigner.<init>(Positions.scala:38)
[info] scala.tools.nsc.ast.Positions$class.posAssigner(Positions.scala:36)
[info] scala.tools.nsc.Global.posAssigner$lzycompute(Global.scala:37)
[info] scala.tools.nsc.Global.posAssigner(Global.scala:37)
[info] scala.reflect.internal.Positions$class.atPos(Positions.scala:61)
[info] scala.reflect.internal.SymbolTable.atPos(SymbolTable.scala:13)
[info] scala.tools.nsc.ast.parser.Parsers$Parser.atPos(Parsers.scala:689)
[info] scala.tools.nsc.ast.parser.Parsers$Parser.literal(Parsers.scala:1115)
[info] scala.tools.nsc.ast.parser.Parsers$Parser.simpleExpr(Parsers.scala:1540)
[info] scala.tools.nsc.ast.parser.Parsers$Parser.prefixExpr(Parsers.scala:1520)
[info] scala.tools.nsc.ast.parser.Parsers$Parser.postfixExpr(Parsers.scala:1482)
[info] scala.tools.nsc.ast.parser.Parsers$Parser.parseOther$1(Parsers.scala:1394)
[info] scala.tools.nsc.ast.parser.Parsers$Parser.expr0(Parsers.scala:1448)
[info] scala.tools.nsc.ast.parser.Parsers$Parser.expr(Parsers.scala:1289)
[info] scala.tools.nsc.ast.parser.Parsers$Parser.expr(Parsers.scala:1284)
[info] scala.tools.nsc.ast.parser.Parsers$Parser.funDefRest(Parsers.scala:2573)
[info] scala.tools.nsc.ast.parser.Parsers$Parser.funDefOrDcl(Parsers.scala:2540)
[info] scala.tools.nsc.ast.parser.Parsers$Parser.defOrDcl(Parsers.scala:2423)
[info] scala.tools.nsc.ast.parser.Parsers$Parser.nonLocalDefOrDcl(Parsers.scala:2435)
[info] scala.tools.nsc.ast.parser.Parsers$Parser$$anonfun$templateStatSeq$1$$anonfun$apply$5.apply(Parsers.scala:3001)
[info] scala.tools.nsc.ast.parser.Parsers$Parser$$anonfun$templateStatSeq$1$$anonfun$apply$5.apply(Parsers.scala:3001)
[info] scala.tools.nsc.ast.parser.Parsers$Parser.joinComment(Parsers.scala:683)
[info] scala.tools.nsc.ast.parser.Parsers$Parser$$anonfun$templateStatSeq$1.apply(Parsers.scala:3001)
[info] scala.tools.nsc.ast.parser.Parsers$Parser$$anonfun$templateStatSeq$1.apply(Parsers.scala:2970)
[info] scala.tools.nsc.ast.parser.Parsers$Parser.checkNoEscapingPlaceholders(Parsers.scala:428)
[info] scala.tools.nsc.ast.parser.Parsers$Parser.templateStatSeq(Parsers.scala:2970)
[info] scala.tools.nsc.ast.parser.Parsers$Parser.templateStats(Parsers.scala:2954)
[info] ammonite.repl.interp.CompilerCompatibility$.trees(CompilerCompatibility.scala:19)
[info] ammonite.repl.interp.Compiler$$anon$4.parse(Compiler.scala:219)
[info] ammonite.repl.interp.Interpreter$$anonfun$3$$anonfun$apply$5.apply(Interpreter.scala:199)
[info] ammonite.repl.interp.Interpreter$$anonfun$3$$anonfun$apply$5.apply(Interpreter.scala:199)
[info] ammonite.repl.interp.Preprocessor$$anon$1$$anonfun$13.apply(Preprocessor.scala:106)
[info] ammonite.repl.interp.Preprocessor$$anon$1$$anonfun$13.apply(Preprocessor.scala:106)
[info] scala.collection.TraversableLike$$anonfun$map$1.apply(TraversableLike.scala:244)
[info] scala.collection.TraversableLike$$anonfun$map$1.apply(TraversableLike.scala:244)
[info] scala.collection.mutable.ResizableArray$class.foreach(ResizableArray.scala:59)
[info] scala.collection.mutable.ArrayBuffer.foreach(ArrayBuffer.scala:47)
[info] scala.collection.TraversableLike$class.map(TraversableLike.scala:244)
[info] scala.collection.AbstractTraversable.map(Traversable.scala:105)
[info] ammonite.repl.interp.Preprocessor$$anon$1.complete(Preprocessor.scala:106)
[info] ammonite.repl.interp.Preprocessor$$anon$1.apply(Preprocessor.scala:101)
[info] ammonite.repl.interp.Interpreter$$anonfun$2.apply(Interpreter.scala:64)
[info] ammonite.repl.interp.Interpreter$$anonfun$2.apply(Interpreter.scala:64)
[info] scala.collection.TraversableLike$$anonfun$map$1.apply(TraversableLike.scala:244)
[info] scala.collection.TraversableLike$$anonfun$map$1.apply(TraversableLike.scala:244)
[info] scala.collection.mutable.ResizableArray$class.foreach(ResizableArray.scala:59)
[info] scala.collection.mutable.ArrayBuffer.foreach(ArrayBuffer.scala:47)
[info] scala.collection.TraversableLike$class.map(TraversableLike.scala:244)
[info] scala.collection.AbstractTraversable.map(Traversable.scala:105)
[info] ammonite.repl.interp.Interpreter.processScript(Interpreter.scala:64)
[info] ammonite.repl.interp.Interpreter.processModule(Interpreter.scala:58)
[info] ammonite.repl.interp.Interpreter$$anon$1$load$.module(Interpreter.scala:128)
[info] ammonite.repl.interp.Interpreter$$anon$1$load$.module(Interpreter.scala:133)
[info] ammonite.repl.ScriptTests$$anonfun$32$$anonfun$apply$12$$anonfun$apply$23$$anonfun$apply$34.apply(ScriptTests.scala:245)
[info] ammonite.repl.ScriptTests$$anonfun$32$$anonfun$apply$12$$anonfun$apply$23$$anonfun$apply$34.apply(ScriptTests.scala:9)
[info] Tests: 1
[info] Passed: 0
[info] Failed: 1
[error] Failed tests:
[error] 	ammonite.repl.ScriptTests
[error] (repl/test:testOnly) sbt.TestsFailedException: Tests unsuccessful
[error] Total time: 5 s, completed Jul 25, 2015 7:28:01 PM
```

laci37:
I rebased this, so it's ready for review.

lihaoyi:
Looks great =D

PR #123, opened 2015-07-27
Fix compiler error on parsing

lihaoyi:
- Don't cache manually-entered commands; it makes the mangled names of classes really ugly =/
- Don't cache the `ReplBridge` instantiation; it turns out that this was what was causing that compiler error that @laci37 was seeing. There's probably a separate bug here, but we can remove this instantiation when we fix that bug. Apart from happening on 2.10, the same compiler error was happening intermittently in 2.11 too
- Re-enable all the tests we disabled in 2.10
- Move the last few manual calls to `scala.util.Properties` onto `scala2_10`


lihaoyi:
Review by @laci37 since it's touching your code all over the place.

laci37:
Looks OK to me.

PR #142, opened 2015-08-06
Limit imports from scripts to last compilation block

laci37:
These changes limit imports coming from scripts to the last compilation block of the script, so scripts won't spam the REPL scope with internal variables.


lihaoyi:
This looks good! There's some finnickiness with loading modules in modules, but i'll open a new issue for that

PR #146, opened 2015-08-07
Fix Ctrl-f and add Alt-delete.

non:
Ctrl-f was erroneously moving back intead of forward.
Alt-delete's behavior is identical to Ctrl-w, but was
missing.


non:
Oh wait, huge bug, sorry! Don't merge this as-is.

non:
OK, this should be good to go now.

lihaoyi:
Looks great. Thanks!

Issue #145, opened 2015-08-07
Support compiler plugins


non:
Would it be possible to load compiler plugins, similarly to how libraries can be loaded?

Since Ammonite already supports restarting the compiler (in the case of a crash) it might be possible to do the same thing to load a compiler plugin.

alexarchambault:
This can definitely be done, I'm already doing it [here](https://github.com/alexarchambault/ammonite-shell/blob/master/interpreter/src/main/scala/ammonite/interpreter/Compiler.scala#L165) (illustrated with [simulacrum](https://github.com/alexarchambault/jupyter-scala/blob/master/examples/libraries/Simulacrum.ipynb)).

non:
Interesting. Do you see yourself continuing to keep Ammonite-Shell in sync with Ammonite. Is it possible @lihaoyi would accept your changes upstream?

alexarchambault:
I plan to update it this month. I wouldn't be against some of its features landing here, the main ones being class wrapper based repl, and cleaner class loader. @lihaoyi wasn't too interested last time I asked.

lihaoyi:
Adding compiler plugins would be really easy AFAIK; anyone want to take a shot at implementing it? We can already reload the compiler completely transparently, so it's just a matter of making a new one with some new plugins or options

> @lihaoyi wasn't too interested last time I asked.

That's not an accurate assessment, I just wanted a clear explanation of what the features were for, and some unit tests to demonstrate their usage, and then you got bored and left to do your own thing =P

I'm open to all of these features, but surely I can be excused for wanting some kind of discussion/understanding/testing before accepting a moderately-large diff..

alexarchambault:
@lihaoyi There were things in my devs that were not clear enough to me at the time. I needed to try more things, out of a PR / reviewing frame, before having something more satisfactory. (Sorry if it felt harsh :-| )

I'm giving the compiler plugin - in this repo - a try (if nobody's already on it?).

lihaoyi:
No offense taken; go for it =D

PR #147, opened 2015-08-07
Load available compiler plugins

alexarchambault:
Fixes https://github.com/lihaoyi/Ammonite/issues/145

I haven't added tests for when loading a plugin fails. Having the ability to add resources to the class loader on the fly, to add a `scalac-plugin.xml`, would make that easier. One could directly write the plugin from Ammonite then.

### DIFF
--- a/repl/src/main/scala/ammonite/repl/frontend/ReplAPI.scala
+++ b/repl/src/main/scala/ammonite/repl/frontend/ReplAPI.scala
@@ -145,7 +145,7 @@ trait OpsAPI{
 
 }
 // End of OpsAPI
-trait Load extends (String => Unit){
+trait LoadJar {
   /**
    * Load a `.jar` file
    */
@@ -154,7 +154,8 @@ trait Load extends (String => Unit){
    * Load a library from its maven/ivy coordinates
    */
   def ivy(coordinates: (String, String, String), verbose: Boolean = true): Unit
-
+}
+trait Load extends (String => Unit) with LoadJar{
   /**
    * Loads a command into the REPL and
    * evaluates them one after another
@@ -169,6 +170,8 @@ trait Load extends (String => Unit){
   def exec(path: Path): Unit
 
   def module(path: Path): Unit
+
+  def plugin: LoadJar
 
 }
 
--- a/repl/src/main/scala/ammonite/repl/interp/Compiler.scala
+++ b/repl/src/main/scala/ammonite/repl/interp/Compiler.scala
@@ -117,13 +117,14 @@ object Compiler{
             dirDeps: Seq[java.io.File],
             dynamicClasspath: VirtualDirectory,
             evalClassloader: => ClassLoader,
+            pluginClassloader: => ClassLoader,
             shutdownPressy: () => Unit): Compiler = new Compiler{
 
     val PluginXML = "scalac-plugin.xml"
 
     lazy val plugins0 = {
       import scala.collection.JavaConverters._
-      val loader = evalClassloader
+      val loader = pluginClassloader
 
       val urls = loader
         .getResources(PluginXML)
--- a/repl/src/main/scala/ammonite/repl/interp/Compiler.scala
+++ b/repl/src/main/scala/ammonite/repl/interp/Compiler.scala
@@ -119,6 +119,42 @@ object Compiler{
             evalClassloader: => ClassLoader,
             shutdownPressy: () => Unit): Compiler = new Compiler{
 
+    val PluginXML = "scalac-plugin.xml"
+
+    lazy val plugins0 = {
+      import scala.collection.JavaConverters._
+      val loader = evalClassloader
+
+      val urls = loader
+        .getResources(PluginXML)
+        .asScala
+        .toVector
+
+      val plugins = for {
+        url <- urls
+        elem = scala.xml.XML.load(url.openStream())
+        name = (elem \\ "plugin" \ "name").text
+        className = (elem \\ "plugin" \ "classname").text
+        // acyclic seems to conflict with AmmonitePlugin (happens during the tests in particular), so it's
+        // filtered out here. Else it raises:
+        //   scala.reflect.internal.FatalError: Multiple phases want to run right after typer; followers:
+        //   AmmonitePhase,acyclic; created phase-order.dot
+        if name != "acyclic"
+        if name.nonEmpty && className.nonEmpty
+        classOpt =
+          try Some(loader.loadClass(className))
+          catch { case _: ClassNotFoundException => None }
+      } yield (name, className, classOpt)
+
+      val notFound = plugins.collect{case (name, className, None) => (name, className) }
+      if (notFound.nonEmpty) {
+        for ((name, className) <- notFound.sortBy(_._1))
+          Console.err.println(s"Implementation $className of plugin $name not found.")
+      }
+
+      plugins.collect{case (name, _, Some(cls)) => name -> cls }
+    }
+
     var logger: String => Unit = s => ()
 
     var lastImports = Seq.empty[ImportData]
@@ -128,7 +164,20 @@ object Compiler{
         jarDeps, dirDeps, dynamicClasspath, logger
       )
       val scalac = new nsc.Global(settings, reporter) { g =>
-        override lazy val plugins = List(new AmmonitePlugin(g, lastImports = _))
+        override lazy val plugins = List(new AmmonitePlugin(g, lastImports = _)) ++ {
+          for {
+            (name, cls) <- plugins0
+            plugin = Plugin.instantiate(cls, g)
+            initOk =
+              try CompilerCompatibility.pluginInit(plugin, Nil, g.globalError)
+              catch { case ex: Exception =>
+                Console.err.println(s"Warning: disabling plugin $name, initialization failed: $ex")
+                false
+              }
+            if initOk
+          } yield plugin
+        }
+
         override def classPath = platform.classPath // Actually jcp, avoiding a path-dependent type issue in 2.10 here
         override lazy val platform: ThisPlatform = new JavaPlatform{
           val global: g.type = g
--- a/repl/src/test/scala/ammonite/repl/AdvancedTests.scala
+++ b/repl/src/test/scala/ammonite/repl/AdvancedTests.scala
@@ -262,13 +262,32 @@ object AdvancedTests extends TestSuite{
     }
     'compilerPlugin{
       check.session("""
+        @ // Make sure plugins from eval class loader are not loaded
+
         @ load.ivy("org.spire-math" %% "kind-projector" % "0.6.3")
+
+        @ trait TC0[F[_]]
+        defined trait TC0
+
+        @ type TC0EitherStr = TC0[Either[String, ?]]
+        error: not found: type ?
+
+        @ // This one must be loaded
+
+        @ load.plugin.ivy("org.spire-math" %% "kind-projector" % "0.6.3")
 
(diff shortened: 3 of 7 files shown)


lihaoyi:
Code looks fine.

What do you think about making someone explicitly specify which plugin they want to use with a separate command, or adding the `classOf[MyPlugin]` to some `Ref[List[Class[Plugin]]]` that's exposed in the `ReplAPI`?

Auto-detecting plugins on classpath is convenient but I feel kinda sketchy =/

alexarchambault:
It feels a bit less natural this way. Most of the time, one doesn't know the exact name of the plugin class. We're just given ivy coordinates (like [here](https://github.com/non/kind-projector#using-the-plugin), or anywhere else).

Or maybe the autoloading could be disabled by default, and enabled through `ReplAPI`?

lihaoyi:
Good point. How does SBT do it? I know they require a separate incantation
for loading plugins vs normal libraries. I'm pretty sure just adding e.g.
acyclic to your librayDependencies doesn't cause the plugin to run.

If it turns out they're doing crazy hacky stuff to make that happen the
status quo is fine too

alexarchambault:
I was thinking too: it could be the way you're proposing, with a helper to give a `List[Class[Plugin]]` from what's available in the class loader?

About SBT, I haven't looked at the very details, but I think it just puts things in a class loader (in particular, dependencies from the `plugin` Ivy scope - the `addCompilerPlugin` just puts dependencies there), and then scalac finds the plugin itself. There are helpers for that in it for sure (that's where I saw how to load plugins this way.)

lihaoyi:
> About SBT, I haven't looked at the very details, but I think it just puts things in a class loader (in particular, dependencies from the plugin Ivy scope - the addCompilerPlugin just puts dependencies there), and then scalac finds the plugin itself

That seems like a great way to do it =) We can easily maintain our own `PluginClassLoader`, have our own `addCompilePlugin` that stuffs things into that, and use that to do all the discovery. It would take a bit more work than dumping everything in the current classloader, but shouldn't be too hard.

alexarchambault:
That's how I'm doing it in https://github.com/alexarchambault/ammonite-shell :-) It felt more straightforward to do here, as I was already using a "cleaned" class loader (that is, users don't see scala-compiler, or the ammonite internals, from the repl, only ReplAPI and scala-library, when it starts)

alexarchambault:
One problem though, is that macros are executed in this special class loader. So if one defines some from the repl, or calls macros from the standard dependencies, these have to be available in it. The plugin class loader must have everything from the current class loader, plus plugin specific things.

lihaoyi:
> users don't see scala-compiler, or the ammonite internals, from the repl, only ReplAPI and scala-library, when it starts

That sounds really nice =D Want to see if you could port that over as well? It's something I've wanted to do for a while but haven't gotten to it.

> The plugin class loader must have everything from the current class loader, plus plugin specific things.

That sounds reasonable, and slightly knotty. Does your ammonite-shell fork support defining macros in the prompt?

alexarchambault:
I was editing my previous answer: I did it with same classloader for macros and plugins in my fork, but it may possible to do it another way. That would make 3 class loaders at the end :-) (one for plugins, one for macros, one for the runtime)

lihaoyi:
> I did it with same classloader for macros and plugins in my fork

But aren't "macros" just user code, and live in the same classloader as all the other user stuff? Does that mean that although the user doesn't see `scala-compiler` and all the other rubbish in his current classloader, if he `load.ivy`s a wrong version of `scala-compiler`, he could break the shell's internals?

Not that that would be any worse than the current behavior =P I just want to understand it

alexarchambault:
They are. What I'm doing in https://github.com/alexarchambault/ammonite-shell, is that I have two class loaders when starting, the user one with ReplAPI + scala-library, and the macro one with this + scala-compiler. Then, during the session, everything gets added to both class loaders. I was using only the first one initially, but some macro require scala-compiler (like the ones of shapeless), so it has to be around for them.

But if plugins have a third class loader (only for plugins, not macros), maybe it doesn't need to be given the macros et al. Just scala-compiler at the start, then plugin dependencies, and that's it. I haven't tested, but I think it should work. Plugins don't need the session classes, or all the dependencies. Only scala-compiler, and their own dependencies.

alexarchambault:
I was about to have a look at https://github.com/alexarchambault/ammonite-shell, during the month. I'll try to see what can be brought back here.

lihaoyi:
From what I can figure a working classloader arrangement would look something like this:

```

 ----------> Ammonite Code (scalaparse, upickle, ivy, jline, ...)
 |
 |
 |
Boot ------> Compiler Plugins
 |
 |
 |
 ----------> ReplAPI + scala-library + scala-reflect + scala-compiler 
                                     |
                                     |
                                     |
                                     v
                           User Code + Loaded Libs + Macros
                                     |
                                     |
                                     |
                                     v
                           User Code + Loaded Libs + Macros
                                     |
                                     |
                                     |
                                     v
                                    ...
```

Where each of the phrases is a classloader and the arrows represent the dependencies between them. In this design, the Ammonite Code and the Compiler Plugins are totally separate from each other, and all the user code (including macros) which lives in an "almost-clean" classloader which includes `scala-{library, reflect, compiler}`. Not as good as totally clean, but better than having all the Ammonite rubbish leaking into it =P

If we wanted to we can merge all the user code + loaded libs and the replAPI stuff into a single classloader, but eventually I want people to be able to "checkpoint" their session and roll back to different points via throwing away portions of the classloader chain.

lihaoyi:
That diagram is a bit idealistic, but as an intermediate action stuffing the Compiler Plugins in a separate classloader would both make me happy and be a step towards the eventual goal =)

non:
Just want to add +1 on this PR. Thanks @alexarchambault!

alexarchambault:
I pushed an extra commit that adds a separate ClassLoader for plugins. @lihaoyi I changed a bit `Load` for that, in a way that was the most straightforward to me, tell me if it's fine for you.

@non Thanks!

alexarchambault:
About further developments regarding the class loader(s), I'll have a look very soon at my related projects, I'll reach you here or on gitter to explain quickly how I'm doing it for now, and how this could be ported here (taking your remarks into account too).

repl/src/main/scala/ammonite/repl/interp/Compiler.scala
```diff
@@ -117,8 +117,45 @@ object Compiler{
...
+        .asScala
+        .toVector
+        .map { url =>
+          val elem = scala.xml.XML.load(url.openStream())
+          val name = (elem \\ "plugin" \ "name").text
+          val className = (elem \\ "plugin" \ "classname").text
+          name -> className
+        }
+        .filter{
+          case ("acyclic", _) => false // conflicts with AmmonitePlugin
```
lihaoyi: I can't make sense of this comment, even though I think I know what it does (avoid loading Acyclic, which lives on the parent classpath).

On the other hand, I don't really see why loading Acyclic as a compiler plugin is a problem? Why does it conflict with AmmonitePlugin?

Is this filter still necessary? If so, we should merge it with the `collect` phase below and add a better description. If not we should kill it

repl/src/main/scala/ammonite/repl/frontend/ReplAPI.scala
```diff
@@ -145,7 +145,7 @@ trait OpsAPI{
 
 }
 // End of OpsAPI
-trait Load extends (String => Unit){
+trait LoadJar {
```
lihaoyi: This is great =D

repl/src/main/scala/ammonite/repl/interp/Compiler.scala
```diff
@@ -117,8 +117,45 @@ object Compiler{
...
+          case _ => true
+        }
+        .collect { case (name, className) if name.nonEmpty && className.nonEmpty =>
+          name -> (
+            className,
+            try Some(loader.loadClass(className)) catch { case _: ClassNotFoundException => None }
```
lihaoyi: Can we move this fella into a local `val`? I find this kind of multi-line-tuple-following-infix-operation (in this case `->`) pretty awkward to read

repl/src/main/scala/ammonite/repl/interp/Compiler.scala
```diff
@@ -117,8 +117,45 @@ object Compiler{
...
+          )
+        }
+
+      val (found, notFound) = plugins.partition(_._2._2.nonEmpty)
```
lihaoyi: It seems to me that it would be easier for the fast path to use `collect` on both the `found` and `notFound` cases; you could then do partial destructing within the collect and wouldn't need to ignore the things down below

repl/src/main/scala/ammonite/repl/interp/Compiler.scala
```diff
@@ -128,7 +165,18 @@ object Compiler{
         jarDeps, dirDeps, dynamicClasspath, logger
       )
       val scalac = new nsc.Global(settings, reporter) { g =>
-        override lazy val plugins = List(new AmmonitePlugin(g, lastImports = _))
+        override lazy val plugins = List(new AmmonitePlugin(g, lastImports = _)) ++
+          plugins0
```
lihaoyi: This chain would be way neater as a for comprehension =)

repl/src/test/scala/ammonite/repl/AdvancedTests.scala
```diff
@@ -260,6 +260,17 @@ object AdvancedTests extends TestSuite{
         error: not found: value x
       """)
     }
+    'compilerPlugin{
+      check.session("""
+        @ load.plugin.ivy("org.spire-math" %% "kind-projector" % "0.6.3")
```
lihaoyi: Can we add a negative test to ensure that `load.ivy`ing `kind-projector` without the `plugin` bit will fail to activate the plugin, and that `TC[Either[String, ?]]` results in a compile error? Could even be part of the same `check.session`

lihaoyi:
That's all for now. This looks great, thanks for sticking through this!

repl/src/test/scala/ammonite/repl/AdvancedTests.scala
lihaoyi: We probably should also have a negative compile test to make sure stuff loaded via `load.plugin` doesn't end up in the `evalClassloader`.

repl/src/main/scala/ammonite/repl/interp/Compiler.scala
alexarchambault: It still is. See the comment in the new commit.

repl/src/main/scala/ammonite/repl/interp/Compiler.scala
alexarchambault: Done.

repl/src/main/scala/ammonite/repl/interp/Compiler.scala
alexarchambault: Done too.

repl/src/main/scala/ammonite/repl/interp/Compiler.scala
alexarchambault: Fixed.

repl/src/test/scala/ammonite/repl/AdvancedTests.scala
alexarchambault: Fixed too.

alexarchambault:
Just pushed new commits.

lihaoyi:
Looks great!

PR #153, opened 2015-08-10
Add support for C-p/C-n for prev/next in history.

non:
This commit also makes some adjustments to the history code
that hopefully make it a bit clearer what is going on.

It also makes a change to cursor positioning. In most
readline-based programs (e.g. bash), if the cursor is
currently at the end of the line, it will remain at the
end of the line after moving through the history. The
current behavior leaves the cursor at position 0. With
this change, a specific check is made to see if the
cursor is at the end of line, and if so the cursor
is moved to the "new" end of the line.


lihaoyi:
> In most readline-based programs (e.g. bash), if the cursor is currently at the end of the line, it will remain at the end of the line after moving through the history.

From my experimentation, it seems that in Bash/Python/ghci, the cursor is always set to the end of the line after navigating in history, no matter where it was before you navigated. Should we do that instead? It seems to me that would simplify things.

lihaoyi:
Thanks!

Issue #151, opened 2015-08-10
load.ivy should fail more helpfully in the absence of artifact dependencies.


mrmechko:
Failure is acceptable from time to time, particularly if caused by someone who didn't quite read the docs, but it should make us better as programmers and as people.  Late night grandiloqence aside, we should be getting helpful error messages when certain activities, such as `ivy.load`, fail.  This issue distills the problem encountered in #150.

In particular, **Ammonite should return detailed explanation of why `ivy.load` failed regardless of whether or not `verbose = true` is specified, and do so every time it runs**

What ensues is an description of the current behavior:

_Currently, it is explicitly stated that Ammonite only loads from Maven Central, so this issue does not address tactics on loading the offending snapshot dependencies._

**On the first attempt** to load an maven artifact with snapshot dependencies, and only if verbose output is requested, we get a full description of why things went wrong:

```sbt
:: problems summary ::
:::: WARNINGS
		module not found: com.github.mrmechko#swordnet_2.11;2.0-SNAPSHOT

	==== central: tried

	  http://repo1.maven.org/maven2/com/github/mrmechko/swordnet_2.11/2.0-SNAPSHOT/swordnet_2.11-2.0-SNAPSHOT.pom

	  -- artifact com.github.mrmechko#swordnet_2.11;2.0-SNAPSHOT!swordnet_2.11.jar:

	  http://repo1.maven.org/maven2/com/github/mrmechko/swordnet_2.11/2.0-SNAPSHOT/swordnet_2.11-2.0-SNAPSHOT.jar
```

**However** further attempts to load the offending library result in a rather uninformatively broad error:

```sbt
scala.reflect.internal.MissingRequirementError: object java.lang.Object in compiler mirror not found.
	scala.reflect.internal.MissingRequirementError$.signal(MissingRequirementError.scala:17)
	scala.reflect.internal.MissingRequirementError$.notFound(MissingRequirementError.scala:18)
	scala.reflect.internal.Mirrors$RootsBase.getModuleOrClass(Mirrors.scala:53)
	scala.reflect.internal.Mirrors$RootsBase.getModuleOrClass(Mirrors.scala:45)
	scala.reflect.internal.Mirrors$RootsBase.getModuleOrClass(Mirrors.scala:45)
	scala.reflect.internal.Mirrors$RootsBase.getModuleOrClass(Mirrors.scala:66)
	scala.reflect.internal.Mirrors$RootsBase.getClassByName(Mirrors.scala:102)
[...]

PR #154, opened 2015-08-10
An improvement to the state of IvyThing.scala

mrmechko:
A small fix to help the state of #151.  `resolveArtifact` now returns an `Option[Array[File]]` explicitly, instead of the previous `Array[File]`, where `None` indicates that the resolution failed somehow.  Corresponding updates to `ivy(...)` in `Interpreter.scala` were made.  In addition, `ivy(...)` prints `success` or `failure` as appropriate on attempting to get artifacts from Maven.

### DIFF
--- a/repl/src/main/scala/ammonite/repl/interp/Interpreter.scala
+++ b/repl/src/main/scala/ammonite/repl/interp/Interpreter.scala
@@ -146,17 +146,24 @@ class Interpreter(prompt0: Ref[String],
       def ivy(coordinates: (String, String, String), verbose: Boolean = true): Unit = {
         val (groupId, artifactId, version) = coordinates
         storage().ivyCache().get((groupId, artifactId, version)) match{
-          case Some(ps) => ps.map(new java.io.File(_)).map(handleJar)
+          case Some(ps) => {
+            ps.map(new java.io.File(_)).map(handleJar)
+            Console.err.println("success")
+          }
           case None =>
-            val resolved = IvyThing.resolveArtifact(groupId, artifactId, version, if (verbose) 2 else 1)
-            storage().ivyCache() = storage().ivyCache().updated(
-              (groupId, artifactId, version),
-              resolved.map(_.getAbsolutePath).toSet
-            )
+            val resolution = IvyThing.resolveArtifact(groupId, artifactId, version, if (verbose) 2 else 1)
+            resolution map { resolved =>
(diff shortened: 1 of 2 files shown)


repl/src/main/scala/ammonite/repl/interp/Interpreter.scala
```diff
@@ -146,17 +146,24 @@ class Interpreter(prompt0: Ref[String],
...
+              storage().ivyCache() = storage().ivyCache().updated(
+                (groupId, artifactId, version),
+                resolved.map(_.getAbsolutePath).toSet
+              )
+
+              resolved.map(handleJar)
+            } match {
+              case Some(_) => Console.err.println("success")
+              case None => Console.err.println("failure")
+            }
```
non: Maybe consider using `.foreach` instead of `.map` here, since the code is side-effecting and the return value is not needed. For the bottom `match`, I think you can use `resolution` instead.

You could also simplify the output code by doing something like:

```scala
Console.err.println(resolution.fold("failure")(_ => "success"))
```

mrmechko:
Thanks for the suggestions. I live to learn.  Looking more at the Java code for Ivy, it seems that there's a lot of logic for identifying how and when things fail.  Should all those potential errors and messages be dealt with and output from `resolveArtifact` as necessary and just have the `Interpreter` return success or failure or should `resolveArtifact` push the failure message up to `Interpreter`?

non:
@mrmechko My inclination would be to pass the data along to the interpreter so it can decide how to output errors. I could imagine a case class to wrap failures that has a method to produce the kind of `String` you'd want to print, and let the interpreter use that. That way it's easy to modify the interpreter to help debug artifact resolution (or just increase verbosity). Just my 2c.

lihaoyi:
I agree with @non that we should make it pass some kind of structured data back out. Printing `success` or `failure` to stderr is basically the worst way of doing it. In particular, it's not testable at all =P And why are we printing `success` to `Console.err` anyway?

Could we at least throw an `ArtifactResolutionFailed` exception or something to make it extra-extra-clear that something failed? Perhaps with the names of the dependencies which failed. At least that way it'd be hard to miss the stack trace that pops up.

lihaoyi:
Replaced by https://github.com/lihaoyi/Ammonite/commit/9103e0a6f4600c9bfe634c807ba930d638ad7c29

PR #156, opened 2015-08-11
Improve the REPL's startup banner.

non:
This commit does a few things:

 * Shorten the "loading" message
 * Include Ammonite version in banner
 * Include Scala and Java versions in banner too


non:
I'm 98% sure that the test failures are just resolution issues in Travis, FWIW.

lihaoyi:
Should be ok

PR #157, opened 2015-08-11
Enable C-d to delete when line is not empty.

non:
Most tools only use C-d to exit when the line is empty. Previously
a C-d would exit instead of delete, which will be surprising for
bash users.


lihaoyi:
Looks good, thanks!

PR #158, opened 2015-08-11
Support clearing the screen via C-l.

non:
This should work the same as bash et al.

If there is an "in progress" line it is preserved and redrawn
at position (0, 0).


lihaoyi:
Looks great

PR #165, opened 2015-08-13
Add Scala 2.10.5 for SBT 0.13.9

nightscape:
It would be great if you could publish for Scala 2.10.5 because otherwise one can't upgrade to SBT 0.13.9 (which brings a lot of fixes and maybe some speedups).

Thanks!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 language: scala
 scala:
 - 2.10.4
+- 2.10.5
 - 2.11.4
 - 2.11.7
 jdk:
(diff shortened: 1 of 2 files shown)


lihaoyi:
This doesn't even compile right now, as far as I can tell. This probably should get added to the `.travis.yml` build matrix too to make sure it actually works

lihaoyi:
(the reason we don't run +test on travis is because the combinatorial explosion would make it realllyyyy slowwwwwww, so I just picked a random-ish subset that I hope will catch most problems)

nightscape:
I've updated ScalaMacros to 2.0.1 in https://github.com/lihaoyi/Ammonite/pull/166
This should give us artifacts for Scala 2.10.5.
Once you merge the PR 166, I'll rebase this branch. Then we should decide which Scala versions to keep and which to ditch.

nightscape:
Much better now :)

lihaoyi:
Thanks!

PR #175, opened 2015-08-17
Use Ctrl-h for backspace

SRGOM:
Does not modify the yank ring.

Tests pass fine. Please let me know if more tests are required.


SRGOM:
I should note that there are certain tests failing in the main branch itself. I ran `sbt 'testOnly ammonite.terminal.EditTests'`

lihaoyi:
Thanks!

PR #176, opened 2015-08-17
Ammonite sshd REPL

dant3:
This change introduces the whole new module of sshd server serving Ammonite REPL shell sessions. The whole point of having such server is to be able to embed this with your existing java/scala software. Some possible use cases are:
- Debugging with server running in progress & serving real requests, having access to real database.
- Running repl on Android device as simple Android application with your desktop shell connectivity. This has advantage of no need in root access to do this, compared to installing sshd and ammonite as is on device. This of course needs compiler support as well, but doable.
- Very useful for thin tuneup of desktop gui app. Imagine how you can simply tune view margins back and forth on a running Swing/FX app (like IntelliJ Idea?).


lihaoyi:
I haven't looked at this in detail but for starters

- A PR of this size probably needs some explanation of "what is what", both in the PR description, as well as short doc-comments on each of the classes involved. You don't need 100% scaladoc coverage, but you should have *something* to tell a newbie what each major class represents.

- I'd want tests to verify that any of this stuff works at all, at least in a few cases (e.g. successful login works, can run code over SSH, bad password fails, ...). 

- It needs docs (in `Index.scalatex`), probably as a new section under `Debugging`, with examples of how to use, taken from the unit tests in order to verify the examples are correct

- The build will need to be green

I'll do a more detailed review later once all the above things make the code easier to review =P

lihaoyi:
Let me know when you're done working and I should look at this again

dant3:
Sure thing. I primary started this PR in a first place to see if it passes cross-compiling and tests with different scala/jvm versions. I think some api refactoring might be involved as well, since the whole thing could have a bit more flexibility than it is now.

dant3:
@lihaoyi Looks like I've finished with all that you mentioned in the beginning. There are some docs, there are some tests that makes obvious that the thing is working and hopefully the build will turn green now, given that Travis will not give new surprises. Please, review and lets discuss that else needs to be done.
I've put a couple of words in `Index.scalatex`, but not sure about the sectioning and actual markup. Also, perhaps we could come with a better example, I have one with spray in mind but not sure I'll manage to put it together this week.

build.sbt
```diff
@@ -151,6 +151,24 @@ lazy val repl = project
...
 
+
+/**
+ * REPL available via remote ssh access. Pluggable into any app environment to allow live hacking on a live application.
+ */
+lazy val sshd = project.
+    dependsOn(repl).
+    settings(sharedSettings:_*).
+    settings(
+      name := "ammonite-repl-sshd",
```
lihaoyi: Let's just name this `ammonite-sshd`, and the package `ammonite.sshd`, rather than putting it into the `ammonite.repl` namespace

readme/Index.scalatex
```diff
@@ -368,6 +368,29 @@
...
+
+    @hl.scala
+      libraryDependencies += "com.lihaoyi" % "ammonite-repl-sshd" % "@ammonite.Constants.version" cross CrossVersion.full
+
+    @p
+      Now add repl server to your application:
+
+    @hl.scala
+      import ammonite.repl.sshd._
+      val replServer = new SshdRepl(SshServerConfig(port = 22222, username = "repl", password = "your_secure_password"))
```
lihaoyi: Does it currently support key-based authentication? If so we should describe how. If not, we should also mention it =)

sshd/src/main/scala/ammonite/repl/sshd/SshServer.scala
```diff
@@ -0,0 +1,65 @@
...
+import org.apache.sshd.{ SshServer ⇒ SshServerImpl }
+
+/**
+ * a factory to produce shell-only ssh server
+ */
+object SshServer {
+  import Implicits._
+  type RemoteShellServer = ((InputStream, OutputStream) ⇒ Unit)
+
+  def apply(options:SshServerConfig, terminal: RemoteShellServer) = {
```
lihaoyi: Syntax: space after `:`

sshd/src/main/scala/ammonite/repl/sshd/util/Environment.scala
```diff
@@ -0,0 +1,63 @@
...
+ * @param systemErr System.err value
+ */
+case class Environment(
+  thread:Thread,
+  contextClassLoader: ClassLoader,
+  systemIn: InputStream,
+  systemOut:PrintStream,
+  systemErr:PrintStream
+) {
+  def this(classLoader: ClassLoader, in: InputStream, out:PrintStream) = this(Thread.currentThread(), classLoader, in, out, out)
```
lihaoyi: Let's use secondary `apply` methods on the companion object, instead of secondary constructors

sshd/src/main/scala/ammonite/repl/sshd/util/Environment.scala
```diff
@@ -0,0 +1,63 @@
...
+
+/**
+ * Container for staging environment important for Ammonite repl to run correctly.
+ * @param thread a thread where execution takes place. Important for restoring contextClassLoader
+ * @param contextClassLoader thread's context class loader. Ammonite repl uses that to load it's classes
+ * @param systemIn System.in value
+ * @param systemOut System.out value
+ * @param systemErr System.err value
+ */
+case class Environment(
```
lihaoyi: The indentation on these argument lists should be consistent with the rest of the code: first argument on same line as open-paren, following args left-aligned, close-paren on same line as last arg

sshd/src/main/scala/ammonite/repl/sshd/SshServer.scala
```diff
@@ -0,0 +1,65 @@
...
+import org.apache.sshd.server.keyprovider.SimpleGeneratorHostKeyProvider
+import org.apache.sshd.server.session.ServerSession
+import org.apache.sshd.{ SshServer ⇒ SshServerImpl }
+
+/**
+ * a factory to produce shell-only ssh server
+ */
+object SshServer {
+  import Implicits._
+  type RemoteShellServer = ((InputStream, OutputStream) ⇒ Unit)
```
lihaoyi: Let's use non-unicode `=>` arrows just to be consistent

sshd/src/main/scala/ammonite/repl/sshd/util/Environment.scala
```diff
@@ -0,0 +1,63 @@
+package ammonite.repl.sshd.util
+
+import java.io.{InputStream, OutputStream, PrintStream}
+
+/**
+ * Container for staging environment important for Ammonite repl to run correctly.
+ * @param thread a thread where execution takes place. Important for restoring contextClassLoader
+ * @param contextClassLoader thread's context class loader. Ammonite repl uses that to load it's classes
+ * @param systemIn System.in value
```
lihaoyi: No need for no-op `@param` descriptions

sshd/src/test/scala/ammonite/repl/sshd/SshTestingTraits.scala
```diff
@@ -0,0 +1,131 @@
...
+import java.io._
+
+import ammonite.ops._
+import com.jcraft.jsch.{Channel, Session, JSch, UserInfo}
+import org.scalacheck.Gen
+
+import scala.annotation.tailrec
+import scala.util.Random
+
+trait SshTestingTraits {
```
lihaoyi: This stuff seems like it should be an `object` and not a `trait`.

build.sbt
```diff
@@ -151,6 +151,24 @@ lazy val repl = project
...
+      libraryDependencies ++= Seq(
+        "org.apache.sshd" % "sshd-core" % "0.14.0",
+        //-- test --//
+        "org.slf4j" % "slf4j-nop" % "1.7.12" % "test",
+        "com.jcraft" % "jsch" % "0.1.53" % "test",
+        "org.scalacheck" %% "scalacheck" % "1.12.4" % "test"
```
lihaoyi: I don't really see the value of ScalaCheck in this test suite. Do we think that "some, but not all, incorrect credentials will pass" is a likely failure mode? It doesn't seem likely to me, but maybe you have some failure modes in mind I haven't thought about

sshd/src/main/scala/ammonite/repl/sshd/ShellSession.scala
```diff
@@ -0,0 +1,63 @@
...
+  var in: InputStream = _
+  var out: OutputStream = _
+  var exit: Option[sshd.server.ExitCallback] = None
+  var thread: Option[Thread] = None
+
+  def setInputStream(in: InputStream) = { this.in = in }
+  def setOutputStream(out: OutputStream) = { this.out = new SshOutputStream(out) }
+  def setErrorStream(err: OutputStream) = { }
+
+  // called by ssh server to instrument this session with a callback that it finished serving a user
```
lihaoyi: All these could be `/** */`-style comments I think

sshd/src/main/scala/ammonite/repl/sshd/ShellSession.scala
```diff
@@ -0,0 +1,63 @@
...
+  def setExitCallback(exit: sshd.server.ExitCallback) { this.exit = Option(exit) }
+
+  // called then ssh server is ready to start this session. Starts the actual shell-serving task.
```
lihaoyi: `then` -> `when`

sshd/src/main/scala/ammonite/repl/sshd/util/Implicits.scala
```diff
@@ -0,0 +1,21 @@
+package ammonite.repl.sshd.util
+
+import org.apache.sshd.common.Factory
+import org.apache.sshd.server.{CommandFactory, Command}
+
+import scala.language.implicitConversions
+
+trait Implicits {
```
lihaoyi: Why do these need to be implicit? I don't see them used anywhere in the example you showed in the readme, and I only see three call-sites inside the codebase. I think we should inline them.

sshd/src/main/scala/ammonite/repl/sshd/ShellSession.scala
```diff
@@ -0,0 +1,63 @@
...
+  def start(env: sshd.server.Environment) {
+    import Implicits._
+    val thread = new Thread({
```
lihaoyi: I don't think this does what you think it does, due to the way implicits work

```scala
scala> implicit def runnable(block: => Unit) = new Runnable{ override def run() = block }
warning: there was one feature warning; re-run with -feature for details
runnable: (block: => Unit)Runnable

scala> {println(1); println(2)}: Runnable
1
res0: Runnable = $anon$1@234bef66

scala> res0.run()
2
```

This one really shouldn't be an implicit, and since it's called only once it probably shouldn't even be a helper

sshd/src/main/scala/ammonite/repl/sshd/SshServerConfig.scala
```diff
@@ -0,0 +1,23 @@
...
+
+/**
+ * Ssh server parameters
+ * @param port a port to be used by ssh server. Set it as `0` to let server choose some random port.
+ * @param username username to authenticate on ssh server
+ * @param password password to authenticate on ssh server
+ * @param ammoniteHome path that ammonite repl sessions will be using as their home directory
+ * @param hostKeyFile path to the place where to store server's identity key
+ */
+case class SshServerConfig(
```
lihaoyi: Formatting: first arg on same line as open-paren, following args left-aligned, close-paren on same line as last arg

sshd/src/main/scala/ammonite/repl/sshd/SshdRepl.scala
```diff
@@ -0,0 +1,78 @@
...
+  private lazy val sshd = SshServer(sshConfig, terminal = SshdRepl.runRepl(sshConfig.ammoniteHome, predef, replArgs))
+
+  def port = sshd.getPort
+  def start():Unit = sshd.start()
+  def stop():Unit = sshd.stop()
+  def stopImmediately():Unit = sshd.stop(/*immediately =*/ true)
+}
+
+object SshdRepl {
+  def main(args:Array[String]):Unit = {
```
lihaoyi: What's the purpose of this `main` method? I don't see it in the example. Is it just for testing? If so, perhaps we could move it to the `test/` folder

build.sbt
```diff
@@ -151,6 +151,24 @@ lazy val repl = project
...
+ */
+lazy val sshd = project.
+    dependsOn(repl).
+    settings(sharedSettings:_*).
+    settings(
+      name := "ammonite-repl-sshd",
+      libraryDependencies ++= Seq(
+        "org.apache.sshd" % "sshd-core" % "0.14.0",
+        //-- test --//
+        "org.slf4j" % "slf4j-nop" % "1.7.12" % "test",
```
lihaoyi: Where are we using this? I couldn't find any use cases, but maybe I just don't know what to grep for

sshd/src/test/scala/ammonite/repl/sshd/SshTestingTraits.scala
```diff
@@ -0,0 +1,131 @@
...
+      new InputStreamReader(pipeEnd) with LineReader
+    }
+
+    def connect() = shell.connect()
+    def isConnected = shell.isConnected
+    def disconnect() = shell.disconnect()
+  }
+
+
+  trait LineReader { reader:Reader ⇒
```
lihaoyi: I only see this being used once. AFAICT this should just be a helper method on a static object

sshd/src/test/scala/ammonite/repl/sshd/SshTestingTraits.scala
```diff
@@ -0,0 +1,131 @@
...
+  def withTmpDirectory[T](block:Path ⇒ T):T = {
+    lazy val tmpDir = Path(Path.makeTmp)
+    try {
+      block(tmpDir)
+    } finally {
+      rm! tmpDir
+    }
+  }
+
+  def testSshServer(user:(String, String), shell: SshServer.RemoteShellServer)(implicit dir:Path) = {
```
lihaoyi: This could be inlined since it's only called once

sshd/src/test/scala/ammonite/repl/sshd/SshTestingTraits.scala
```diff
@@ -0,0 +1,131 @@
...
+    override def promptPassphrase(message: String): Boolean = false
+    override def promptYesNo(message: String): Boolean = true
+
+    override def showMessage(message: String): Unit = { /*noop*/ }
+    override def getPassword: String = password
+    override def getPassphrase: String = ???
+  }
+
+
+  class Shell private(shell:Channel) {
```
lihaoyi: What is this? It could probably use a more descriptive name and a doc-comment. I've jumped around a bit and still don't really understand what role it plays

sshd/src/main/scala/ammonite/repl/sshd/ShellSession.scala
```diff
@@ -0,0 +1,63 @@
+package ammonite.repl.sshd
+
+import java.io.{InputStream, OutputStream}
+
+import ammonite.repl.sshd.util.Implicits
+import org.apache.sshd
+
+/**
+ * Implementation of ssh server's remote shell session, which will be serving remote user.
```
lihaoyi: This isn't a very useful doc-comment, and the `@param` description below is just C&P and serves no purpose. What I want to know: how does this relate to the `SshServer` and `SshRepl`? I suspect a new `ShellSession` is created every time someone SSHs in, but I'm not certain.

sshd/src/test/scala/ammonite/repl/sshd/SshdReplTest.scala
```diff
@@ -0,0 +1,48 @@
...
+        shell.input.println("exit")
+
+        assert(Await.result(remotePromise.future, 60 seconds))
+
+        val outputLines = shell.output.readAll.lines.toList
+        val firstBannerLine :: secondBannerLine :: Nil = outputLines.take(2)
+        val mathTestPrompt :: mathOutput :: Nil = outputLines.slice(2, 4)
+        val importAppPackage :: importResult :: Nil = outputLines.slice(4, 6)
+
+        assert(firstBannerLine.startsWith("Welcome to the Ammonite Repl"))
```
lihaoyi: uTest's `assert` can take multiple expressions as varargs; you can have one `assert` call and just pass each of these expressions on a new line. You also shouldn't need the `&&`s; those can just be separate expressions in the one, large `assert` call

sshd/src/test/scala/ammonite/repl/sshd/SshServerTests.scala
```diff
@@ -0,0 +1,89 @@
...
+          check(forAll { randomChannel:String ⇒
+            cantConnectToChannel(client, randomChannel)
+          })
+        }
+      }
+    }
+  }
+
+  private lazy val rejectedChannelTypes = {
+    "exec" :: "env" :: "x11-req" :: "x11" :: "subsystem" :: "exit-signal" :: "auth-agent-req@openssh.com" :: Nil
```
lihaoyi: Let's convert all of these to the `Seq(...)` constructor

sshd/src/test/scala/ammonite/repl/sshd/SshdReplTest.scala
```diff
@@ -0,0 +1,48 @@
...
+
+        shell.input.println("2 + 2")
+        shell.input.println("import ammonite.repl.sshd.SshdReplTest")
+        shell.input.println("SshdReplTest.remotePromise.success(true)")
+        shell.input.println("exit")
+
+        assert(Await.result(remotePromise.future, 60 seconds))
+
+        val outputLines = shell.output.readAll.lines.toList
+        val firstBannerLine :: secondBannerLine :: Nil = outputLines.take(2)
```
lihaoyi: ... and the `Seq(...)` extractor

sshd/src/main/scala/ammonite/repl/sshd/util/Environment.scala
```diff
@@ -0,0 +1,63 @@
...
+    Thread.currentThread().getContextClassLoader,
+    System.in,
+    System.out,
+    System.err
+  )
+
+  /**
+   * Runs your code with supplied environment installed.
+   * After execution of supplied code block will restore original environment
+   * @param env
```
lihaoyi: No-op doc-comments should be removed

build.sbt
```diff
@@ -151,6 +151,24 @@ lazy val repl = project
 //    fork in (Test, testOnly) := true
   )
 
+
+/**
+ * REPL available via remote ssh access. Pluggable into any app environment to allow live hacking on a live application.
```
lihaoyi: Try to keep your lines below 80 chars wide, and definitely nothing more than 100

sshd/src/main/scala/ammonite/repl/sshd/ShellSession.scala
```diff
@@ -0,0 +1,63 @@
+package ammonite.repl.sshd
+
```
lihaoyi: Every file should have the [Acyclic Plugin](https://github.com/lihaoyi/acyclic) enabled via `import acyclic.file`.

sshd/src/main/scala/ammonite/repl/sshd/ShellSession.scala
```diff
@@ -0,0 +1,63 @@
...
+      this.thread = None
+      exit.foreach(_.onExit(0, "repl finished"))
+    })
+    this.thread = Some(thread)
+    thread.start()
+  }
+
+  // called then ssh server wants to destroy shell session. Whatever shell session serving a user
+  // was doing at this moment we are free to stop it.
+  def destroy() {
```
lihaoyi: No procedure syntax since it's deprecated https://issues.scala-lang.org/browse/SI-7605

sshd/src/test/scala/ammonite/repl/sshd/SshTestingTraits.scala
```diff
@@ -0,0 +1,131 @@
...
+    } finally {
+      rm! tmpDir
+    }
+  }
+
+  def testSshServer(user:(String, String), shell: SshServer.RemoteShellServer)(implicit dir:Path) = {
+    SshServer(SshServerConfig(port = 0, username = user._1, password = user._2, ammoniteHome = dir), shell)
+  }
+
+  def withTestSshServer[T](user:(String, String), shell: SshServer.RemoteShellServer)
```
lihaoyi: As far as I can tell, `SshServer.RemoteShellServer` never gets its callback arguments ever used. This should take a `Function0` or a by-name param

sshd/src/test/scala/ammonite/repl/sshd/SshServerTests.scala
```diff
@@ -0,0 +1,89 @@
...
+import scala.concurrent.{Await, Promise}
+import scala.language.postfixOps
+
+object SshServerTests extends TestSuite with ScalaCheckSupport with SshTestingTraits {
+  override val tests = TestSuite {
+    'canConnectAndAuthenticate{
+      withTmpDirectory { implicit tmpDir ⇒
+        check {
+          forAll(genCreds) { input ⇒
+            assert(canAuthenticate(input))
```
lihaoyi: This is only called once, and probably should be in-lined

sshd/src/test/scala/ammonite/repl/sshd/SshServerTests.scala
```diff
@@ -0,0 +1,89 @@
...
+import utest.framework.TestSuite
+
+import scala.concurrent.duration._
+import scala.concurrent.{Await, Promise}
+import scala.language.postfixOps
+
+object SshServerTests extends TestSuite with ScalaCheckSupport with SshTestingTraits {
+  override val tests = TestSuite {
+    'canConnectAndAuthenticate{
+      withTmpDirectory { implicit tmpDir ⇒
```
lihaoyi: `withTmpDirectory` can probably be combined with `withTestSshServer`; there are no cases where one is called but not the other AFAICT, so we can avoid the boilerplate around every test case.

sshd/src/main/scala/ammonite/repl/sshd/ShellSession.scala
```diff
@@ -0,0 +1,63 @@
...
+
+
+  // proxy which fixes output to the remote side to be ssh compatible.
+  private class SshOutputStream(out:OutputStream) extends OutputStream {
+    override def close() { out.close() }
+    override def flush() { out.flush() }
+
+    override def write(b: Int) {
+      // ssh only accepts new lines with \r
+      if (b.toChar == '\n') out.write('\r')
```
lihaoyi: If this code encounters a `...\n\r...` in the bytes being written, it will output `...\n\r\r...`. Is that what we want?

sshd/src/main/scala/ammonite/repl/sshd/SshdRepl.scala
```diff
@@ -0,0 +1,78 @@
...
+ */
+class SshdRepl(sshConfig:SshServerConfig,
+               predef: String = "",
+               replArgs: Seq[Bind[_]] = Nil) {
+  private lazy val sshd = SshServer(sshConfig, terminal = SshdRepl.runRepl(sshConfig.ammoniteHome, predef, replArgs))
+
+  def port = sshd.getPort
+  def start():Unit = sshd.start()
+  def stop():Unit = sshd.stop()
+  def stopImmediately():Unit = sshd.stop(/*immediately =*/ true)
```
lihaoyi: No need for the comment; I can instantly jump to the implementation of `stop` if I want to know what `true` means

sshd/src/main/scala/ammonite/repl/sshd/ShellSession.scala
```diff
@@ -0,0 +1,63 @@
...
+ */
+private[sshd] class ShellSession(remoteShell: SshServer.RemoteShellServer) extends sshd.server.Command {
+  var in: InputStream = _
+  var out: OutputStream = _
+  var exit: Option[sshd.server.ExitCallback] = None
+  var thread: Option[Thread] = None
+
+  def setInputStream(in: InputStream) = { this.in = in }
+  def setOutputStream(out: OutputStream) = { this.out = new SshOutputStream(out) }
+  def setErrorStream(err: OutputStream) = { }
```
lihaoyi: This deserves a comment as to why it's empty

sshd/src/main/scala/ammonite/repl/sshd/ShellSession.scala
```diff
@@ -0,0 +1,63 @@
...
+
+import ammonite.repl.sshd.util.Implicits
+import org.apache.sshd
+
+/**
+ * Implementation of ssh server's remote shell session, which will be serving remote user.
+ * @param remoteShell actual shell implementation, which will serving remote user's shell session.
+ */
+private[sshd] class ShellSession(remoteShell: SshServer.RemoteShellServer) extends sshd.server.Command {
+  var in: InputStream = _
```
lihaoyi: Is there a reason some of these are `null` and some of them `None`?

sshd/src/main/scala/ammonite/repl/sshd/SshServer.scala
```diff
@@ -0,0 +1,65 @@
...
+
+  def touch(file:Path):Path = {
+    import ammonite.ops._
+    if (!exists(file)) {
+      write(file, Array.empty[Byte])
+    }
+    file
+  }
+
+  private def passwordAuthenticator(credentials:(String, String)):PasswordAuthenticator = new PasswordAuthenticator {
```
lihaoyi: This should just take two arguments instead of a tuple

sshd/src/main/scala/ammonite/repl/sshd/SshServer.scala
```diff
@@ -0,0 +1,65 @@
...
+import org.apache.sshd.common.session.ConnectionService
+import org.apache.sshd.common.{FactoryManager, Channel, NamedFactory, Session}
+import org.apache.sshd.common.file.{FileSystemView, FileSystemFactory}
+import org.apache.sshd.server.PasswordAuthenticator
+import org.apache.sshd.server.keyprovider.SimpleGeneratorHostKeyProvider
+import org.apache.sshd.server.session.ServerSession
+import org.apache.sshd.{ SshServer ⇒ SshServerImpl }
+
+/**
+ * a factory to produce shell-only ssh server
```
lihaoyi: A more informative comment might be "A factory to construct `org.apache.sshd.SshServer`s with a nicer interface". As far as I can tell, this has nothing to do with shell-only or not-shell-only: we're just passing in `{Input, Output}-Stream`s.

Also, `shell-only` is kind of an odd phrase that doesn't really meaning anything to me. `Ammonite-REPL ssh server` would be more descriptive

lihaoyi:
That's all for now. The test failure looks spurious so that's fine. 

In general my feedback would be:

- Standardize formatting (space after `:`, multi-line arg-lists)
- Doc-comments in terms of "what you might not know after reading the code" rather than "here's a fact"; no boilerplate doc-comments
- Simplification: don't use a trait when you can use an object, don't use a function when you only call the thing once ever (just inline it) especially not an implicit one

This is very exciting; thanks for bearing with me =)

readme/Index.scalatex
dant3: It doesn't supports key-based auth currently. I've seen it as a too much work for a beginning, but of course it could be added later.

SSH always asks for password in case there is no key-based auth, but you are free to left password blank if you wish. There is no other way to make passwordless auth other than enabling key verification AFAIK.

WRT port if you have it opened everybody will be able to access it. My personal recommendation would be to use non-standard port for a couple of reasons:
* port 22 most likely already used by system's sshd
* port 22 may be reserved to be opened only with root access
* it's not a very usual sshd shell access so might be a little unexpected to be found at port 22
* very basic protection from potential attacks - if nobody know which port to attack there are less chances it will be attacked

build.sbt
dant3: ScalaCheck really helped me to find some interesting properties of my initial implementation of auth which I wasn't aware of. I think I would never find them if I would use only predefined testcases. There is no notion of "some incorrect credentials will pass", test clearly states that "for all different credential pairs assert that authentication fails".

In general having ScalaCheck-based tests run against some predefined table is a quite popular pattern, I don't see a problem combining this approaches. ScalaTest even have some special support for such testing.

sshd/src/main/scala/ammonite/repl/sshd/SshdRepl.scala
dant3: This was used for manual standalone testing. I'll move it to the `test` directory.

build.sbt
dant3: Apache sshd depends on `slf4j-api` for logging, which will init default console appender if it will not find anything else at runtime. By adding this dependency this log is redirected to the void so that we have readable console output.

sshd/src/main/scala/ammonite/repl/sshd/ShellSession.scala
dant3: It it will encounter a '...\n\r...' it will output '...\r\n\r...'. The output might be bloated otherwise (depends on client as well).

dant3:
Thanks. I'll fix those, thanks for a sharp eye.

readme/Index.scalatex
lihaoyi: Ok. Let's put some of this down in the documentation =)

build.sbt
lihaoyi: Ok SGTM

build.sbt
lihaoyi: Ok. Let's have a comment here just explaining that =)

sshd/src/test/scala/ammonite/repl/sshd/SshServerTests.scala
dant3: The reasoning behind this is to have one directory used for all runs performed by ScalaCheck. Tests will generate big IO creating & removing files for every ScalaCheck run, and I don't currently see difference between having dedicated directory for every run and having one directory to run all tests.

lihaoyi: Ok. I'll probably take a crack at some refactors after you land it, but this'll do for now

lihaoyi:
FWIW I'm doing a large-ish refactor, so you may have some merge conflicts, though it shouldn't be a hard merge.

dant3:
The build is green now \o/
Additional work on readme and scaladocs is required, but code issues hopefully fixed now.

lihaoyi:
Looks great, thanks! =D This'll go out next release

dant3:
\o/ Looking forward \o/

PR #182, opened 2015-08-20
mention import in spawning process docs

alekar:


lihaoyi:
Thanks! Will go out next release which should be in a few days

PR #184, opened 2015-08-20
Add SBT plugin

alexarchambault:
No unit tests yet, maybe scripted could help for that.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -165,6 +165,16 @@ lazy val repl = project
 //    fork in (Test, testOnly) := true
   )
 
+lazy val sbtPlugin = project.in(file("sbt-plugin"))
+  .settings(acyclicSettings)
+  .settings(publishSettings)
+  .settings(
+    // not specifying a scala version, should default to the one of SBT, which is what we want
+    name := "ammonite-sbt",
+    sbt.Keys.sbtPlugin := true,
+    (unmanagedSources in Compile) += baseDirectory.value/".."/"project"/"Constants.scala"
+  )
+
 lazy val readme = ScalatexReadme(
   projectId = "readme",
   wd = file(""),
--- a/build.sbt
+++ b/build.sbt
@@ -1,28 +1,16 @@
 import scalatex.ScalatexReadme
 import sbtassembly.AssemblyPlugin.defaultShellScript
 
-scalaVersion := "2.11.6"
-
-crossScalaVersions := Seq(
-  "2.10.3", "2.10.4", "2.10.5", "2.11.3", "2.11.4", "2.11.5", "2.11.6", "2.11.7"
-)
-
-publishArtifact := false
-
-publishTo := Some(Resolver.file("Unused transient repository", file("target/unusedrepo")))
-
-val sharedSettings = Seq(
+lazy val sharedSettings = Seq(
   scalaVersion := "2.11.7",
-  organization := "com.lihaoyi",
-  version := _root_.ammonite.Constants.version,
+  crossScalaVersions := Seq(
+    "2.10.3", "2.10.4", "2.10.5", "2.11.3", "2.11.4", "2.11.5", "2.11.6", "2.11.7"
+  ),
   libraryDependencies += "com.lihaoyi" %% "utest" % "0.3.0" % "test",
   testFrameworks += new TestFramework("utest.runner.Framework"),
   scalacOptions += "-target:jvm-1.7",
-  autoCompilerPlugins := true,
-  addCompilerPlugin("com.lihaoyi" %% "acyclic" % "0.1.2"),
   parallelExecution in Test := !scalaVersion.value.contains("2.10"),
   libraryDependencies ++= Seq(
-    "com.lihaoyi" %% "acyclic" % "0.1.2" % "provided",
     "org.scala-lang" % "scala-reflect" % scalaVersion.value % "provided"
   ) ++ (
     if (scalaVersion.value startsWith "2.11.") Nil
@@ -30,8 +18,20 @@ val sharedSettings = Seq(
       compilerPlugin("org.scalamacros" % s"paradise" % "2.0.1" cross CrossVersion.full),
       "org.scalamacros" %% s"quasiquotes" % "2.0.1"
     )
-  ),
+  )
+)
+
+lazy val acyclicSettings = Seq(
+  autoCompilerPlugins := true,
+  addCompilerPlugin("com.lihaoyi" %% "acyclic" % "0.1.2"),
+  libraryDependencies += "com.lihaoyi" %% "acyclic" % "0.1.2" % "provided"
+)
+
+lazy val publishSettings = Seq(
+  organization := "com.lihaoyi",
+  version := _root_.ammonite.Constants.version,
   publishTo := Some("releases" at "https://oss.sonatype.org/service/local/staging/deploy/maven2"),
+  pomIncludeRepository := { _ => false },
   pomExtra :=
     <url>https://github.com/lihaoyi/Ammonite</url>
       <licenses>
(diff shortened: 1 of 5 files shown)


alexarchambault:
Build was broken before the PR, but it looks like something like sbt-doge should be used to disable the sbt-plugin project if the scala version is wrong. I'll try as soon as I can reach my laptop.

lihaoyi:
yeah, i broke it, fixing...

lihaoyi:
master _should_ be green soon i think

alexarchambault:
Travis CI error looks like a random failure.

alexarchambault:
Note that the third commit (https://github.com/alexarchambault/Ammonite/commit/ab58374a637fac4507ea428483a21d95ed82eb85) *replaces* the current SBT explanation by the use of the SBT plugin.

lihaoyi:
Quick question: how hard would it be to have this be a task rather than a scope? So I can hit e.g. 

```
sbt myproject/repl
sbt myproject/test:repl
```

That would make more sense to me than duplicating `run`/`test` with `ammonite`/`ammonite-test`. OTOH I don't know SBT so I don't know if what I'm asking is impossible

alexarchambault:
This should be feasible (and straightforward), at least for the first one, in addition to the scopes (configurations actually). I just hope I won't run into https://github.com/alexarchambault/sbt-ammonite/issues/1, which I didn't understand.

Using scripted to unit test that seems easier than what I initially though too, I'll give it a try.

alexarchambault:
About unit tests for this: scripted (examples [here](https://github.com/xerial/sbt-pack/tree/master/src/sbt-test/sbt-pack) or [there](https://github.com/sbt/sbt-doge/tree/master/src/sbt-test/sbt-doge)) can test the return code of commands it runs. How I'd like to test this is by having various projects, with things in their main and test sources, and code that would rely on that in `initialCommands` (provided as predef by the plugin). This code would exit with success or failure codes, depending on what was available to it, testing what was put in its classpath this way (main or test code, dependencies, ...).

For that, I would need to be able to exit the Ammonite REPL programatically from the predef, specifying the exit code of the process at the same time. This, in turn, would/could require:
- using the value of the `Exit` ([here](https://github.com/lihaoyi/Ammonite/blob/e8454a7b5bc68d43a6f8feb0da9cae24d5ec9bbd/repl/src/main/scala/ammonite/repl/Util.scala#L41)) as the exit code (giving it to `sys.exit`),
- allow the predef to exit this way (it seemed to me that it doesn't really handle exits).

I'm new to `load.module` that predef seems to use, so I don't know if the second point is straightforward.

Else, it wouldn't be that much a problem for me not to unit test the plugin, but then I think I'd prefer to to recommend it as the default way of using Ammonite from SBT (as I did in the last commit of the PR). Unless some expert SBT eyes can validate what the plugin does.

alexarchambault:
I tried to make this a task, [here](https://github.com/alexarchambault/Ammonite/tree/topic/sbt-plugin-v3), but ran into the issue I was seeing before adding the scope/configuration (it's as if the underlying task is run multiple times in parallel - really don't know what's going on there).

readme.md
```diff
@@ -20,14 +20,13 @@ The layout of the repository is roughly:
...
-- `modules/` is a synthetic project used for publishing, purely intended to exclude the readme
 
 ## Common Commands
 
 - `sbt ~repl/run` brings up the Ammonite-REPL using the source code in the repository, and automatically restarts it on-exit if you have made a change to the code. Useful for manual testing both of `repl/` as well as `ops/`, since you can just `import ammonite.ops._` and start using them.
 - `sbt ~terminal/run` is useful for manual testing the terminal interaction; it basically contains a minimal echo-anything terminal, with multiline input based on the count of open- and closed-parentheses. This lets you test all terminal interactions without all the complexity of the Scala compiler, classloaders, etc. that comes in `repl/`
 - `sbt ~repl/test`/`sbt ~ops/test`/`sbt ~terminal/test`  runs tests after every change. `repl/test` can be a bit slow because of the amount of code it compiles, so you may want to specify the test manually via `repl/test-only -- ammonite.repl.TestObject.path.to.test`
-- `sbt +modules/publishLocal` or `+sbt modules/publishSigned` is used for publishing.
+- `sbt +publishLocal` or `+sbt publishSigned` is used for publishing.
```
lihaoyi: `sbt publishLocal` doesn't seem to publish the SBT plugin; I needed to do `sbt sbtPlugin/publishLocal` to make that work.

readme/Index.scalatex
```diff
@@ -56,22 +56,27 @@
...
   @p
-    You can also pass a string to the @code{run} call containing any commands or imports you want executed at the start of every run. If you want Ammonite to be available in all projects, simply add the above snippet to a new file @code{~/.sbt/0.13/global.sbt}.
+    Hit
+  @code
+    sbt projectName/ammonite-test:run
+  @p
+    To activate the Ammonite REPL with the test build products and dependencies.
+
+  @p
+    You can also give @code{initialCommands in console} any commands or imports you want executed at the start of every run. If you want Ammonite to be available in all projects, simply add the above snippet to @code{~/.sbt/0.13/plugins/build.sbt}.
```
lihaoyi: Woohoo

readme/Index.scalatex
```diff
@@ -56,22 +56,27 @@
...
-    sbt projectName/test:console
+    sbt projectName/ammonite:run
   @p
     To activate the Ammonite REPL
 
   @p
-    You can also pass a string to the @code{run} call containing any commands or imports you want executed at the start of every run. If you want Ammonite to be available in all projects, simply add the above snippet to a new file @code{~/.sbt/0.13/global.sbt}.
+    Hit
+  @code
+    sbt projectName/ammonite-test:run
```
lihaoyi: Is it all possible to make `ammonite:run` and `ammonite:test` do the right thing? If not that's ok, but that certainly sounds like a better place to be in if we can make it happen

lihaoyi:
w.r.t. unit testing, here's another idea: the Ammonite executable can already be used to run files via `./amm script.scala`. I don't know what the current status of the plugin is, but it makes sense to be able to run script files via `sbt proj/ammonite:run script.scala`. Then you can have a demo project where you run it in various configurations (run, test) on various scripts (scripts-that-use-run-classes, scripts-that-use-test-classes) and see how it behaves.

I don't know if running scripts via `./amm script.scala` does the right thing w.r.t. return codes. If it doesn't, we can fix it. An alternative is we can just not use sbt-scripted and do our own shelling out to a separate SBT process in the demo-project's folder, and see what gets printed to stdout/stderr

build.sbt
```diff
@@ -151,6 +165,16 @@ lazy val repl = project
...
   )
 
+lazy val sbtPlugin = project.in(file("sbt-plugin"))
+  .settings(acyclicSettings)
+  .settings(publishSettings)
+  .settings(
+    // not specifying a scala version, should default to the one of SBT, which is what we want
+    name := "ammonite-sbt",
+    sbt.Keys.sbtPlugin := true,
+    (unmanagedSources in Compile) += baseDirectory.value/".."/"project"/"Constants.scala"
```
lihaoyi: Since we now have three of these, let's collapse them into a `val`

lihaoyi:
This looks good. I still think the ideal user interface would be `sbt repl` or `sbt test:repl`, to match `console`'s behavior. If you can't figure it out I may take a crack at it; I think having the CLI be consistent with the `console` command everyone already knows is worth the extra delay/effort/frustration

alexarchambault:
I proposed to unit test that, but I now think I won't have enough time to put into that though.

About having a `repl` command, it's easy to try, see how it's done [here](https://github.com/alexarchambault/sbt-ammonite/blob/da95c3809bd7b91949e1ce332ec0f1b6b310c31b/src/main/scala/ammonite/AmmonitePlugin.scala#L46-L53) (it's actually simpler than that, one has just to call `run in Ammonite` in the `repl` task instead of all that's done in the previous link.) On my config, it feels as if two Ammonite are run, so it just doesn't work. I don't really know why. It may have to see with the way Ammonite handles the terminal now.

If ever you're fine with anything from this PR, just cherry-pick it (maybe not the doc in a first time, to be cautious).

lihaoyi:
No worries, when I have time I'll commandeer this and get it in. Thanks for
putting up so far =D

lihaoyi:
My cherry picking has so far failed; I'm not sure whether I'm merging correctly, because the changes you made to build.sbt conflicts all over the place =/ But no matter what I do I haven't managed to successfully publishLocal -> ammonite:run

alexarchambault:
Closing this, feel free to ping me or pick some code from this if you want this feature to land here.

Issue #149, opened 2015-08-09
Can't load modules from modules


lihaoyi:
This doesn't seem to behave correctly in the following case:

```scala
// LimitImports.scala
val res = 1
@
val asd = 1
```
```scala
// Foo.scala
import ammonite.ops._

load.module(wd/"LimitImports.scala")

@

println(res)
println(asd)
```
```scala
@ // Repl Session
@ import ammonite.ops._
@ load.module(wd/"Foo.scala")
ammonite.repl.CompilationError: Compilation Failed
Main.scala:174: not found: value res
println(res)
        ^
Main.scala:177: not found: value asd
println(asd)
        ^
	ammonite.repl.interp.Interpreter.loop$1(Interpreter.scala:97)
	ammonite.repl.interp.Interpreter.processScript(Interpreter.scala:86)
	ammonite.repl.interp.Interpreter.processModule(Interpreter.scala:67)
	ammonite.repl.interp.Interpreter$$anon$1$load$.module(Interpreter.scala:134)
	cmd4$$anonfun$1.apply$mcV$sp(Main.scala:110)
	cmd4$.<init>(Main.scala:111)
	cmd4$.<clinit>(Main.scala:-1)
```

This should work. Will take some refactoring to make `load.module` put the imports in the right place depending on whether it's being called in a module or directly from the REPL, but shouldn't be hard in principle

lihaoyi:
@laci37 this is probably on you when you come back, since you own much of the module/script loading stuff =)

dynofu:
it's broken again

```
// Foo.scala
import ammonite.ops._
interp.load.module(pwd/"LimitImports.sc")

@
import ammonite.$file.dummy.LimitImports._
println(res)
import ammonite.$file.dummy.LimitImports_2._
println(asd)
```

this makes writing script compatible with different versions through `interp.load.module` really hard. 
as old version e.g. 1.6.7 does not support `import ammonite.$file.dummy.LimitImports_2._` 
need 1.6.7 to support scala 2.11 and spark.

dynofu:
1.6.7 does support `import ammonite.$file.dummy.LimitImports_2._` as long as you clear the cache in the start, `rm -rf ~/.ammonite/cache`

PR #189, opened 2015-08-22
Fix loading scripts from other scripts

laci37:
This fixes issue #149.

### DIFF
--- a/repl/src/main/scala/ammonite/repl/interp/Interpreter.scala
+++ b/repl/src/main/scala/ammonite/repl/interp/Interpreter.scala
@@ -2,6 +2,7 @@ package ammonite.repl.interp
 
 import java.io.File
 import java.nio.file.NotDirectoryException
+import scala.collection.mutable
 import acyclic.file
 import ammonite.ops._
 import pprint.{Config, PPrint}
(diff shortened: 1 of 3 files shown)


repl/src/main/scala/ammonite/repl/interp/Interpreter.scala
```diff
@@ -86,16 +91,24 @@ class Interpreter(prompt0: Ref[String],
...
-          case Res.Success(ev) => loop(blocks.tail, imports :+ ev.imports)
+          case Res.Success(ev) => loop(blocks.tail, imports :+ (ev.imports ++ nestedScriptImports))
           case _ => loop(blocks.tail, imports)
         }
-      } else imports.lastOption.foreach(eval.update(_))
+      } else {
+        //if we have imports to pass to the upper layer we do that
+        imports.lastOption.foreach(outerScriptImportCallback)
+        //restore the old value
+        scriptImportCallback = outerScriptImportCallback
```
lihaoyi: Could we put this in a `finally` block? We want it to properly reset in the case of an unhandled exception

lihaoyi:
One small nit, but looks good

lihaoyi:
Looks great, thanks!

PR #190, opened 2015-08-22
Disable Ctrl-y wherever possible (fix #180)

### LINKED ISSUE #180: C-y kills Ammonite
Most Scala REPLs (including Ammonite right now) tend* to die after I type C-y (that is, Ctrl-y) and resume, instead of pasting what I just C-k'ed, with an InterruptedException on reading.

I tracked this down to C-y causing (by default) a [delayed suspend](https://www.gnu.org/software/bash/manual/bashref.html#Job-Control), that is a suspend when the process reads from input (like C-z). This can be disabled through `stty dsusp undef`, also used in JLine, but that's commented out with a "blows up on ubuntu" comment. Alternatively, one could catch the `InterruptedException`, match on the error message and ignore the exceptions with the right message (although that's probably not extremely robust) — you often don't want to ignore all `InterruptedException`.

*That is, they repeatably die until this bug is somewhen fixed — both SBT and `scala` had it at some point.

Blaisorblade:
This PR simply reenables the code disabling Ctrl-y, while ensuring it does no harm where Ctrl-y is unsupported (Linux). After this commit, Ctrl-y works as expected in Emacs/bash/readline: it yanks/pastes text killed/cut with Ctrl-k and similar combos.

Fixes #180.

### DIFF
--- a/terminal/src/main/scala/ammonite/terminal/Utils.scala
+++ b/terminal/src/main/scala/ammonite/terminal/Utils.scala
@@ -63,7 +63,6 @@ object TTY{
 //    Debug("Initializing, Height " + height)
     val initialConfig = stty("-g").trim
     stty("-icanon min 1 -icrnl -inlcr -ixon")
-    // this command fails on ubuntu, but that is harmless.
     sttyFailTolerant("dsusp undef")
     stty("-echo")
     stty("intr undef")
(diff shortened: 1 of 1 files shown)


lihaoyi:
Looks great, thanks!

PR #193, opened 2015-08-23
Sort output of ls! (fix #192)

### LINKED ISSUE #192: `ls!` does not sort output on Linux (unlike expected by tests)
On Linux, [a test](https://github.com/lihaoyi/Ammonite/blob/e8454a7b5bc68d43a6f8feb0da9cae24d5ec9bbd/repl/src/test/scala/ammonite/repl/session/AdvancedTests.scala#L48-L60) (the same I mentioned in #191) fails because the output of `ls!` is not sorted like the test expects. If the test is correct, this seems an actual bug.
```
@ import ammonite.ops._
import ammonite.ops._
@ ls! wd/'ops
res1: LsSeq = LsSeq(
  'target,
  'src
)
@ %%ls 'ops
res2: CommandResult =
src
target
```

Blaisorblade:
Straightforward fix (the whitespace change slipped in). Do you agree with sorting `RelPath` by `ups` first, then by `segments`?

### DIFF
--- a/ops/src/main/scala/ammonite/ops/FileOps.scala
+++ b/ops/src/main/scala/ammonite/ops/FileOps.scala
@@ -222,7 +222,7 @@ trait ImplicitOp[V] extends Function1[Path, V]{
  * List the files and folders in a directory
  */
 object ls extends StreamableOp1[Path, Path, LsSeq] with ImplicitOp[LsSeq]{
-  def materialize(src: Path, i: Iterator[Path]) = new LsSeq(src, i.toStream.map(_ - src).sorted:_*)
+  def materialize(src: Path, i: Iterator[Path]) = new LsSeq(src, i.toStream.map(_ - src).sorted(RelPath.alphaOrdering):_*)
   def !!(arg: Path): SelfClosingIterator[Path] = {
     import scala.collection.JavaConverters._
     val dirStream = Files.newDirectoryStream(arg.nio)
--- a/ops/src/main/scala/ammonite/ops/FileOps.scala
+++ b/ops/src/main/scala/ammonite/ops/FileOps.scala
@@ -222,7 +222,7 @@ trait ImplicitOp[V] extends Function1[Path, V]{
  * List the files and folders in a directory
  */
 object ls extends StreamableOp1[Path, Path, LsSeq] with ImplicitOp[LsSeq]{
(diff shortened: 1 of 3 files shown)


Blaisorblade:
I tested this on the offending box.

Ammonite:
```scala
@ ls! 'ops
res1: LsSeq = LsSeq(
  'src,
  'target
)
```

bash:
```sh
$ ls -fF ops
target/  ./  src/  ../
$ ls -aF ops
./  ../  src/  target/
$ ls -fF
terminal/  readme/  project/  modules/  readme.md  tools/  target/  .travis.yml  ./  ops/  build.sbt  .gitignore  repl/  ../  .git/
$ ls -aF
./  ../  build.sbt  .git/  .gitignore  modules/  ops/  project/  readme/  readme.md  repl/  target/  terminal/  tools/  .travis.yml
```
(`ls -f` returns unsorted output on Linux and OS X, `-F` makes monochrome output more readable through trailing slashes).

lihaoyi:
Looks great! Two nits:

- Could we add an `Ordering[Path]` too, while you're at it? Doesn't make sense that `RelPath` is sortable but `Path` isn't
- Let's have at least one unit test for sorting an intentionally-unsorted `RelPath` and `Path` =)

Blaisorblade:
Both are good points.
I just realized you have a different partial order in BasePathImpl through the <= and >= operators. Having them behave differently would be _worse_, so I'll make alphabetic order non-implicit.

I might also rephrase the existing operators in terms of `PartialOrdering`, if they're a semantic fit.

Blaisorblade:
Changes done.

> I might also rephrase the existing operators in terms of PartialOrdering, if they're a semantic fit.

Not for today (or for this PR). I'm coding in vacation mode ;-)

Blaisorblade:
The failure (https://travis-ci.org/lihaoyi/Ammonite/jobs/76871968) appears fully spurious :smile:

Blaisorblade:
> I might also rephrase the existing operators in terms of PartialOrdering, if they're a semantic fit.

Done in #194.

lihaoyi:
Substituted by https://github.com/lihaoyi/Ammonite/commit/6f918e07fc5152e4bab4399aed93a9268d18b059 I think

PR #196, opened 2015-08-24
- Remove ordered path operators for reals, replace with `startsWith`

lihaoyi:
- Remove redundant tests

@Blaisorblade what do you think? It matches the behavior of http://docs.oracle.com/javase/7/docs/api/java/nio/file/Path.html#startsWith(java.nio.file.Path) now

### DIFF
--- a/ops/src/main/scala/ammonite/ops/FileOps.scala
+++ b/ops/src/main/scala/ammonite/ops/FileOps.scala
@@ -154,7 +154,7 @@ object mv extends Function2[Path, Path, Unit] with Internals.Mover{
 object cp extends Function2[Path, Path, Unit] {
   def apply(from: Path, to: Path) = {
     def copyOne(p: Path) = {
-      Files.copy(Paths.get(p.toString), Paths.get((to/(p - from)).toString))
+      Files.copy(Paths.get(p.toString), Paths.get((to/(p relativeTo from)).toString))
     }
 
     copyOne(from)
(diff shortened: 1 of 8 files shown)


lihaoyi:
I'm ambivalent about having to use long names for these things. Using words like `startsWith` or `relativeTo` definitely has a cost over using `>` or `-`

Blaisorblade:
From a quick look:
1. Your new `RelPath.startsWith` looks much better! That should definitely go in.
2. Instead of trying to shorten every method, can we get better Tab completion? The only real problem I see is the delay, but maybe that can be improved?
I'll agree that some things are common enough to warrant operators, maybe non-standard ones, or ones that you get with extra imports; but I can't really judge frequency (except for things matching some `bash` builtin).
3. I'm not against operators per se, but satisfying the principle of least surprise when reusing _standard_ operator is very hard, because operators carry more expectations.
Implementing some type class is a nice way to make sure you conform to (some of) those expectations... But PartialOrdering gets little support from the standard library, and of course, few have seen a partial ordering based on `startsWith`, but that is just unfamiliarity.

Even if `-` is not part of any type class (except `Number`?), I'd still expect that `a - b == c <=> a = c + b` (where `<=>` stands for logical equivalence) while before this patch you only got

`a - b == c` <=> `a = b / c` which has the wrong order. When I noticed this I didn't file a bug, but it's a bit annoying.

Inventing non-standard operators has other problems — people have no intuition of what it should mean, so they must learn it up-front.

PR #219, opened 2015-09-09
Fix for multiline blocks separated by line breaks 

vmarquez:
Previous version was just looking for CR.  This meant when pasting a multiline block without {} from Vim, an error would be thrown.  Now we will look for both CR and newline.


lihaoyi:
Looks good. Probably not strictly correct, but it'll do for now...

Issue #220, opened 2015-09-10
allow bash commands to take arbitrary sequences of arguments


yakticus:
Would like to be able to mix and match single arguments and lists of arguments when calling a bash command:

```
val thing: String = ???
val listOfThings: List[String] = ???
val anotherThing: String = ???
%cmd(thing, listOfThings, anotherThing)
```

yakticus:
@lihaoyi says: `I was thinking of making Shellable into Shellable(parts: Seq[String])`

PR #225, opened 2015-09-10
fix for #220

yakticus:
Shellable now takes Seq[String] to allow for programmatically building up command-line args


lihaoyi:
Looks good, thanks!

PR #231, opened 2015-09-14
Remove "Main.scala: xx" from compilation failure reports in repl

lastland:
The "Main.scala: xx" from the compilation failure reports do not help diagnose the problem. Further, it could cause confusion to beginners.

For example, the following code:

    @ a.foreach(println(_))

will result in:

    Compilation Failed
    Main.scala:33: not found: value a
    a.foreach(println(_))
    ^

This PR changes it to:

    Compilation Failed
    not found: value a
    a.foreach(println(_))
    ^

### DIFF
--- a/readme/Repl.scalatex
+++ b/readme/Repl.scalatex
@@ -216,10 +216,10 @@
     @hl.scala
       @@ mutable.Seq(x) // doesn't work
       Compilation Failed
-      Main.scala:122: not found: value mutable
+      not found: value mutable
       mutable.Seq(x) // doesn't work
       ^
-      Main.scala:122: not found: value x
+      not found: value x
       mutable.Seq(x) // doesn't work
                   ^
       @@ import ammonite.ops._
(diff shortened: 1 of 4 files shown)


lihaoyi:
Will loo at it when i get back from work, thanks!

lihaoyi:
What if the output has multiple lines, and one of it has a non-`Main.scala` colon `:` in it? Won't this incorrectly chop it off in such a case?

lastland:
You are right. Thanks for pointing it out!

I've made some updates on the code, together with a corresponding test case.

lihaoyi:
Sorry for the long delay! Was busy at work -.-

This doesn't seem to behave properly with multiple lines with errors

```scala
haoyi-Ammonite@ {
                asd
                ads
                }
Compilation Failed
not found: value asd
asd
^
Main.scala:124: not found: value ads
ads
^
```

I'm honestly doubtful that this string-splicing approach to removing this is going to work and be robust overall, even by Ammonite's low standards =/ e.g. we can't just regex Main.scala because that may be part of the legitimate output if the line with an error starts with the tail end of a triple-quoted string. Perhaps some kind of compiler plugin that intercepts these errors before they get stringified would be more successful.

lastland:
No worries :smile: 

Yeah you make a very good point here. However, do you thinking renaming "Main.scala" to some special name that would be unlikely to appear (for example, `__Main__.scala`?) would be a possible solution?

PR #232, opened 2015-09-14
load(scalaCode: String, visible: Boolean = false) REPL command addition

abbourne:
My first pull request ever... so there's a chance I've screwed something up!
As per our gitter conversation, this pull request contains 2 changes:

Modifications the load.apply method in Interpreter.scala file to add support for evaluating code and getting a full printout, just as if the code was entered in the REPL itself. 

Here's an example of it in use:
```
@: load("""val foo = "I am cow!" """)

@: foo
res19: String = "I am cow!"
@: load("""val bar = 42; val baz = 2; bar*baz""", true)
bar: Int = 42
baz: Int = 2
res21_2: Int = 84
@: res21_2
res22: Int = 84
@:
```

I've tried all kinds of code, def's, multiple lines of input... it all seems to work fine.
I've included a small update to the Script Files section of the Repl.scalatex file to document the new REPL command.

A small modification to the Compiler.scala file to print error message if Jar dependencies cannot be found. It *only* prints if there is a fatal error, otherwise there is no additional output. This allowed me debug my problem with a Jar dependency issue, and I think it will help other people as well. The change is in the `def display(pos: Position, msg: String, severity: Severity) ` and its exactly the change from the diff file in lihaoyi/Ammonite#168

P.S. I too am a fan of the Arrogant Worms!

### DIFF
--- a/readme/Repl.scalatex
+++ b/readme/Repl.scalatex
@@ -263,6 +263,32 @@
     @p
       All types, values and imports defined in scripts are available to commands entered in REPL after loading the script.
 
+     @p
+      Ammonite also provides a function to load a string of Scala code as if it were entered from the REPL. It comes with an option to load the code silently, or to get a printout of the results in the REPL exactly as if it was entered in the REPL:
+
+      @@
+
+      // scalaCode can be multiple lines, define functions, etc
+      // visible is an optional parameter that defaults to false (no output)
+      load(scalaCode: String, visible: Boolean = false)
+ 
+     @p
+      Here's an example:
+
+      @@
+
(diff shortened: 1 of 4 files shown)


lihaoyi:
Thanks! I will take a look when i get home

lihaoyi:
This is really hard to review. Could you revert all the meaningless syntax reformating, so I can just see the important bits??

lihaoyi:
This'll also need to be rebased on master for CI to run

abbourne:
I'm closing this pull request, and resubmitting a new one, which is cleaned up and branched from the latest version of master

PR #236, opened 2015-09-17
Adding: Load(scalaCode:String, visible: Boolean) + Jar dependency error message

abbourne:
I've closed my previous messy pull request and replaced it with this one, which is clean and based on the latest version of your Master in Github. Apologies for the previous mess!
As per our gitter conversation, this pull request contains 2 changes:

Modifications the load.apply method in Interpreter.scala file to add support for evaluating code and getting a full printout, just as if the code was entered in the REPL itself. 

Here's an example of it in use:
```
@: load("""val foo = "I am cow!" """)

@: foo
res19: String = "I am cow!"
@: load("""val bar = 42; val baz = 2; bar*baz""", true)
bar: Int = 42
baz: Int = 2
res21_2: Int = 84
@: res21_2
res22: Int = 84
@:
```

I've tried all kinds of code, def's, multiple lines of input... it all seems to work fine.
I've included a small update to the Script Files section of the Repl.scalatex file to document the new REPL command.

A small modification to the Compiler.scala file to print error message if Jar dependencies cannot be found. It *only* prints if there is a fatal error, otherwise there is no additional output. This allowed me debug my problem with a Jar dependency issue, and I think it will help other people as well. The change is in the `def display(pos: Position, msg: String, severity: Severity) ` and its exactly the change from the diff file in lihaoyi/Ammonite#168

### DIFF
--- a/repl/src/main/scala/ammonite/repl/interp/Compiler.scala
+++ b/repl/src/main/scala/ammonite/repl/interp/Compiler.scala
@@ -101,7 +101,9 @@ object Compiler{
 
       def display(pos: Position, msg: String, severity: Severity) = {
         severity match{
-          case ERROR => logger(Position.formatMessage(pos, msg, false))
+          case ERROR => 
+            println(s"ERROR: $msg")
+            logger(Position.formatMessage(pos, msg, false))
           case _ => logger(msg)
         }
       }
--- a/repl/src/main/scala/ammonite/repl/interp/Interpreter.scala
+++ b/repl/src/main/scala/ammonite/repl/interp/Interpreter.scala
@@ -226,7 +226,13 @@ class Interpreter(prompt0: Ref[String],
         eval.addJar(jar.toURI.toURL)
       }
 
-      def apply(line: String) = processExec(line)
+      def apply(line: String, visible: Boolean) = {
+        if (!visible) processExec(line)
+        else {
+          val res = interp.processLine(line, Parsers.split(line).get.get.value, _.foreach(interp.stdout))
+          handleOutput(res)
+        }
+      }
 
       def exec(file: Path): Unit = apply(read(file))
 
--- a/readme/Repl.scalatex
+++ b/readme/Repl.scalatex
@@ -263,6 +263,32 @@
     @p
       All types, values and imports defined in scripts are available to commands entered in REPL after loading the script.
 
+   @p
+      Ammonite also provides a function to load a string of Scala code as if it were entered from the REPL. It comes with an option to load the code silently, or to get a printout of the results in the REPL exactly as if it was entered in the REPL:
+
+      @@
+
+      // scalaCode can be multiple lines, define functions, etc
+      // visible is an optional parameter that defaults to false (no output)
+      load(scalaCode: String, visible: Boolean = false)
+ 
+     @p
+      Here's an example:
+
+      @@
+
+      $: load("""val foo = "I am cow!" """)
+
+      $: foo
+      res19: String = "I am cow!"
+      $: load("""val bar = 42; val baz = 2; bar*baz""", true)
+      bar: Int = 42
+      baz: Int = 2
+      res21_2: Int = 84
+      $: res21_2
+      res22: Int = 84
+      $:
+
   @sect{Configuration}
     @p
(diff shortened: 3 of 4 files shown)


repl/src/main/scala/ammonite/repl/interp/Interpreter.scala
```diff
@@ -226,7 +226,13 @@ class Interpreter(prompt0: Ref[String],
         eval.addJar(jar.toURI.toURL)
       }
 
-      def apply(line: String) = processExec(line)
+      def apply(line: String, visible: Boolean) = {
+        if (!visible) processExec(line)
```
lihaoyi: Could we make `processExec` take an optional `stdout` argument, instead of duplicating the code below?

repl/src/main/scala/ammonite/repl/interp/Compiler.scala
```diff
@@ -101,7 +101,9 @@ object Compiler{
 
       def display(pos: Position, msg: String, severity: Severity) = {
         severity match{
-          case ERROR => logger(Position.formatMessage(pos, msg, false))
+          case ERROR => 
+            println(s"ERROR: $msg")
```
lihaoyi: won't this show on every compilation error?

repl/src/main/scala/ammonite/repl/interp/Interpreter.scala
abbourne: I'll look at your suggestion and try it out.

repl/src/main/scala/ammonite/repl/interp/Compiler.scala
abbourne: Yes... darn! You are right, it shows on every compilation error. Any suggestion as to the best way to fix this?

Editorial comment: I actually prefer the error printed by this line, to the more detailed "Compilation Failed" error the REPL currently prints.
for example, the current REPL:

```
billb-billb @: bad line
Compilation Failed
Main.scala:125: not found: value bad
bad line
^
billb-billb @:
```

* its verbose
* the `Main.scala:125:` is meaningless to the user.

The error printed by the new error println is quite succinct: 

```
billb-Ammonite (master) @: bad line
ERROR: not found: value bad
Compilation Failed
Main.scala:125: not found: value bad
bad line
^
billb-Ammonite (master) @:
```

Still... the new error println is inconsistent and repetitive.

repl/src/main/scala/ammonite/repl/interp/Interpreter.scala
abbourne: OK... I tried having `processExec` take an optional `stdout` argument, and I could not figure out a straightforward way to make it work:

*  `processExec` just calls `processScript`, `processScript` itself does not take any kind of printer.
*  `processExec` does pass `processScript` an `evaluateLine` function, which _does_ take a printer:  `{ (c, i) => evaluateLine(c, Nil, _ => () , i)}` But when I replace the `_ => ()` argument with an actual printer, I still don't get anything printed out.
* the signature for `evaluateLine` is : 

```
def evaluateLine(code: String,
                   printSnippet: Seq[String],
                   printer: Iterator[String] => Unit,
                   extraImports: Seq[ImportData] = Seq() ) 
```

*  I suspect I need to replace the `Nil` value passed to `evaluateLine` for the `printSnippet` with something to get printing to work, but I'm not sure what! 
* Here is what I have so far:

```
def apply(line: String, visible: Boolean) = {
     if (!visible) processExec(line) else processExec(line, _.foreach(interp.stdout)) 
     /*
     else {
        val res = interp.processLine(line, Parsers.split(line).get.get.value, _.foreach(interp.stdout))
        handleOutput(res)
     }
     */
}

def processExec(code: String, printer: Iterator[String] => Unit = _ => () ) =
    processScript(prepareScript(code), { (c, i) => evaluateLine(c, Nil, printer, i)})
```

Any suggestions to to where to go from here?

lihaoyi:
Sorry for taking long to get back to this!

Given the new `import $file` workflow for doing things, I'd like to phase out using the `load.foo` methods. I wouldn't mind a similar "verbose" mode for the `import $file`s, but that would be a different change than this one, so closing this for now...

abbourne:
Yeah... that's fine ... makes sense. Thanks for getting back to me on this.

PR #242, opened 2015-09-30
Added support for exiting on end of file.

### LINKED ISSUE #223: EOF on redirected stdin causes an infinite loop
Running

    echo "" | amm

produces infinite output (on OSX 10.10.5).

    Welcome to the Ammonite Repl 0.4.7
    (Scala 2.11.7 Java 1.8.0_45)
    prompt@ ......repeatingcharacters.....

patrick-premont:
Fixes #223. No tests added. The test infrastructure seems to only run tests encoded into strings at the moment. So this test would be slightly more involved.


lihaoyi:
Thanks @patrick-premont!

PR #254, opened 2015-11-01
Fix documentation url

ssanj:
Updated documentation link from lihaoyi.github.io/Ammonite to
http://lihaoyi.github.io/Ammonite.

When lihaoyi.github.io/Ammonite is used as the link to the
documentation it resolves to: https://github.com/lihaoyi/Ammonite/blob/master/lihaoyi.github.io/Ammonite because it is a relative url.


lihaoyi:
Thanks!

Issue #252, opened 2015-10-11
Tab completion on Array causes ammonite to exit


dickwall:
If you hit tab to complete on the following line after opening ammonite:

`val a = new Array` then hit tab to invoke completion, you get a `scala.reflect.internal.FatalError` and ammonite immediately exits. Example transcript is shown below.

Note that I found this because I created an `ArrayElement` class for a demo, and hit tab to complete after typing Array. Ammonite probably shouldn't completely bug out in this case even if there is an error in the completion.

```
Loading...
Welcome to the Ammonite Repl 0.4.8
(Scala 2.11.7 Java 1.8.0_60)
@ class ArrayElement {
    def x = 10
  }
defined class ArrayElement
@ val ae = new ArrayException in thread "main" scala.reflect.internal.FatalError: Something is wrong: cannot find T in applied type Array
   sought  T in Array
 classSym  Array in scala
  tparams  T in Array

	at scala.reflect.internal.Reporting$class.abort(Reporting.scala:59)
	at scala.reflect.internal.SymbolTable.abort(SymbolTable.scala:16)
	at scala.reflect.internal.tpe.TypeMaps$AsSeenFromMap.correspondingTypeArgument(TypeMaps.scala:565)
...
```

lihaoyi:
Fixed by @ssanj , thanks!

PR #258, opened 2015-11-12
Fix Tab completion on Array causes ammonite to exit (#252)

ssanj:
Added a Try around the awaitResponse method to prevent an Exception
killing the Ammonite Repl. The awaitResponse method throws an Exception
if a match can't be found for a completion.

The Failure case simply returns an empty List[pressy.Member]. This seems
like the simplest and least invasive fix for issue #252, although it may
not be ideal.


lihaoyi:
Thanks!

PR #263, opened 2015-11-17
Add regression test around #252 and fix formatting

ssanj:
Added a regression test against the fix for #252 to prevent it from resurfacing.


lihaoyi:
eek looks like it conflicted. Could you help rebase?

ssanj:
Yep, will push up once I rebase.

lihaoyi:
Looks like a flake, lgtm

Issue #256, opened 2015-11-09
`~repl/run` fails on a fresh checkout


Blaisorblade:
FYI, on a fresh checkout from ba949ec7c515fda5ba09a4b65100b18afab25510 `sbt ~repl/run` fails repeatedly. It seems it's trying to download `ammonite-shell` from some Ivy repository?

I'm sure I could workaround this by building and installing it locally, so it's a small regression, but maybe this could be avoided through an extra integration test?

```
$ sbt ~repl/run
[info] Loading global plugins from /Users/pgiarrusso/.sbt/0.13/plugins
[info] Loading project definition from /Users/pgiarrusso/AeroFS/Repos/Ammonite-tatooine/project
[info] Updating {file:/Users/pgiarrusso/AeroFS/Repos/Ammonite-tatooine/project/}ammonite-tatooine-build...
[info] Resolving org.fusesource.jansi#jansi;1.4 ...
[info] downloading https://repo.scala-sbt.org/scalasbt/sbt-plugin-releases/com.eed3si9n/sbt-assembly/scala_2.10/sbt_0.13/0.13.0/jars/sbt-assembly.jar ...
[info] 	[SUCCESSFUL ] com.eed3si9n#sbt-assembly;0.13.0!sbt-assembly.jar (1903ms)
[info] downloading https://repo1.maven.org/maven2/com/lihaoyi/scalatex-sbt-plugin_2.10_0.13/0.3.1/scalatex-sbt-plugin-0.3.1.jar ...
[info] 	[SUCCESSFUL ] com.lihaoyi#scalatex-sbt-plugin;0.3.1!scalatex-sbt-plugin.jar (484ms)
[info] downloading https://jcenter.bintray.com/org/scalactic/scalactic_2.10/2.2.1/scalactic_2.10-2.2.1.jar ...
[info] 	[SUCCESSFUL ] org.scalactic#scalactic_2.10;2.2.1!scalactic_2.10.jar(bundle) (454ms)
[info] Done updating.
[info] Compiling 1 Scala source to /Users/pgiarrusso/AeroFS/Repos/Ammonite-tatooine/project/target/scala-2.10/sbt-0.13/classes...
[info] Set current project to ammonite-tatooine (in build file:/Users/pgiarrusso/AeroFS/Repos/Ammonite-tatooine/)
[info] Updating {file:/Users/pgiarrusso/AeroFS/Repos/Ammonite-tatooine/}ops...
[...]

PR #265, opened 2015-11-18
Fix typo in command line

Blaisorblade:
As described in https://github.com/lihaoyi/Ammonite/issues/256#issuecomment-156995719, `--home=.ammonite-dev` doesn't work, you need `--home .ammonite-dev`.


lihaoyi:
Should be fixed in master

Issue #262, opened 2015-11-17
 ammonite.integration.StandaloneTests.hello fails if predef.scala has output


ssanj:
When running the  ammonite.integration.StandaloneTests, I was getting an unusual error:

```
[info] 1/4     ammonite.integration.StandaloneTests.hello
[info]              utest.AssertionError: assert(evaled == "Hello World")
[info]              evaled: String = loading common dependencies...
[info]              loaded scalaz-core, scalaz-concurrent, joda-time
[info]              Hello World
[info] 2/4     ammonite.integration.StandaloneTests.complex
[info]              java.lang.RuntimeException: Nonzero exit code: 1
[info] 3/4     ammonite.integration.StandaloneTests.shell       Success
[info] 4/4     ammonite.integration.StandaloneTests     Success
[info] -----------------------------------Results-----------------------------------
[info] ammonite.integration.StandaloneTests     Success
[info]     hello        Failure('utest.AssertionError: assert(evaled == "Hello World")
[info]     evaled: String = loading common dependencies...
[info]     loaded scalaz-core, scalaz-concurrent, joda-time
[info]     Hello World')
[info]     complex      Failure('java.lang.RuntimeException: Nonzero exit code: 1')
[info]     shell        Success
[info] Failures:
[info] 1/4     ammonite.integration.StandaloneTests.hello
[info]              utest.AssertionError: assert(evaled == "Hello World")
[info]              evaled: String = loading common dependencies...
[info]              loaded scalaz-core, scalaz-concurrent, joda-time
[info]              Hello World
```

After a bit of digging around I figured out that the StandaloneTests where expecting:

```
Hello World
```

but received:

```
loading common dependencies...
loaded scalaz-core, scalaz-concurrent, joda-time
Hello World
```

Since the evaluation used == for the comparison it was failing:

```
'hello{
   val evaled = exec("Hello.scala")
   assert(evaled == "Hello World")
}
```

I had a look at the Hello.scala file:

```
println("Hello World")
```

and it did not have any dependencies. 

Where were these dependencies being loaded from?

It turned out that I had customised my ~/.ammonite/predef.scala file as:

```
println("loading common dependencies...")
load.ivy("org.scalaz" %% "scalaz-core" % "7.1.5")
load.ivy( "org.scalaz" %% "scalaz-concurrent" % "7.1.5")
load.ivy("org.scalacheck" %% "scalacheck" % "1.12.5")
[...]

PR #266, opened 2015-11-18
Fix StandaloneTests.hello failure with predef.scala #262

ssanj:
I've updated the exec method to use an empty predef file, thus
isolating the tests from the effects of a user-supplied predef file.

It used to be that any output from a user's predef.scala file would
conflict with the 'hello test causing it to fail.

issue: #262

### DIFF
--- a/integration/src/test/scala/ammonite/integration/StandaloneTests.scala
+++ b/integration/src/test/scala/ammonite/integration/StandaloneTests.scala
@@ -16,28 +16,38 @@ import utest.framework.TestSuite
 object StandaloneTests extends TestSuite{
   // Prepare standalone executable
   val scalaVersion = scala.util.Properties.versionNumberString
+
   val tests = TestSuite {
     val ammVersion = ammonite.Constants.version
     val executableName = s"ammonite-repl-$ammVersion-$scalaVersion"
     val Seq(executable) = ls.rec! cwd |? (_.last == executableName)
-    val resources = cwd/'repl/'src/'test/'resources/'standalone
+    val replStandaloneResources = cwd/'repl/'src/'test/'resources/'standalone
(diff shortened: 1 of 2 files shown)


ssanj:
@lihaoyi  Any idea why this build is failing?

lihaoyi:
Looks like a flake, probably can be ignored

integration/src/test/scala/ammonite/integration/StandaloneTests.scala
```diff
@@ -16,28 +16,38 @@ import utest.framework.TestSuite
...
+
   val tests = TestSuite {
     val ammVersion = ammonite.Constants.version
     val executableName = s"ammonite-repl-$ammVersion-$scalaVersion"
     val Seq(executable) = ls.rec! cwd |? (_.last == executableName)
-    val resources = cwd/'repl/'src/'test/'resources/'standalone
+    val replStandaloneResources = cwd/'repl/'src/'test/'resources/'standalone
+    val shellAmmoniteResources = cwd/'shell/'src/'main/'resources/'ammonite/'shell
+    //use Symbol to wrap symbols with dashes.
+    val emptyPrefdef = shellAmmoniteResources/Symbol("empty-predef.scala")
```
lihaoyi: You can just use the string `"empty-predef.scala"` here, no need the `Symbol` wrapper

lihaoyi:
Bump, @ssanj are you gonna swap the symbol constructor for a String?

ssanj:
Yes will swap it out. Just out of town for the weekend. Should be in Monday.

ssanj:
@lihaoyi any idea why this build is failing?

lihaoyi:
@ssanj looks like a flake, should be fine =)

lihaoyi:
Looks good, thanks!

PR #267, opened 2015-11-19
Fix readme with correct predef file used in shell test

ssanj:
Updated the readme to point to the correct predef file used to run
the application under shell/test.

The existing link pointed to:
readme/resources/example-predef.scala

The new link points to:
shell/src/main/resources/ammonite/shell/example-predef.scala


lihaoyi:
Can you rebase? =D

ssanj:
Fixed in #268

Issue #259, opened 2015-11-14
Ammonite-Shell fail to run shell command on Mac OSX with gun coreutils installed


wuhx:
```
wuhx@gynoid:λ ./amm
Loading...
Welcome to the Ammonite Repl 0.4.8
(Scala 2.11.7 Java 1.8.0_11)
wuhx-sc@ wd
stty: standard input: unable to perform all requested operations
Exception in thread "main" java.lang.RuntimeException: Nonzero exit value: 1
	at scala.sys.package$.error(package.scala:27)
	at scala.sys.process.ProcessBuilderImpl$AbstractBuilder.slurp(ProcessBuilderImpl.scala:132)
	at scala.sys.process.ProcessBuilderImpl$AbstractBuilder.$bang$bang(ProcessBuilderImpl.scala:102)
	at ammonite.terminal.TTY$.stty(Utils.scala:78)
	at ammonite.terminal.TermCore$.readLine(TermCore.scala:240)
	at ammonite.repl.frontend.AmmoniteFrontEnd.readLine(AmmoniteFrontEnd.scala:114)
	at ammonite.repl.frontend.AmmoniteFrontEnd.action(AmmoniteFrontEnd.scala:25)
	at ammonite.repl.Repl.action(Repl.scala:45)
	at ammonite.repl.Repl.loop$1(Repl.scala:78)
	at ammonite.repl.Repl.run(Repl.scala:96)
	at ammonite.repl.Repl$.run(Repl.scala:195)
	at ammonite.repl.Repl$$anonfun$main$1.apply(Repl.scala:159)
	at ammonite.repl.Repl$$anonfun$main$1.apply(Repl.scala:158)
	at scala.Option.foreach(Option.scala:257)
	at ammonite.repl.Repl$.main(Repl.scala:158)
	at ammonite.repl.Repl.main(Repl.scala)
```

when I remove gnu coreutils from PATH env， It works fine。
so it seems to have some incompatible issue with gnu coreutils (installed by brew) on osx.

lihaoyi:
Should be fixed by https://github.com/lihaoyi/Ammonite/pull/269

Issue #212, opened 2015-09-07
bare repl exits with stacktrace on every command


aberezin:
I have created a fresh, bare Ammonite on OSX 10.10.5 running OS X bash 3.2.57
JDK 1.8.0_45 and Scala 2.11.7

Anything I enter on the cmd line causes the repl to exit with a stack trace.  I have installed the repl on an Ubuntu box and do not have a problem.  Here is the terminal session on the osx iTerm. Note I entered *val x=1* on the cmd line in the example below. 

```bash
(py278)bin $ version
10.10.5
(py278)bin $ echo $BASH_VERSION
3.2.57(1)-release
(py278)bin $ java -version
java version "1.8.0_45"
Java(TM) SE Runtime Environment (build 1.8.0_45-b14)
Java HotSpot(TM) 64-Bit Server VM (build 25.45-b02, mixed mode)
(py278)bin $ scala -version
Scala code runner version 2.11.7 -- Copyright 2002-2013, LAMP/EPFL
(py278)bin $ sbt --version
Java HotSpot(TM) 64-Bit Server VM warning: ignoring option MaxPermSize=384m; support was removed in 8.0
sbt launcher version 0.13.7
(py278)bin $ rm -rf ~/.ammonite
(py278)bin $ curl -L -o amm http://git.io/vZJoU; chmod +x amm; ./amm
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
100   404    0   404    0     0    567      0 --:--:-- --:--:-- --:--:--   567
100 29.3M  100 29.3M    0     0  4499k      0  0:00:06  0:00:06 --:--:-- 6145k
Loading...
Welcome to the Ammonite Repl 0.4.7
(Scala 2.11.7 Java 1.8.0_45)
@ val x=1
stty: standard input: unable to perform all requested operations
Exception in thread "main" java.lang.RuntimeException: Nonzero exit value: 1
	at scala.sys.package$.error(package.scala:27)
	at scala.sys.process.ProcessBuilderImpl$AbstractBuilder.slurp(ProcessBuilderImpl.scala:132)
	at scala.sys.process.ProcessBuilderImpl$AbstractBuilder.$bang$bang(ProcessBuilderImpl.scala:102)
	at ammonite.terminal.TTY$.stty(Utils.scala:78)
[...]

Issue #143, opened 2015-08-06
ammonite crashes on any input


bretthoerner:
`ammonite 0.4.2`

I did the `curl` install from the website. Regardless of what I do I get the following.

```
$ amm 
Loading Ammonite Repl...
@ 1
stty: standard input: unable to perform all requested operations
Exception in thread "main" java.lang.RuntimeException: Nonzero exit value: 1
	at scala.sys.package$.error(package.scala:27)
	at scala.sys.process.ProcessBuilderImpl$AbstractBuilder.slurp(ProcessBuilderImpl.scala:132)
	at scala.sys.process.ProcessBuilderImpl$AbstractBuilder.$bang$bang(ProcessBuilderImpl.scala:102)
	at ammonite.terminal.TTY$.stty(Utils.scala:97)
	at ammonite.terminal.TermCore$.readLine(TermCore.scala:216)
	at ammonite.repl.frontend.FrontEnd$Ammonite$.readLine(FrontEnd.scala:148)
	at ammonite.repl.frontend.FrontEnd$Ammonite$.action(FrontEnd.scala:63)
	at ammonite.repl.Repl.action(Repl.scala:44)
	at ammonite.repl.Repl.loop$1(Repl.scala:68)
	at ammonite.repl.Repl.run(Repl.scala:82)
	at ammonite.repl.Repl$.run(Repl.scala:146)
	at ammonite.repl.Repl$$anonfun$main$1.apply(Repl.scala:130)
	at ammonite.repl.Repl$$anonfun$main$1.apply(Repl.scala:130)
	at scala.Option.foreach(Option.scala:257)
	at ammonite.repl.Repl$.main(Repl.scala:130)
	at ammonite.repl.Repl.main(Repl.scala)
```

Thanks!

lihaoyi:
Thanks for the report!

What operating system are you using? What terminal? What version of each? What version of Java?

bretthoerner:
OS X 10.10.4
GNU bash, version 4.3.39(1)-release (x86_64-apple-darwin14.4.0)
java version "1.7.0_71"
Terminal.app 2.5.3 (343.7)

lihaoyi:
Could you try running
```bash
$ bash -c "stty -g < /dev/tty"
```

And then taking the output and pasting it back in in place of the `-g`, e.g. in my case it looks like

```bash
haoyi-mbp:citools haoyi$ bash -c "stty -g < /dev/tty"
[...]

PR #269, opened 2015-11-20
Have terminal.TTY commands prefer standard paths

LannyRipple:
Fix for #143, #212, #259.
Tested on MacOSX and Ubuntu.  No Windows box to determine
standard path for `stty` and `tput`.


lihaoyi:
Red build looks like a flake, so it should be fine.

Could you explain to me, as someone who knows nothing about tput (and very little about bash) what this is doing? I trust you that it works but I'd love to understand a bit more about the code I'm merging ^_^

lihaoyi:
Merged as https://github.com/lihaoyi/Ammonite/commit/6a0e7eefe20f1a3f6cf1342d71d79e3dfcf59bdc

LannyRipple:
Sorry Haoyi,  wasn't paying attention to this.

On both Mac and Linux the tput and stty commands that come with the system are found on system paths /usr/bin/tput and /bin/stty.  (I don't have any knowledge on tput and stty beyond what the man pages says about them so can't comment further there.)  With bash any command that is not provided by rooted or relative path is searched for on the directories of the user's PATH environment variable with first-found being used.  On Macs using homebrew for gnu's coreutils the package will suggest placing a path containing tput and stty before the system defaults.  This patch attempts to first locate tput and stty along system paths and if they cannot be found falls back to letting the system find them using the user's PATH (i.e., pre-patch behavior).

As an aside I'm not sure why the coreutil's stty doesn't handle -g and reading the results back in.  The flag and behavior seem to be there.

PR #273, opened 2015-11-23
Add support for word navigation via Ctrl+Arrow. Fixes #217.

### LINKED ISSUE #217: support for ctrl + arrow
In scala repl it works fine. It is very helpful and I miss that.

cranst0n:


lihaoyi:
@cranst0n is this for Ctrl-Arrow or Alt-Arrow? Alt-Arrow is already working on my machine (OSX iTerm2), and seems to be something different from what's being described in#217

cranst0n:
You're right. My mistake.This actually provides Ctrl+arrow. I can rename and push tomorrow if that's the only issue you see.

lihaoyi:
Yeah that's all i see. Could you add a line to the 0.5.1 changelog too? =)

cranst0n:
I believe you've named the unicode escape character as [ALT](https://github.com/lihaoyi/Ammonite/blob/4e285d76200ab8182506321c40d563a4568722b4/terminal/src/main/scala/ammonite/terminal/SpecialKeys.scala#L23) (see [here](http://www.fileformat.info/info/unicode/char/1b/index.htm)). Maybe this should be changed for clarity?

lihaoyi:
Odd, on my machine ALT is the unicode escape. Could it be the same character? If so, Esc-Left and Esc-Right should already work since they're defined here https://github.com/lihaoyi/Ammonite/blob/master/terminal/src/main/scala/ammonite/terminal/GUILikeFilters.scala#L110-L115. 

I guess these keys just have different bindings on different OSs? In that case maybe renaming the patch `LinuxAltLeft` or something and putting them right under the existing `AltLeft`s would work...

cranst0n:
@lihaoyi how do the updates sit with you?

lihaoyi:
Looks good, thanks @cranst0n! I have not tested this since I don't have a linux machine but I assume you know what you're doing

Issue #197, opened 2015-08-25
reverse-i-search not working in Repl


fsvehla:
In the SBT repl `CTRL` + `R` searches backwards in the history, for example:

1) Enter Text

```text
scala> val foo = 22
```

2) `CTRL` + `R` + `f` `o` will find and highlight the line, with this output:

```
(reverse-i-search)`fo': val foo = 22
```

3) `RETURN` accepts the choice

lihaoyi:
There's no reason this is missing, just nobody's implemented it yet. However, I am probably not the best person to implement this, as I'm not actually familiar with how this works. If anyone wants to take a crack at it, you can put it in https://github.com/lihaoyi/Ammonite/blob/master/terminal/src/main/scala/ammonite/terminal/ReadlineFilters.scala with the rest of the readline hotkey filters

fsvehla:
Switching the frontend to use JlineUnix works as a workaround, but when I do that, what functionality am I missing?

lihaoyi:
You're missing

- Multiline editing
- Input-syntax-highlighting
- Desktop-style Alt-{left, right} navigation, Shift-selection
- Probably a few other things I don't remember off the top of my head

lastland:
From the code of jline2, how `reverse-i-search` works seems quite simple. It just searches the most recent command that contains the given string. See https://github.com/jline/jline2/blob/8beae1a471aef17e4c721021ea4be3be918da37f/src/main/java/jline/console/ConsoleReader.java#L2844-L2864 and https://github.com/jline/jline2/blob/8beae1a471aef17e4c721021ea4be3be918da37f/src/main/java/jline/console/ConsoleReader.java#L3890-L3930 .

BTW, notice that there are a bunch of searching commands besides `reverse-i-search` in `readline`: http://www.delorie.com/gnu/docs/readline/rlman_15.html
Perhaps we should implement all of them?

lihaoyi:
Fixed in https://github.com/lihaoyi/Ammonite/pull/335

PR #281, opened 2015-11-29
Make up/down search command by prefix.

olafurpg:
Up/down will only list commands that share prefix with the last manually
edited line. This makes it easy to rerun commands like `load.ivy(???)`.

### DIFF
--- a/terminal/src/main/scala/ammonite/terminal/ReadlineFilters.scala
+++ b/terminal/src/main/scala/ammonite/terminal/ReadlineFilters.scala
@@ -1,5 +1,6 @@
 package ammonite.terminal
 import acyclic.file
+import scala.collection.mutable
 import FilterTools._
 import ammonite.terminal.LazyList._
 import SpecialKeys._
@@ -40,9 +41,7 @@ object ReadlineFilters {
     Case(Ctrl('b'))((b, c, m) => (b, c - 1)), // <- one char
     Case(Ctrl('f'))((b, c, m) => (b, c + 1)), // -> one char
     Case(Alt + "b")((b, c, m) => GUILikeFilters.wordLeft(b, c)), // <- one word
-    Case(LinuxCtrlLeft)((b, c, m) => GUILikeFilters.wordLeft(b, c)), // <- one word
     Case(Alt + "f")((b, c, m) => GUILikeFilters.wordRight(b, c)), // -> one  word
-    Case(LinuxCtrlRight)((b, c, m) => GUILikeFilters.wordRight(b, c)), // -> one word
     Case(Home)((b, c, m) => BasicFilters.moveStart(b, c, m.width)), // <- one line
     Case(HomeScreen)((b, c, m) => BasicFilters.moveStart(b, c, m.width)), // <- one line
     Case(Ctrl('a'))((b, c, m) => BasicFilters.moveStart(b, c, m.width)),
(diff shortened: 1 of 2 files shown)


olafurpg:
Example use case.
Demo

olafurpg:
After some consideration:
* we might want to hold off on this PR for now, because it doesn't properly handle the scenario when the user 1) presses up/down a few times 2) edits the command 3) presses up/down again. The `index` variable will be off. Also, the tests don't cover this behavior.
* Users might want to configure how up/down history behaves, that is whether it should be like bash or (oh-)my-zsh.

olafurpg:
Ping. Any chance to get feedback on this PR? I am badly missing this feature :smile:

lihaoyi:
Sorry, I haven't had a chance to look at this I'm afraid!!! Will take a look when I have some time...

lihaoyi:
Sorry I've been a bit overloaded and unable to review properly!

- This is cool
- This could potentially solve https://github.com/lihaoyi/Ammonite/issues/197 as well, if done properly
- Bash has Ctrl-R, Zsh has both Ctrl-R and this matched-prefix-history, Fish has only the match-prefixed-history except it works for arbitrary matched-substrings. Fish doesn't have Ctrl-R

Ideally I'd want to be able to satisfy both requirements with the same feature. There's been a bunch of discussion on the fish issue tracker about the missing features of the Fish implementation: https://github.com/fish-shell/fish-shell/issues/602

I'm pretty sure we'd be able to satisfy the concerns in https://github.com/fish-shell/fish-shell/issues/602, and be able to provide one feature that subsumes both the zsh-prefix-history thing and the reverse-history-search thing. In that case I'm reluctant to implement them both =P

--------------------

Here's what I think we should do:

- Make history-with-up work on substring rather than prefix
- Highlight the "current search term" differently from the "rest" similar to what fish does, so it's clear what's being search for and what's just part of the match

screen shot 2015-12-13 at 12 25 29 pm

- Leave the cursor at the end of the search term (instead of EOL like fish does) and allow editing of the search term. This would, AFAICT, satisfy the complaints of https://github.com/fish-shell/fish-shell/issues/602 and allow us to totally replace Ctrl-R search
- Add a stub so that if you hit Ctrl-R, it basically presses `[up]` for you with your current input to start the search with the current input.

Some musings:

- How do you "end" a search? I guess navigating the cursor "out" of the search term would do it. Pressing `[enter]` should end the search and execute the command. `[tab]` might bring you out of the search and put your cursor at end-of-line
- Should it support regexes? If so it probably couldn't be done in-line, and the search-string and search-result would have to be separated
- How will search interact with multi-line input? "It doesn't" might be enough, but worth thinking about to decide that explicitly.

This is probably more ambitious than your current diff, but I think it's probably a good long term solution.

WDYT? If you think it's too much work to do that's fine, I'll pick it up and implement it when I have the time, using your code as a base.

olafurpg:
Sorry for the slow response, holidays... 

Thanks for the detailed review.

I am not convinced that Ctrl-R should work the same. I may be wrong, but my gut tells me I sometimes want to search by prefix only (for example `load.ivy`).

Also, the substring match is more complicated UX wise, problems pop up like "how do I edit the search term?", "how do I end the search?". Prefix search is in many ways simpler.

So here's my suggestion. I complete the prefix search, I'll underline the search term and improve the test suite to cover cases like `edit -> up -> edit -> up`. That should give a good foundation for future substring search. WDYT?

lihaoyi:
@olafurpg Now my turn to be sorry for the slow response :stuck_out_tongue: 

What you said sounds fine. Make the improvements you suggested (underlining, test coverage) and i'll give it another non-feature-wishlist review and then we'll land it. This should be a sufficient improvement for now and in future if we want to make it "better" we can do so on top of your diff.

Sorry for the long round trips, and thanks for putting in the effort to do this!

olafurpg:
All righty, here we go!

I refactored the HistoryFilter quite a bit to aid in debugging/reasoning in tricky corner cases. WDYT? :smile_cat:

olafurpg:
I added yet another commit so that edited commands overwrite the original command, it can be super annoying to make a large edit to a previous command, press up/down by accident and loose the large edit. I recommend you play a bit around with it to see how it feels. It's possible to get the original line (before edit) by Ctrl-c and searching again.

PS. seems like there are flaky tests, I don't think my commit caused the broken Travis build.

lihaoyi:
This is superseded by #335

PR #284, opened 2015-12-05
add VI mode filters for Ammonite

chengpohi:
Enable VI mode in Ammonite repl, For now it supports:

|ShortCut | Function|
|:----|:----|
|Esc + Enter |Toggle VI Model|
|Esc|Toggle Visual Mode|
|h|move to left char|
|l|move to right char|
|i|insert mode|
|a|insert mode|
|x|delete current char|
|0|go to the start of line|
|d + d|delete current line|
|$|go to the end of line|
|b|go to the left word|
|w|go to the next word|
|j|go to the next history|
|k|go to the previous history|

### DIFF
--- a/repl/src/main/scala/ammonite/repl/frontend/AmmoniteFrontEnd.scala
+++ b/repl/src/main/scala/ammonite/repl/frontend/AmmoniteFrontEnd.scala
@@ -1,12 +1,11 @@
 package ammonite.repl.frontend
 
-import java.io.{OutputStreamWriter, OutputStream, InputStream}
+import java.io.{InputStream, OutputStream, OutputStreamWriter}
 
 import ammonite.repl._
 import ammonite.terminal.LazyList.~:
 import ammonite.terminal._
 import fastparse.core.Result
-import scala.annotation.tailrec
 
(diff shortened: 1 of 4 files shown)


lihaoyi:
This looks reasonable. 

If you want to have it in a separate module, you'd hook into Ammonite's filters by mutating Ammonite's `repl.frontEnd()` variable, as is done [here](https://github.com/lihaoyi/Ammonite/blob/master/shell/src/main/scala/ammonite/shell/Configure.scala#L10-L12). Then whoever wants to enable it will `load.ivy` your library and then call a similar `configure(repl)` snippet to add it to the list of the `AmmoniteFrontEnd`'s `extraFilters`.  

Since there will be more than one "extra filter" in the world, you probably want to pattern match on `repl.frontEnd()`, do nothing in the case the user has a `JLineFrontEnd`, and create a new `AmmoniteFrontEnd` with your extra filter appended (or prepended?) to the previous `AmmoniteFrontEnd`'s `extraFilters`. 

That would allow both your module and the `ammonite-shell` module to co-exist, rather than having one stomp over the other. I will have to update `ammonite-shell` to do the same thing, which I'll likely do in the next release

chengpohi:
Thanks haoyi, I will try to add the vi mode as a plugin.

PR #298, opened 2015-12-10
Added the possibility to specify resolvers from the repl

etorreborre:
I don't know if this is what you had in mind but this works ok for me. Feel free to change, adjust, etc...

When I ran the repl tests, I had contradictory messages like:
```
[info] Tests: 166
[info] Passed: 166
[info] Failed: 0
[error] (ops/test:test) sbt.TestsFailedException: Tests unsuccessful
[error] Total time: 632 s, completed 10/12/2015 10:16:55 PM
```

Maybe you can help me there?

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -31,7 +31,7 @@ val sharedSettings = Seq(
   organization := "com.lihaoyi",
   version := _root_.ammonite.Constants.version,
   libraryDependencies += "com.lihaoyi" %% "utest" % "0.3.0" % "test",
-  testFrameworks += new TestFramework("utest.runner.Framework"),
+  testFrameworks := Seq(new TestFramework("utest.runner.Framework")),
   scalacOptions += "-target:jvm-1.7",
   autoCompilerPlugins := true,
   addCompilerPlugin("com.lihaoyi" %% "acyclic" % "0.1.3"),
--- a/build.sbt
+++ b/build.sbt
@@ -66,7 +66,7 @@ val sharedSettings = Seq(
           <url>https://github.com/lihaoyi</url>
         </developer>
       </developers>
-)
+) ++ prompt
 
 /**
  * Concise, type-safe operating-system operations in Scala: filesystem,
@@ -78,6 +78,10 @@ lazy val ops = project
     name := "ammonite-ops"
   )
 
+lazy val prompt = shellPrompt in ThisBuild := { state =>
+  val name = Project.extract(state).currentRef.project
+  (if (name == "ammonite") "" else name) + "> "
+}
 
 /**
  * A standalone re-implementation of a composable readline-style REPL,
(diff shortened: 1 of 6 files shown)


lihaoyi:
Looks great. Not sure why the test framework is misbehaving, let's see what CI says

lihaoyi:
CI seems unhappy. Looks like the test you added failed

https://travis-ci.org/lihaoyi/Ammonite/jobs/96009323#L1661

Seems like it can't find the dependency. No idea why

And your lines are too long :stuck_out_tongue: 

https://travis-ci.org/lihaoyi/Ammonite/jobs/96009323#L622

lihaoyi:
Now it looks broken in 2.10. Is that library not published for 2.10? We could just skip the test if that's the case

build.sbt
```diff
@@ -78,6 +78,10 @@ lazy val ops = project
     name := "ammonite-ops"
   )
 
+lazy val prompt = shellPrompt in ThisBuild := { state =>
```
lihaoyi: What's this about?

etorreborre:
Cool, I better understand how this works now :-). I have added the possibility to specify resolvers using ivy-style patterns (I think my previous manual test might just have worked because the jar was already cached).

I also added 2 small sbt improvements:

 1. make utest the only test framework to avoid ScalaCheck intercepting arguments on the command line and reporting errors

 2. add a prompt with the module name when navigating to a given submodule, like `repl>`

build.sbt
etorreborre: This is about getting a better prompt for submodules inside sbt.

etorreborre:
Yes the build is still broken. I changed the jar version to something that is published for both `2.10` and `2.11`.

etorreborre:
@lihaoyi [This 2.11.4 build](https://travis-ci.org/lihaoyi/Ammonite/jobs/96175902) is failing, do you have any idea why? It looks like it has been failing for quite some time.

lihaoyi:
Yeah that one's been misbehaving and i haven't had time to investigate =/ If the rest are green it should be fine

etorreborre:
Yes the rest is fine.

build.sbt
lihaoyi: I don't understand =/

lihaoyi: What prompt are you talking about? Why do we care about the submodule's prompt? Why is this prompt better?

lihaoyi:
Looks good

build.sbt
etorreborre: Sorry I should have been clearer. When you have a multi-module sbt project you can "navigate" from the top-level module to a submodule and only run the tests for that module:
```
./sbt
> project repl
> test
```

With this setting you will see the module name before the prompt so you know where you are
```
./sbt
> project repl
repl> test
```

lihaoyi:
@etorreborre this should be in 0.5.2, which contains this with some tweaks to make it properly invalidate ammonite's own ivy cache when you change resolvers 

http://lihaoyi.github.io/Ammonite/#0.5.2

Try it out and let me know if it works for you!

PR #299, opened 2015-12-10
change the fullHistory type to StableRef for extrafilters can get the…

chengpohi:
Change the fullHistory type to StableRef for the extraFilters can get the reference of fullHistory.

So extraFilters can sync the history between filters.


lihaoyi:
Can you take a moment to explain your use case for this? We discussed it briefly but I still don't really understand. Why do you want to modify the `fullHistory` from inside your filters?

chengpohi:
@lihaoyi, In my filter, I will switch the history and I don't need to modify `fullHistory`, but I want to get the newest `fullHistory` when `fullHistory` has updated, So I want to get the *reference* of fullHistory.

```
 addHistory = (code) => if (code != "") {
        storage().fullHistory() = storage().fullHistory() :+ code //if fullHistory has appended, my filter also need to get the appended
        history = history :+ code
 }
```

chengpohi:
This is my filter code:

```
 val viHistoryFilter = VIHistoryFilter(repl.fullHistory)
 case class VIHistoryFilter(history: StableRef[History]) extends TermCore.DelegateFilter {
    val historyFilter = ReadlineFilters.HistoryFilter(() => history().reverse) 

    override def filter: Filter = {
      case TS('j' ~: rest, b, c) if VISUAL_MODE =>
        historyFilter.nextHistory(b, rest) 
      case TS('k' ~: rest, b, c) if VISUAL_MODE =>
        historyFilter.previousHistory(b, rest)
    }
  }
```

Does this explanation is clearly?

lihaoyi:
If you don't need to modify it, and just need to get the latest fullHistory when it has updated, how about define your argument as 

```scala
history: => History
```

That will make it get pulled in fresh every time it gets used, rather than once at the start when the filter is initialized. If you aren't editing it you don't really need a `Ref`, and I probably mis-understood your requirements when I told you to use one :stuck_out_tongue:

chengpohi:
Thanks @lihaoyi, This way worked, I only want to get the newest fullHistory.

Issue #301, opened 2015-12-11
java.util.NoSuchElementException: None.get on Shift+Tab without selection.


senia-psm:
Ammonite REPL fails with `java.util.NoSuchElementException: None.get` on `Shift+Tab` without selection.

Steps to reproduce:
1) Open Ammonite REPL
2) Press `Shift+Tab`

On empty line:
```scala
scala> ammonite.repl.Main.run("")
Loading...
Welcome to the Ammonite Repl 0.5.1
(Scala 2.11.7 Java 1.7.0_91)
@ java.util.NoSuchElementException: None.get
...
```
In multi-line edit mode without selection:
```scala
scala> ammonite.repl.Main.run("")
Loading...
Welcome to the Ammonite Repl 0.5.1
(Scala 2.11.7 Java 1.7.0_91)
@ {
  1
  java.util.NoSuchElementException: None.get
  at scala.None$.get(Option.scala:347)
  at scala.None$.get(Option.scala:345)
  at ammonite.terminal.GUILikeFilters$SelectionFilter.doIndent(GUILikeFilters.scala:26)
  at ammonite.terminal.GUILikeFilters$SelectionFilter$$anonfun$filter$1.applyOrElse(GUILikeFilters.scala:72)
  at ammonite.terminal.GUILikeFilters$SelectionFilter$$anonfun$filter$1.applyOrElse(GUILikeFilters.scala:70)
  at scala.runtime.AbstractPartialFunction.apply(AbstractPartialFunction.scala:36)
  at ammonite.terminal.FilterTools$$anon$1$$anonfun$apply$2.apply(FilterTools.scala:14)
  at ammonite.terminal.FilterTools$$anon$1$$anonfun$apply$2.apply(FilterTools.scala:14)
  at scala.Option.map(Option.scala:146)
  at ammonite.terminal.FilterTools$$anon$1.apply(FilterTools.scala:14)
  at ammonite.terminal.TermCore$DelegateFilter$class.apply(TermCore.scala:106)
  at ammonite.terminal.GUILikeFilters$SelectionFilter.apply(GUILikeFilters.scala:15)
  at ammonite.terminal.GUILikeFilters$SelectionFilter.apply(GUILikeFilters.scala:15)
  at scala.PartialFunction$class.applyOrElse(PartialFunction.scala:123)
  at ammonite.terminal.GUILikeFilters$SelectionFilter.applyOrElse(GUILikeFilters.scala:15)
[...]

PR #302, opened 2015-12-11
#301 None.get on Shift+Tab without selection.

senia-psm:
https://github.com/lihaoyi/Ammonite/issues/301


lihaoyi:
Looks good, thanks!

lihaoyi:
Should be fixed in 0.5.2 which is out http://lihaoyi.github.io/Ammonite/#0.5.2

PR #309, opened 2015-12-19
fix crash when autocompleting an unknown import

ssanj:
When trying to autocomplete an unknown or mistyped import such as:

```
import scala.collction.<tab>
```

Ammonite exits with the following error:

```
@ import scala.collction.Exception in thread "main"
java.lang.ClassCastException: scala.reflect.internal.Trees$Select
    cannot be cast to scala.reflect.internal.Trees$Ident
    at ammonite.repl.interp.Pressy$Run.prefixed(Pressy.scala:135)
    at ammonite.repl.interp.Pressy$$anon$3.complete(Pressy.scala:231)
```

The error is due to casting a scala.reflect.internal.Trees$Select to a
scala.reflect.internal.Trees$Ident. The fix simply tests for the
Trees$Ident type before casting to it.

If the expr type of the Import is not a Trees$Ident then we return no
matches.

Also added a couple of tests for the fix to repl/AutocompleteTests.

### DIFF
--- a/repl/src/main/scala/ammonite/repl/interp/Pressy.scala
+++ b/repl/src/main/scala/ammonite/repl/interp/Pressy.scala
@@ -131,13 +131,15 @@ object Pressy {
         // If the selectors haven't been defined yet...
         if (selectors.head.name.toString == "<error>") {
           if (expr.tpe.toString == "<error>") {
-            // If the expr is badly typed, try to scope complete it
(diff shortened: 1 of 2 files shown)


lihaoyi:
CI seems to be failing consistently in 2.11.x, probably a real issue

lihaoyi:
Actually it looks like a flake in maven central, re-kicked off tests...

lihaoyi:
Thanks!

PR #311, opened 2015-12-21
Ability to save things defined in REPL to jar - praveens/repl/write jar

praveen-srinivasan:
Strict a proof-of-concept as this point (but still usable). Will await guidance about what changes are needed to eventually get into master.

Use as 

repl.writeJar("/tmp/foo.jar")

---
This change is [Reviewable](https://reviewable.io/reviews/lihaoyi/ammonite/311)

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -224,16 +224,6 @@ lazy val sshd = project
       )
   )
 
-lazy val jupyter_kernel = project
-  .dependsOn(repl)
-  .settings(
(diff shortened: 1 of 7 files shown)


lihaoyi:
Going to close this for now since it's fallen behind, and will fall further behind with changes I have planned that will make this task simpler in future. This is still a feature I want "eventually", but I don't think I'm ready to merge this right now

PR #318, opened 2016-01-11
Correctly print a multi-line prompt

thirstycrow:


lihaoyi:
Thanks! will take a look wwhen i get time

lihaoyi:
This seems to cause problems with the coloring of the last line of the multi-line prompt
screenshot 2016-01-17 12 41 36

The likely reason seems to be that in here:

```scala
 +      writer.write(
 +        if (fullPrompt) prompt
 +        else prompt.split("\n").last
 +      )
```

You are trimming the last piece of the `prompt` string, but also discarding any possibly earlier color being set in `prompt`. One possible solution would be to walk the pre-last segments of `prompt` and pick out the last ANSI escapes in the few categories we probably want to support (color, background-color, underline, reset) and pre-pending those to the last segment that forms the prompt.

That's the main thing. Apart from that, could you add this to the changelog in the readme?

thirstycrow:
Fixed the coloring issue. But this test is now broken: ammonite.sshd.SshdReplTest.canExecuteRemoteCommand. I have no idea what's going wrong there.

thirstycrow:
Found the problem. I forgot the single-line-prompt case.

lihaoyi:
Looks good, thanks!

PR #326, opened 2016-01-16
Fix ls'ing empty directory gives exception bug

coderabhishek:


lihaoyi:
Thanks for investigating this!

It seems that a slightly better solution would be to make `FrontEndUtils.tabulate` check for an empty input list, and have it return an empty string (rather than crashing) if the list is empty. Presumably if anyone else wants to use `FrontEndUtils.tabulate`, they would be interested in it not crashing too. That way the code for `lsSeqRepr` itself will not need to change and everyone else can also benefit.

An empty string is fine I think, no need to put any sort of `"Empty Directory"` message: the user should be able to tell that it was empty by virtue of the fact that nothing shows up.

The travis build looks flaky; I doubt it's your fault it's broken. You can tell because the scattering of passed/failed builds appears random, and all of them seem broken for nondeterministic reasons.

Lastly, could you add a message to the changelog in the readme? It's at the bottom of `Footer.scalatex`

Thanks again!!!

coderabhishek:
I have implemented a better fix for this bug in FrontEndUtils.tabulate function

lihaoyi:
Green enough for me, thanks!

PR #332, opened 2016-01-29
Create Conscript launch configurations

andrelfpinto:
Conscript (https://github.com/n8han/conscript) is a tool for installing and updating Scala software programs.
The launchconfig file contains the launch configurations used by Conscript to install and update the Ammonite-Shell.
It would be simple as that:
`cs lihaoyi/Ammonite`
But at the moment it must be:
`cs andrelfpinto/Ammonite --branch topic/conscript`

### DIFF
--- a/src/main/conscript/amm/launchconfig
+++ b/src/main/conscript/amm/launchconfig
@@ -1,0 +1,11 @@
+[app]
+  version: 0.5.2
+  org: com.lihaoyi
+  name: ammonite-shell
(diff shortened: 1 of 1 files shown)


lihaoyi:
Honestly I don't think this is useful. Ammonite has its own install isntructions, brew install, and I don't think installing via `cs` provides us any real value over them. Given the lack of follow-up nagging, I suppose it can't be that high a priority for you either, so I'm closing this for now...

Issue #297, opened 2015-12-10
Accidentally pressing the down arrow too many times loses edits


dickwall:
During some demos during training last week, I used the ammonite editing to go back and correct mistakes or simply update things from a previous demo. Often I would hit the down arrow just once too many times, at which point when you press up again all of the changes have disappeared and you have to make them again. It would be great if either ammonite could remember those changes, or prevent accidentally going past the end in a hurry.

lihaoyi:
I've been bitten by this multiple times before. The solution I like would be to make Ammonite remember the edits you make to any history-access, at least until you press `[enter]` and submit the command to run. 

As a plus, this would make the behavior more or less similar to Bash and other REPLs.

lihaoyi:
This might actually be "fixed" by https://github.com/lihaoyi/Ammonite/pull/281, which won't throw away the current input text if you press up/down while typing...

lihaoyi:
This is fixed by https://github.com/lihaoyi/Ammonite/pull/335

PR #335, opened 2016-01-30
History search

lihaoyi:
Re-implementation of https://github.com/lihaoyi/Ammonite/pull/281, but better

- Searches contents, not just prefixes
- Underlines things which are being searched for, making it relatively clear (?) when a search is taking place and when it isn't
- Relatively intuitive (?) behavior around starting, ending, and navigating through the search history. In particular:
  - It supports both `UP` and `Ctrl-R` as ways to start a search
  - Printable characters or backspaces modify the search in place, and narrow it in the case of additional characters.
  - Any control character ends the search and forwards to the default behavior; e.g. `ENTER` ends the search and executes the command, `LEFT`/`RIGHT` end the search and move the cursor, etc.
- Only uses two `var`s instead of seven

Misc changes 
- Upgrade pprint to allow nicely format config files
- disable travis cache because it's misbehaving =/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,12 +11,12 @@ script:
 - sbt ++$TRAVIS_SCALA_VERSION tested/test
 # Tricks to avoid unnecessary cache updates, from
 # http://www.scala-sbt.org/0.13/docs/Travis-CI-with-sbt.html
-- find $HOME/.sbt -name "*.lock" | xargs rm
-- find $HOME/.ivy2 -name "ivydata-*.properties" | xargs rm
-- find $HOME/.ivy2/cache/com.lihaoyi -name "ammonite-*" | xargs rm -rf
+#- find $HOME/.sbt -name "*.lock" | xargs rm
+#- find $HOME/.ivy2 -name "ivydata-*.properties" | xargs rm
+#- find $HOME/.ivy2/cache/com.lihaoyi -name "ammonite-*" | xargs rm -rf
 sudo: false
 # These directories are cached to S3 at the end of the build
-cache:
-  directories:
-    - $HOME/.ivy2/cache
-    - $HOME/.sbt/boot/
+#cache:
+#  directories:
+#    - $HOME/.ivy2/cache
+#    - $HOME/.sbt/boot/
--- a/build.sbt
+++ b/build.sbt
@@ -113,8 +113,8 @@ lazy val repl = project
       "org.scala-lang" % "scala-reflect" % scalaVersion.value,
       "org.apache.ivy" % "ivy" % "2.4.0",
       "com.lihaoyi" %% "scalaparse" % "0.3.4",
-      "com.lihaoyi" %% "upickle" % "0.3.6",
-      "com.lihaoyi" %% "pprint" % "0.3.6",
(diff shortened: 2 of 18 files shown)


lihaoyi:
@olafurpg want to give this a whirl and let me know what you think?

olafurpg:
Absolutely! However, I'm traveling without my laptop until tomorrow evening so I can't provide feedback before then.

lihaoyi:
Tomorrow is fine =)

olafurpg:
First of all, amazing work. I can't wait to get this :smile_cat: 

> Searches contents, not just prefixes

Being used to oh-my-zsh, it's a bit unusual at first. However, it became natural quickly. 

> Underlines things which are being searched for, making it relatively clear (?) when a search is taking place and when it isn't.

I think it's intuitive. Pressing right/left to exit the search is perfect.

> Relatively intuitive (?) behavior around starting, ending, and navigating through the search history.

Nitpick, consider
```
@ 123
res16: Int = 123
@ 123<UP> ...cannot be found in the history
```
123 is in history. Maybe, the message should rather be "...reached top of history."?

```
@ 123
res16: Int = 123
@ 12<UP,UP,UP,...>
```
cycles between 12 and 123 with no hint that you're cycling the history (with a longer history it might not be so obvious). Maybe use the same "...reached top of history." message when you reach 123?

Finally, I would argue that #297 still needs a bit work. Consider:

```
@ val s = "hello"
s: String = "hello"
@ <UP>val s = "hello"<s/hello"/goodbye"/><DOWN><UP><ENTER>
s: String = "hello"
```
I would expect after the `<DOWN><UP>` that my edit from hello to goodbye would be preserved. It's a shame to lose an edit because I accidentally pressed down (happens a lot while editing multiline strings!).

lihaoyi:
> 123 is in history. Maybe, the message should rather be "...reached top of history."?

Yeah, or it should just scroll around,

>  I would expect after the <DOWN><UP> that my edit from hello to goodbye would be preserved. It's a shame to lose an edit because I accidentally pressed down (happens a lot while editing multiline strings!).

Eh? I thought it would keep the thing after making edits, but it clearly doesn't. That's a bug. My expected behavior is that after editing the text, you wouldn't be able to go `<DOWN>`  anymore

olafurpg:
> My expected behavior is that after editing the text, you wouldn't be able to go <DOWN> anymore.

The bug happens when I press up *without* a search term. If I have a search term it works as intended.

Also, in b6ef733ab2c67301e27a8061448f834fb214a753:

* left/right doesn't exit from search anymore.
*  From an empty command `<ctrl-r><enter>` gives me

```
@  ...enter the string to search, then `up` for more results
SyntaxError: found "\u001b[34m...enter the st", expected TopPkgSeq | Import | Prelude ~ BlockDef | Expr | End at index 1
```

lihaoyi:
> left/right doesn't exit from search anymore.

Hmm yeah. The original idea was that you when browsing history without a search term, you should be able to navigate around before making an edit to drop out. The behavior should differ between "empty history browsing" and "searching for term"

> From an empty command <ctrl-r><enter> gives me

I don't see this... it's definitely a bug and i'll fix it if I can repro

PR #344, opened 2016-02-09
Capture exceptions and expose them to the REPL as repl.lastException

coderabhishek:
Last Exception due to ThreadDeath etc. can be accessed and analysed by calling repl.lastException.


repl/src/main/scala/ammonite/repl/interp/Evaluator.scala
```diff
@@ -34,7 +34,9 @@ trait Evaluator{
   def processLine(code: String,
                   printCode: String,
                   printer: Printer,
-                  extraImports: Seq[ImportData] = Seq()): Res[Evaluated]
+                  extraImports: Seq[ImportData] = Seq(),
+                  exception_update:(Throwable) => Unit
```
lihaoyi: `onException` would probably be a better name.

Also, space after the `:` and no parens around the `Throwable`

repl/src/main/scala/ammonite/repl/frontend/ReplAPI.scala
```diff
@@ -50,7 +50,7 @@ trait ReplAPI {
    * Display help text if you don't know how to use the REPL
    */
   def help: String
-
+  def lastException:Throwable
```
lihaoyi: Space after the `:`

repl/src/main/scala/ammonite/repl/interp/Interpreter.scala
```diff
@@ -56,6 +56,8 @@ class Interpreter(prompt0: Ref[String],
       |""".stripMargin
 
   val SheBang = "#!"
+  var lastEx:Throwable=null
```
lihaoyi: Call this `lastException`, and spaces after `:` and around the `:`

repl/src/main/scala/ammonite/repl/interp/Interpreter.scala
```diff
@@ -174,7 +177,7 @@ class Interpreter(prompt0: Ref[String],
     }
   }
 
-  def handleOutput(res: Res[Evaluated]) = {
+  def   handleOutput(res: Res[Evaluated]) = {
```
lihaoyi: Revert this

repl/src/main/scala/ammonite/repl/interp/Interpreter.scala
```diff
@@ -230,6 +233,8 @@ class Interpreter(prompt0: Ref[String],
 
   lazy val replApi: ReplAPI = new DefaultReplAPI { outer =>
 
+    def lastException=lastEx
```
lihaoyi: Spaces around the `:`

lihaoyi:
Overall, it seems that we're doing a lot of work to call `exception_update` inside the `Evaluator`, only to do more work to plumb the value out to the `Interpreter`. Would it be possible to make the `Interpreter` inspect the result of `processLine` and set the field itself?

repl/src/main/scala/ammonite/repl/interp/Interpreter.scala
coderabhishek: I think lastException is a more verbose name so it should be exposed in trait repl for users to call,lastEx is just an inside variable in interpreter trait to which lastException points

lihaoyi: Internal variables should be named the same way as external variables =) We shouldn't cut corners on readability just because it's internal

coderabhishek:
Do you think some better method to modify lastException than passing the partially applied function?
Please tell me

diff.patch
```diff
@@ -0,0 +1,194 @@
+diff --git a/repl/src/main/scala/ammonite/repl/Repl.scala b/repl/src/main/scala/ammonite/repl/Repl.scala
```
lihaoyi: Can you remove this from the commit?

PR #349, opened 2016-02-23
skip over the scct plugin since it doesn't work wtih ammonite

vmarquez:
We use scct at Verizon and it doesn't seem to play nice with Ammonite.


lihaoyi:
lol that `runsRightAfter` thing sure is a problem... I wonder if I should try for a real fix rather than all these hardcoded hacks

PR #351, opened 2016-02-23
Fix Stack overflow when displaying large instances of LsSeq

### LINKED ISSUE #341: Stack overflow when displaying large instances of LsSeq
Displaying LsSeq (e.g. a result of ls.rec!) creates a stack overflow in ammonite.repl.Util.transpose.
This is to be expected as the transpose implementation there is recursive.

I would suggest to replace the implementation by this tail recursive variant:

```scala
@scala.annotation.tailrec
def transpose[A](xs: List[List[A]], result: List[List[A]]): List[List[A]] =
  xs.filter(_.nonEmpty) match {
    case Nil    =>  result
    case ys: List[List[A]] => transpose(ys.map{ _.tail }, ys.map{ _.head }::result)
  }
def transpose[A](xs: List[List[A]]): List[List[A]] = transpose(xs, Nil).reverse
```

rauf:
Fixes #341

### DIFF
--- a/readme/Footer.scalatex
+++ b/readme/Footer.scalatex
@@ -42,6 +42,12 @@
       Although it's best to read the documentation on this page to learn how to use these projects, the Scaladoc is still useful as a reference.
   
   @sect{Changelog}
+
+    @sect{0.5.6}
+      @ul
+        @li
+          Displaying large instances of paths does not cause stack overflow. (#351)
+
     @sect{0.5.5}
       @ul
         @li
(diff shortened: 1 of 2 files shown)


rauf:
@lihaoyi  Need help with the unit test.

lihaoyi:
Superseded by https://github.com/lihaoyi/Ammonite/commit/a6d184fcad9a0b4c5bea26f405f36dd905a1d1ba

PR #358, opened 2016-03-10
build with 2.11.8

ckampfe:
@lihaoyi thanks for Ammonite, love it!

looking to get the conversation started for 2.11.8, and figure out what else should be done :)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ language: scala
 scala:
 - 2.10.5
 - 2.11.4
-- 2.11.7
+- 2.11.8
 jdk:
 - oraclejdk7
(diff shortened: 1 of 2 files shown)


lihaoyi:
@ckampfe can you also sub out `2.11.7` for `2.11.8` in the `.travis.yml` file?

https://github.com/lihaoyi/Ammonite/blob/master/.travis.yml#L5

I'd like to bracket the possible builds and make sure Ammonite passes both on the latest and on an older one (i.e. `2.11.4`). Then travis will make sure that the build on 2.11.8 is successful

lihaoyi:
Looks good to me, thanks @ckampfe!

PR #370, opened 2016-03-31
Display proper error line numbers for errors caught while building AST

coderabhishek:


lihaoyi:
Yay thanks!

Two things immediately come to mind

1. For this sort of change, adding "Here's how I tested it, here's the output before this change, here's the output after" to the PR makes things a lot easier to understand

2. An integration test that runs some script using the ammonite executable, and verifies the properties we want (e.g. the error message is reasonably short, contains some key words) would help avoid regressions in future!

lihaoyi:
A third thing: the body of `processScript` is getting rather large/indented, and you just indented it two more times! Could we break out part of it (e.g. the `loop` code) into a separate function?

coderabhishek:
OK

coderabhishek:
I've updated the PR with appropriate integration test.

for this code:
`object Qs{
  def main(a: Int){
    println("OK")
    var unsorted: List[Int] = List(5,4,2,3,1)
    printlnqs(unsorted))
  }
.......`

It gives following error:

`ammonite.ops.ShelloutException: CommandResult 1
Exception in thread "main" ammonite.repl.CompilationError: ("}" | `case`):5:24 ...")\n  }\n\n  d"`

Line number is correct which was not the case earlier

integration/src/test/scala/ammonite/integration/StandaloneTests.scala
```diff
@@ -63,6 +63,31 @@ object StandaloneTests extends TestSuite{
...
+      val name = "Error_line_number_test.scala"
+      try {
+        val res = %%bash(
+          executable,
+          "--predef-file",
+          emptyPrefdef,
+          replStandaloneResources / name
+          )
+      }
+      catch
```
lihaoyi: Format this as `} catch {`

integration/src/test/scala/ammonite/integration/StandaloneTests.scala
```diff
@@ -63,6 +63,31 @@ object StandaloneTests extends TestSuite{
       println("\n-------------------------")
     }
 
+    'error_test{
+      val name = "Error_line_number_test.scala"
+      try {
```
lihaoyi: You should use [intercept](https://github.com/lihaoyi/utest#intercept) here to make sure that if the script *doesn't* throw, the test fails

integration/src/test/scala/ammonite/integration/StandaloneTests.scala
```diff
@@ -63,6 +63,31 @@ object StandaloneTests extends TestSuite{
...
+        val res = %%bash(
```
lihaoyi: No need to assign `val res =` if we're never gonna use it

integration/src/test/scala/ammonite/integration/StandaloneTests.scala
```diff
@@ -63,6 +63,31 @@ object StandaloneTests extends TestSuite{
...
+        {      // expected_error_msg is expected error message for test file
+          case ex: Throwable => val expected_error_msg = """ammonite.ops.ShelloutException:
+                                                           |CommandResult 1##
+                                                           |Exception in thread "main" ammonite&&
+                                                           |.repl.CompilationError: ("}" | `case`)&&
+                                                           |:5:24 ...")\n  }\n\n  d"""".stripMargin.replaceAll("\n"," ")
+                                                           .replaceAll("## ","\n").replaceAll("&& ","")
+
+                                println("Error thrown by test file:\n" + ex.toString.substring(0,145) + "\n------------------\n\n")
+                                assert(ex.toString.substring(0,145) == expected_error_msg)
```
lihaoyi: Where's this magic number `145` coming from?

integration/src/test/scala/ammonite/integration/StandaloneTests.scala
```diff
@@ -63,6 +63,31 @@ object StandaloneTests extends TestSuite{
...
+          )
+      }
+      catch
+        {      // expected_error_msg is expected error message for test file
+          case ex: Throwable => val expected_error_msg = """ammonite.ops.ShelloutException:
+                                                           |CommandResult 1##
+                                                           |Exception in thread "main" ammonite&&
+                                                           |.repl.CompilationError: ("}" | `case`)&&
+                                                           |:5:24 ...")\n  }\n\n  d"""".stripMargin.replaceAll("\n"," ")
+                                                           .replaceAll("## ","\n").replaceAll("&& ","")
```
lihaoyi: If you format the code as described above you won't need all these we're `replaceAll` calls

k.scala
```diff
@@ -0,0 +1 @@
+load.module("/home/abhishek/qs.scala")
```
lihaoyi: What's this?

coderabhishek:
Error report for 

`object Qs{
  def main(a: Int){
    println("OK"))
    var unsorted: List[Int] = List(5,4,2,3,1)
    printlnqs(unsorted))
  }
`
is  (correct line numbers with caret at error position):
`ammonite.repl.CompilationError: ("}" | `case`):3:18 ...")\n    var "
    println("OK"))
                 ^
  ammonite.repl.interp.Interpreter.processScript(Interpreter.scala:198)
  ammonite.repl.interp.Interpreter.processModule(Interpreter.scala:105)
  ammonite.repl.interp.Interpreter$$anon$1$load$.module(Interpreter.scala:286)
  cmd0$.<init>(Main.scala:275)
  cmd0$.<clinit>(Main.scala:-1)`

compared to earlier :
`fastparse.core.ParseError: found ")\n    var unsorted: ", expected ";" | Newline.rep(1) | "}" | `case` at index 445
    println("OK"))
                 ^
  fastparse.core.Parsed$class.get(Parsing.scala:29)
  fastparse.core.Parsed$Failure.get(Parsing.scala:84)
  ammonite.repl.Parsers$.splitScript(Parsers.scala:47)
  ammonite.repl.interp.Interpreter.processScript(Interpreter.scala:120)
  ammonite.repl.interp.Interpreter.processModule(Interpreter.scala:97)
  ammonite.repl.interp.Interpreter$$anon$1$load$.module(Interpreter.scala:262)
  cmd2$.<init>(Main.scala:313)
  cmd2$.<clinit>(Main.scala:-1)`

lihaoyi:
Looks good, thanks!

Issue #319, opened 2016-01-12
Infix type printing doesn't show grouping


nafg:
``` scala
naftoli-platinum@ trait *[A, B]
defined trait *
naftoli-platinum@ trait +[A, B]
defined trait +
naftoli-platinum@ null: (Int + Int * Int)
res2: Int + Int * Int = null
naftoli-platinum@ null: (Int + (Int * Int))
res3: Int + Int * Int = null
naftoli-platinum@ null: ((Int + Int) * Int)
res4: Int + Int * Int = null
naftoli-platinum@ (null: ((Int + Int) * Int)): (Int + (Int * Int))
Compilation Failed
Main.scala:132: type mismatch;
 found   : cmd0.*[cmd1.+[Int,Int],Int]
 required: cmd1.+[Int,cmd0.*[Int,Int]]
(null: ((Int + Int) * Int)): (Int + (Int * Int))
```

lihaoyi:
Doesn't seem too hard, would be nice if someone did it

lihaoyi:
Probably a wontfix; if you want this to happen send the PR to https://github.com/lihaoyi/pprint

PR #377, opened 2016-04-14
Infix type printing now shows grouping

ulasakdeniz:
Fixes: #319 
Thought that was an easy one, not sure if it's a correct solution though 😄

### DIFF
--- a/repl/src/main/scala-2.11/ammonite/repl/frontend/TPrintImpl.scala
+++ b/repl/src/main/scala-2.11/ammonite/repl/frontend/TPrintImpl.scala
@@ -172,7 +172,7 @@ object TPrintLowPri{
       case TypeRef(pre, sym, List(left, right))
         if lookup(sym) && sym.name.encodedName.toString != sym.name.decodedName.toString =>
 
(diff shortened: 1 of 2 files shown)


lihaoyi:
Yeah this is probably the wrong solution. We want the infix operations, but we want parens around them (only where necessary!) to maintain the correct infix order of nesting.

ulasakdeniz:
If you give me more detail I can work on it.

lihaoyi:
The original test case still applies
```scala
naftoli-platinum@ trait *[A, B]
defined trait *
naftoli-platinum@ trait +[A, B]
defined trait +
naftoli-platinum@ null: (Int + Int * Int)
res2: Int + Int * Int = null
naftoli-platinum@ null: (Int + (Int * Int))
res3: Int + Int * Int = null
naftoli-platinum@ null: ((Int + Int) * Int)
res4: Int + Int * Int = null
```

`(Int + (Int * Int))` and `((Int + Int) * Int)` both are rendered as `Int + Int * Int`, but are different types. Ideally the second case would render as `(Int + Int) * Int`

ulasakdeniz:
I thought that it should give output like Scala REPL, I can take a look at it when I'm available, thanks.

PR #378, opened 2016-04-16
Configure external links in ScalaDoc

Atry:
I found that the [ScalaDoc](http://www.lihaoyi.com/Ammonite/api/ops/index.html) does not link to Scala standard library and other dependencies. This patch fixes the problem.


lihaoyi:
lgtm

PR #390, opened 2016-05-24
Add Line Number correction support for compilation error in scripts

coderabhishek:
Code for line number correction for compilation error in scripts along with tests


integration/src/test/scala/ammonite/integration/StandaloneTests.scala
```diff
@@ -124,6 +124,83 @@ object StandaloneTests extends TestSuite{
       assert(e.toString.contains(expectedErrorMsg))
     }
 
+    'compilationErrorLineNumberTest1{
+      //Make sure correct line numbers are printed when an erroneous script is executed
+      val name = "compilationErrorLineNumberTest1.scala"
```
lihaoyi: Can you extract all the common code for these 4 test cases out into a helper function?

integration/src/test/resources/ammonite/integration/compilationErrorLineNumberTest1.scala
```diff
@@ -0,0 +1,30 @@
+
```
lihaoyi: Could you give these test cases slightly more descriptive names? e.g. "CompileErrorDefinition.scala", "CompileErrorWithSyntheticRes.scala" etc.

lihaoyi:
Could we add a few more test cases:

- Runtime errors: i.e. when you throw an exception, it should show the right line numbers. Given how you modified the AST I think this should work out of the box, but it's worth verifying
- Compilation error in the first block before the first `@`: you have tests for compilation error in the second block and in the third block, but let's verify that an error in the first block works too

repl/src/main/scala/ammonite/repl/interp/AmmonitePlugin.scala
```diff
@@ -24,13 +27,30 @@ class AmmonitePlugin(g: scala.tools.nsc.Global, output: Seq[ImportData] => Unit)
 
       def newPhase(prev: Phase): Phase = new g.GlobalPhase(prev) {
         def name = phaseName
-        def apply(unit: g.CompilationUnit): Unit = AmmonitePlugin(g)(unit, output)
+        def apply(unit: g.CompilationUnit): Unit = AmmonitePlugin(g)(unit, output, importsLen)
       }
+    },
+
+  new PluginComponent {
```
lihaoyi: This should be indented two more spaces to the right

coderabhishek:
ok I will update the PR soon

integration/src/test/scala/ammonite/integration/StandaloneTests.scala
```diff
@@ -124,6 +124,83 @@ object StandaloneTests extends TestSuite{
...
+      val e = intercept[ShelloutException]{
+        %%bash(
+          executable,
+          "--predef-file",
+          emptyPrefdef,
+          replStandaloneResources / name
+          )
+      }
+      val expectedErrorMsg =
+        """Main.scala:14: not found: value printnl
```
lihaoyi: Could you change the file-name down below where you update the positions so it has the actual file name instead of `Main.scala`?

repl/src/main/scala/ammonite/repl/interp/AmmonitePlugin.scala
```diff
@@ -150,4 +170,49 @@ object AmmonitePlugin{
 
     output(open.toVector)
   }
+}
+
+
+object LineNumberModifier {
+  def apply(g: Global)(unit: g.CompilationUnit,
+                       output: Seq[ImportData] => Unit,
+                       importsLen: => Int) = {
```
lihaoyi: Let's call this `topWrapperLen` or something, since it's not just the imports but also includes the `object Foo{\n` at the top of the generated file

repl/src/main/scala/ammonite/repl/interp/AmmonitePlugin.scala
```diff
@@ -150,4 +170,49 @@ object AmmonitePlugin{
...
+              val p = new scala.reflect.internal.util.OffsetPosition(con, s.point - importsLen)
+              transformedTree.pos = p
+            }
+          case _ =>    //for position  = NoPosition
+        }
+        transformedTree
+      }
+    }
+
+    class TreePrinter(unit: CompilationUnit) extends Transformer {
```
lihaoyi: If this isn't being used except for debugging let's remove it.

repl/src/main/scala/ammonite/repl/interp/Compiler.scala
```diff
@@ -249,13 +250,15 @@ object Compiler{
      * Compiles a blob of bytes and spits of a list of classfiles
      */
     def compile(src: Array[Byte],
-                printer: Printer): Output = {
+                printer: Printer,
+                impLen: Int): Output = {
```
lihaoyi: Can we call this `importsLen0` instead of `impLen` so it's more obvious that it's related to `this.importsLen`? And let's update the comment above to briefly mention why we need to pass it in (so we can properly update the positions of the generated ASTs for better errors)

repl/src/main/scala/ammonite/repl/interp/Evaluator.scala
```diff
@@ -300,17 +306,18 @@ $code
...
-      )
-    } yield {
-      Timer("cachedCompileBlock")
-      evalMain(cls)
-      Timer("evalMain")
-      val res = evaluationResult(wrapperName, newImports)
-      Timer("evaluationResult")
-      res
-    }
+        (wrapperName, cls, newImports) <- cachedCompileBlock(
```
lihaoyi: Let's keep the old format

repl/src/main/scala/ammonite/repl/interp/Interpreter.scala
```diff
@@ -103,12 +103,13 @@ class Interpreter(prompt0: Ref[String],
   }
 
   def processModule(code: String) = processScript(
-    skipSheBangLine(code),
-    (code, imports) => withContextClassloader(eval.processScriptBlock(code, imports, printer))
-  )
+      skipSheBangLine(code),
```
lihaoyi: Unnecessary Formatting

repl/src/main/scala/ammonite/repl/interp/Interpreter.scala
```diff
@@ -103,12 +103,13 @@ class Interpreter(prompt0: Ref[String],
...
+      (code, imports) => withContextClassloader(eval.processScriptBlock(code, imports, printer))
+    )
+
 
   def processExec(code: String) =
-    processScript(skipSheBangLine(code), { (c, i) => evaluateLine(c, Nil, printer, i)})
+    processScript(skipSheBangLine(code), { (c, i) => evaluateLine(c, Nil, printer, i) })
```
lihaoyi: Unnecessary Formatting

repl/src/main/scala/ammonite/repl/interp/Interpreter.scala
```diff
@@ -135,12 +136,26 @@ class Interpreter(prompt0: Ref[String],
     s"Syntax Error: $msg\n$locationString"
   }
 
-  def processCorrectScript(parsedCode: Parsed.Success[Seq[Seq[String]]],
+  def processCorrectScript(rawParsedCode: Parsed.Success[Seq[( String, Seq[String])]],
```
lihaoyi: No space between the `(` and `String`

repl/src/main/scala/ammonite/repl/interp/Preprocessor.scala
```diff
@@ -99,30 +99,36 @@ object Preprocessor{
...
     }
     
-    def complete(code: String, wrapperId: String, postSplit: Seq[String]) = {
+    def complete(code: String, wrapperId: String, postSplit: Seq[String], comment: String) = {
       val reParsed = postSplit.map(p => (parse(p), p))
       val errors = reParsed.collect{case (Left(e), _) => e }
       if (errors.length != 0) Res.Failure(None, errors.mkString("\n"))
       else {
-        val allDecls = for (((Right(trees), code), i) <- reParsed.zipWithIndex) yield {
+        var allDecls = for (((Right(trees @ x :: y), code), i) <- reParsed.zipWithIndex) yield {
```
lihaoyi: `allDecls` shouldn't need to be a `var` as far as I can tell

repl/src/main/scala/ammonite/repl/interp/Interpreter.scala
```diff
@@ -135,12 +136,26 @@ class Interpreter(prompt0: Ref[String],
...
+    }
 
-    val blocks0 = Parsers.splitScript(hardcodedPredef).get.value ++ parsedCode.get.value
-    val blocks = blocks0.map(preprocess(_, ""))
+    val parsedHardcodedPredef  = Parsers.splitScript(hardcodedPredef).get.value
+    val blocks0 = parsedHardcodedPredef ++ parsedCode
+
+    val blocks = for{
+      (comment, code) <- blocks0
+    } yield preprocess(code,"",comment)
```
lihaoyi: Could you describe to me what purpose `comment` serves here? It's not quite clear to me why we need to pass it around separately from `code`, though I'm sure you had a reason for doing so

repl/src/main/scala/ammonite/repl/interp/AmmonitePlugin.scala
```diff
@@ -150,4 +170,49 @@ object AmmonitePlugin{
 
     output(open.toVector)
   }
+}
+
+
+object LineNumberModifier {
+  def apply(g: Global)(unit: g.CompilationUnit,
+                       output: Seq[ImportData] => Unit,
```
lihaoyi: `output` seems like it's not being used here

repl/src/main/scala/ammonite/repl/interp/AmmonitePlugin.scala
```diff
@@ -150,4 +170,49 @@ object AmmonitePlugin{
...
+                       importsLen: => Int) = {
+
+    import g._
+
+    class LineNumberCorrector(unit: CompilationUnit) extends Transformer {
```
lihaoyi: Since we're only ever instantiating `LineNumberCorrector` once, could we make it an `object` instead of a `class`?

repl/src/main/scala/ammonite/repl/interp/AmmonitePlugin.scala
```diff
@@ -150,4 +170,49 @@ object AmmonitePlugin{
...
+      override def transform(tree: Tree) = {
+        val transformedTree = super.transform(tree)
+        tree.pos match {
+          case s: scala.reflect.internal.util.OffsetPosition =>
```
lihaoyi: Let's import `scala.reflect.internal.util.{OffsetPosition, BatchSourceFile}` at the top so we can keep our code a bit more concise

repl/src/main/scala/ammonite/repl/interp/Interpreter.scala
```diff
@@ -135,12 +136,26 @@ class Interpreter(prompt0: Ref[String],
     s"Syntax Error: $msg\n$locationString"
   }
 
-  def processCorrectScript(parsedCode: Parsed.Success[Seq[Seq[String]]],
+  def processCorrectScript(rawParsedCode: Parsed.Success[Seq[( String, Seq[String])]],
                            evaluate: EvaluateCallback): Seq[ImportData] = {
 
+    var offset = 0
+    var parsedCode: Seq[( String, Seq[String])] = Seq()
```
lihaoyi: space between `(` and `String`

repl/src/main/scala/ammonite/repl/interp/Interpreter.scala
```diff
@@ -135,12 +136,26 @@ class Interpreter(prompt0: Ref[String],
...
+    for( (comment, code) <- rawParsedCode.get.value ){
+      val ncomment = comment + "\n"*offset
+      offset = offset + comment.count(_ == '\n') + code.foldLeft(0){ (z,x) =>
```
lihaoyi: Could you replace this `foldLeft` construct with `code.map(_.count(_ == '\n')).sum`?

repl/src/main/scala/ammonite/repl/interp/Interpreter.scala
```diff
@@ -135,12 +136,26 @@ class Interpreter(prompt0: Ref[String],
...
+        z + x.count(_ == '\n')
+      } + 1
```
lihaoyi: What's the `+ 1` at the end for? Let's leave a short comment so people who see this magic number know why it exists and must be `1` (and not `2` or `-1` or some other number)

lihaoyi:
That's all for now. Lots of nitpicks, but overall this is a good PR. Nice work! You're now in the comparatively small number of people who have written a Scala compiler plugin 👅 

The only thing I haven't fully understood is the parsing logic and the logic inside `processCorrectScript`; left a comment for you to elaborate on it and let's discuss it.

repl/src/main/scala/ammonite/repl/interp/Interpreter.scala
coderabhishek: The comment or empty lines at top of any code is caught by WL in compilationUnit parser, so splitScript return (String, Seq[String]) that is comments separately, since preprocessor recombines the parsed values into code it skips comments that is why I keep comments separately and combine them in complete method

repl/src/main/scala/ammonite/repl/interp/Interpreter.scala
```diff
@@ -107,12 +107,19 @@ class Interpreter(prompt0: Ref[String],
 
   def processModule(code: String, fileName: String = "Main.scala") = processScript(
     skipSheBangLine(code),
-    (code, imports) => withContextClassloader(eval.processScriptBlock(code, imports, printer, fileName))
-    )
+    (code, imports) => withContextClassloader(eval.processScriptBlock(code,
```
coderabhishek: Not sure if this is a good formatting style but I found it most natural to avoid too long lines.

repl/src/main/scala/ammonite/repl/interp/Interpreter.scala
```diff
@@ -107,12 +107,19 @@ class Interpreter(prompt0: Ref[String],
...
+                                                                      imports,
+                                                                      printer,
+                                                                      fileName))
+  )
```
coderabhishek: Not sure if this is a good formatting style but I found it most natural to avoid too long lines. Is it ok?

repl/src/main/scala/ammonite/repl/interp/Preprocessor.scala
```diff
@@ -99,30 +99,36 @@ object Preprocessor{
...
     }
     
-    def complete(code: String, wrapperId: String, postSplit: Seq[String]) = {
+    def complete(code: String, wrapperId: String, postSplit: Seq[String], comment: String) = {
       val reParsed = postSplit.map(p => (parse(p), p))
       val errors = reParsed.collect{case (Left(e), _) => e }
       if (errors.length != 0) Res.Failure(None, errors.mkString("\n"))
       else {
-        val allDecls = for (((Right(trees), code), i) <- reParsed.zipWithIndex) yield {
+        val allDecls = for (((Right(trees @ x :: y), code), i) <- reParsed.zipWithIndex) yield {
```
lihaoyi: `x :: y` seems un-used to me, can we delete it?

repl/src/main/scala/ammonite/repl/interp/Interpreter.scala
```diff
@@ -135,12 +146,27 @@ class Interpreter(prompt0: Ref[String],
     s"Syntax Error: $msg\n$locationString"
   }
 
-  def processCorrectScript(parsedCode: Parsed.Success[Seq[Seq[String]]],
+  def processCorrectScript(rawParsedCode: Parsed.Success[Seq[(String, Seq[String])]],
                            evaluate: EvaluateCallback): Seq[ImportData] = {
 
+    var offset = 0
+    var parsedCode: mutable.ArrayBuffer[( String, Seq[String])] = mutable.ArrayBuffer()
```
lihaoyi: No need a type annotation here, and let's use `mutable.Buffer` instead of `mutable.ArrayBuffer` since that's what everyone else seems to use (even though it's the same)

repl/src/main/scala/ammonite/repl/interp/Interpreter.scala
```diff
@@ -135,12 +146,27 @@ class Interpreter(prompt0: Ref[String],
...
+
+    // comment holds comments or empty lines above the code which is not caught along with code
+    for( (comment, code) <- rawParsedCode.get.value ){
+      val ncomment = comment + "\n"*offset
+
+      // 1 is added as Separator parser eats up the '\n' following @
+      offset = offset + comment.count(_ == '\n') + code.map(_.count(_ == '\n')).sum + 1
+      parsedCode = parsedCode :+ (ncomment, code)
```
lihaoyi: Since we're using `mutable.{ArrayBuffer, Buffer}` above, `:+` is O(n) time and thus the for-loop is O(n^2). Use `.append` instead which is O(1)

repl/src/main/scala/ammonite/repl/interp/Interpreter.scala
```diff
@@ -97,18 +99,27 @@ class Interpreter(prompt0: Ref[String],
...
         extraImports
       )
 
   }
 
-  def processModule(code: String) = processScript(
+  def processModule(code: String, fileName: String = "Main.scala") = processScript(
     skipSheBangLine(code),
-    (code, imports) => withContextClassloader(eval.processScriptBlock(code, imports, printer))
+    (code, imports) => withContextClassloader(eval.processScriptBlock(code,
```
lihaoyi: Could we instead format this as

```scala
(code, imports) => 
  withContextClassloader(
    eval.processScriptBlock(code, imports, printer, filename)
  )
```

A bit of a nitpick but I think it looks a lot better

repl/src/main/scala/ammonite/repl/interp/Interpreter.scala
```diff
@@ -97,18 +99,27 @@ class Interpreter(prompt0: Ref[String],
...
+                                                                      imports,
+                                                                      printer,
+                                                                      fileName))
   )
 
+
   def processExec(code: String) =
-    processScript(skipSheBangLine(code), { (c, i) => evaluateLine(c, Nil, printer, i)})
+    processScript(skipSheBangLine(code), { (c, i) => evaluateLine(c,
```
lihaoyi: And let's format this as 

```scala
processScript(
  skipSheBangLine(code), 
  { (c, i) => evaluateLine(c, Nil, printer, "Main.scala", i) }
)
```

In general, try to avoid vertical alignment everywhere except class/function signatures. If method/function calls or assignments or other expressions are too long, 

- Kick them to the next line
- Indent the wrapped lines by 2
- Add the necessary closing paren/curly on a new line, not indented.

Basically try to format them about the same as you'd format the multi-line bodies of functions or classes: 2-space indent, one-thing-per-line inside, closing bracket on new line.

lihaoyi:
Last few nitpicks then it'll be ready to merge!

repl/src/main/scala/ammonite/repl/interp/Preprocessor.scala
coderabhishek: Otherwise it can take empty trees also and was giving error accessing trees(0) in this line
`   case trees if trees.forall(_.isInstanceOf[G#Import]) => handleTree(trees(0)) `

lihaoyi: oh that makes sense. in that case could we use a `if trees.nonEmpty` check
instead?

repl/src/main/scala/ammonite/repl/interp/Interpreter.scala
coderabhishek: Without type it seems we can not create an empty Buffer either type or initializing value is required

lihaoyi:
```scala
val parsedCode = mutable.Buffer[(String, Seq[String]))]()
```
should work

lihaoyi:
Looks good to me, good work!

PR #391, opened 2016-05-26
preserve settings when replacing compilers

tpolecat:
It is currently not possible to have a predef that manipulates compiler settings, because the compiler that is used during initialization is replaced before the REPL is available to the user. This PR fixes that as follows:

- `initGlobalBits ` now takes a `Settings` to use, rather than constructing a new one.
- `Compiler` and `Pressy` now take a `Settings`, which is passed to `initGlobalBits`.
- `Interpreter` now constructs a new `Settings` for the initial compiler, and thereafter uses a *copy* of the settings from the previous compiler.

The provided unit test fails in `master` at the moment, but passes with these changes.


tpolecat:
Build failures seem to be unrelated. They look like Ivy failures (master is failing as well) … what's your strategy for this?

lihaoyi:
Looks great. Thanks @tpolecat!

Issue #300, opened 2015-12-11
load.ivy fails to resolve transitive dependencies for locally published artifacts


holgerbrandl:
Using the latest 0.5.1 amm, dependency resolution for locally deployed artifacts does not seem to resolve sub-dependcies. Here's a SSCE

```
## clean up cache and previoulsy deployed versions of our artifact and also the ammonite cache
rm -rf  ~/.ivy2/cache/de.mpicbg.scicomp/ ~/.ivy2/local/de.mpicbg.scicomp/ ~/.ammonite/cache
git clone https://github.com/holgerbrandl/scalautils.git
cd scalautils
sbt publishLocal

## start ammonite shell
amm

## load scalautils dependency
load.ivy("de.mpicbg.scicomp" %% "scalautils" % "0.1-SNAPSHOT")

## load a class from a transitive dependency
import better.files.File
```

And this is a run protocol

```
Desktop]$ rm -rf  ~/.ivy2/cache/de.mpicbg.scicomp/ ~/.ivy2/local/de.mpicbg.scicomp/ ~/.ammonite/cache
[Desktop]$ git clone https://github.com/holgerbrandl/scalautils.git
Cloning into 'scalautils'...
remote: Counting objects: 600, done.
remote: Compressing objects: 100% (23/23), done.
remote: Total 600 (delta 6), reused 0 (delta 0), pack-reused 570
Receiving objects: 100% (600/600), 249.86 KiB | 0 bytes/s, done.
Resolving deltas: 100% (240/240), done.
Checking connectivity... done.
[Desktop]$ cd scalautils
[scalautils]$ sbt publishLocal
[info] Loading project definition from /Users/brandl/Desktop/scalautils/project
[info] Updating {file:/Users/brandl/Desktop/scalautils/project/}scalautils-build...
[info] Resolving org.fusesource.jansi#jansi;1.4 ...
[info] Done updating.
[info] Set current project to scalautils (in build file:/Users/brandl/Desktop/scalautils/)
[info] Updating {file:/Users/brandl/Desktop/scalautils/}scalautils...
[info] Packaging /Users/brandl/Desktop/scalautils/target/scala-2.11/scalautils_2.11-0.1-SNAPSHOT-sources.jar ...
[info] Done packaging.
[info] Wrote /Users/brandl/Desktop/scalautils/target/scala-2.11/scalautils_2.11-0.1-SNAPSHOT.pom
[info] Resolving jline#jline;2.12.1 ...
[...]

PR #395, opened 2016-05-30
Fixed resolver pattern for local ivy

aishfenton:
Addresses this bug: #300 

Problem is that locally published jars in Ivy have a different pattern than they do in the remote repo. Specifically the Ivy.xml file, which lists the jar dependencies, is in a different subdirectory from the Jar. Annoyingly Ivy fails silently on this and just assumes you have no dependencies, and copies your local jar to the ivy cache with a default ivy.xml :/

I don't really grok Ivy that well, but I wonder if some of the other ivy issues are related (such as #387) to this. Maybe other patterns should be updated to include [type] and [classifier]?

### DIFF
--- a/repl/src/main/scala/ammonite/repl/IvyThing.scala
+++ b/repl/src/main/scala/ammonite/repl/IvyThing.scala
@@ -178,7 +178,7 @@ object Resolvers {
    Resolver.File(
      "local",
      "/.ivy2/local",
-     "/[organisation]/[module]/[revision]/jars/[artifact].[ext]",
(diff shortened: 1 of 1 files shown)


aishfenton:
Hmm, I assume those travis build fails are not related to my PR? I can't think how this change would result in some combinations of Scala and JDK not passing, but others passing fine.

Also I didn't include a unit test because I wasn't sure how you'd do it (you need another project to have "sbt publishLocal" first), but open to suggestions?

lihaoyi:
Looks good to me; the test failures are probably flakes. No need for unit tests here

lihaoyi:
Thanks!

PR #389, opened 2016-05-18
Changes to fix #383 (to get ammonite working with Spark)

jvimal:
- Allow repl classes to be accessed via getResourceAsStream.
- Also save compiled classes to a directory.

It is the responsibility of the caller to delete the
`ammonite.repl.class.outputdir` on exit.

There are more changes needed to get it fully working with Spark.  The
most important one being serving these classes via a HttpServer and
configuring spark so it can load these classes (closures) using
URLClassLoader.  Those changes are outside the scope of this PR.

PR #401, opened 2016-06-11
Refactor to enforce Interpreter/Evaluator/Preprocessor layering 

lihaoyi:
Previously, the `Evaluator` had all sorts of source-code-compilation related stuff in it, while the `Interpreter` had tons of source-code-mangling stuff in it too.

This shifts all the source-code-compilation stuff into `Interpreter`, and all the source-code-mangling stuff into `Preprocessor`. This lets us significantly narrow the interfaces between `Interpreter`/`Evaluator`/`Preprocessor` and reduce the number of times we're calling back and forth between them


lihaoyi:
This'll probably wait here until @jvimal's PR #389 gets settled

PR #403, opened 2016-06-16
Implement script level cache

coderabhishek:
Program flow is like:

Speed improvement is achieved by making processModule cache anything( scripts, predefs, imports including $ivy and $file,etc.) it processes at script level( at block level it already gets) and tries to retrieve from cache next time it is asked to process same code.

Process Module checks if the script is available in cacheFolder i.e. `scriptCaches`. If it is not found, processModule0 is called which keeps the rest of program flow same as before this diff but passes back importHooks and other imports accumulated in `processCorrectScript` function. This data is stored by `classFilesListSave` which takes list of `pkgName` concatenated with `wrapperName` and hash values of blocks along with imports and stores them.

If script is found in cache, `classFilesListLoad` reads the list of names and hash values of blocks and loads those blocks from their cache folders made by `compileCacheLoad` function. Loaded import hooks are resolved by `resolveSingleImportHook`. Along with retrieved imports this data is passed to `eval.evalCacheClassFiles` which loads each classFile and evals the main function thus executing the code. 

The final imports are returned by processModule function whether evaled or loaded from cache, thus processModule is opaque from outside whether cache HIT or MISS, it behaves in exactly same way in return as well as side effects except cache save.

### DIFF
--- a/repl/src/main/scala/ammonite/repl/Main.scala
+++ b/repl/src/main/scala/ammonite/repl/Main.scala
@@ -9,6 +9,9 @@ import scala.reflect.internal.annotations.compileTimeOnly
 import scala.reflect.runtime.universe.TypeTag
 import language.experimental.macros
 import reflect.macros.Context
+import scala.collection.mutable
+import scala.util._
+import Util.CompileCache
 
 
 /**
@@ -90,7 +93,7 @@ case class Main(predef: String = "",
     ) match{
       case x: Res.Failing => x
       case Res.Success(imports) =>
-        repl.interp.init()
+        repl.interp.reInit()
         imports.value.find(_.toName == Name("main")) match {
           case None => Res.Success(imports)
           case Some(i) =>
--- a/repl/src/main/scala/ammonite/repl/Storage.scala
+++ b/repl/src/main/scala/ammonite/repl/Storage.scala
@@ -2,7 +2,8 @@ package ammonite.repl
 
 import acyclic.file
 import ammonite.ops._
-import ammonite.repl.Util.{IvyMap, CompileCache, ClassFiles}
+import Parsers.ImportTree
+import ammonite.repl.Util.{IvyMap, CompileCache, ClassFiles, CacheOutput}
 import org.apache.ivy.plugins.resolver.RepositoryResolver
 
 import scala.util.Try
@@ -22,6 +23,15 @@ trait Storage{
   val ivyCache: StableRef[IvyMap]
   def compileCacheSave(path: String, tag: String, data: CompileCache): Unit
   def compileCacheLoad(path: String, tag: String): Option[CompileCache]
+  def classFilesListSave(pkg: String,
+                         wrapper: String,
+                         dataList: Seq[(String, String)],
+                         imports: Imports,
+                         tag: String,
+                         importTreesList: Seq[ImportTree]): Unit
+  def classFilesListLoad(pkg: String,
+                         wrapper: String,
+                         cacheTag: String): Option[CacheOutput]
 }
 
 object Storage{
@@ -42,6 +52,9 @@ object Storage{
     }
 
     var compileCache: mutable.Map[String, (String, CompileCache)] = mutable.Map.empty
+    val classFilesListcache = {
+      mutable.Map.empty[String, (String, Seq[(String, String)], Imports, Seq[ImportTree])]
+    }
     def compileCacheSave(path: String, tag: String, data: CompileCache): Unit = {
       compileCache(path) = (tag, data)
     }
@@ -50,6 +63,35 @@ object Storage{
         (loadedTag, data) <- compileCache.get(path)
         if loadedTag == tag
       } yield data
+    }
+
+    def classFilesListSave(pkg: String,
+                           wrapper: String,
+                           dataList: Seq[(String, String)],
+                           imports: Imports,
+                           tag: String,
+                           importTreesList: Seq[ImportTree]): Unit = {
+      val dir = pkg + "." + wrapper
+      classFilesListcache(dir) = (tag, dataList.reverse, imports, importTreesList)
+    }
+
+    def classFilesListLoad(pkg: String,
+                           wrapper: String,
+                           cacheTag: String): Option[CacheOutput] = {
+      val dir = pkg + "." + wrapper
+      classFilesListcache.get(dir) match{
+        case None => None
+        case Some((loadedTag, classFilesList, imports, importTreesList)) =>
+          if (loadedTag == cacheTag) {
+            val res = {
+              for((path, tag) <- classFilesList) yield {
+                compileCacheLoad(path, tag)
+              }
+            }.flatten
+            Some((classFilesList.unzip._1, res.unzip._1, imports, importTreesList))
+          }
+          else None
+      }
     }
   }
 
@@ -62,6 +104,7 @@ object Storage{
     // someone upgrades Ammonite.
     val cacheDir = dir/'cache/ammonite.Constants.version
     val compileCacheDir = cacheDir/'compile
+    val classFilesOrder = "classFilesOrder.json"
     val ivyCacheFile = cacheDir/"ivycache.json"
     val metadataFile = "metadata.json"
     val historyFile = dir/'history
@@ -86,6 +129,62 @@ object Storage{
         write(tagCacheDir/metadataFile, metadata)
         classFiles.foreach{ case (name, bytes) =>
           write(tagCacheDir/s"$name.class", bytes)
+        }
+      }
+    }
+
+    def classFilesListSave(pkg: String,
+                           wrapper: String,
+                           dataList: Seq[(String, String)],
+                           imports: Imports,
+                           tag: String,
+                           importTreesList: Seq[ImportTree]): Unit = {
+      val dir = pkg.replace("/", "$div") + "." + wrapper.replace("/", "$div")
+      val codeCacheDir = compileCacheDir/'scriptCaches/dir
+      if(!exists(codeCacheDir)) {
+        mkdir(codeCacheDir)
+        write(
+          codeCacheDir/classFilesOrder,
+          upickle.default.write((tag, dataList.reverse), indent = 4)
+        )
+        write(codeCacheDir/"imports.list", upickle.default.write(imports))
+        write(codeCacheDir/"importTrees.json", upickle.default.write(importTreesList, indent = 4))
+      }
+    }
+
+    def classFilesListLoad(pkg: String,
+                           wrapper: String,
+                           cacheTag: String): Option[CacheOutput] = {
+
+      val dir = pkg.replace("/", "$div") + "." + wrapper.replace("/", "$div")
+      val codeCacheDir = compileCacheDir/'scriptCaches/dir
+      if(!exists(codeCacheDir)) None
+      else {
+        val metadataJson = Try {
+          read(codeCacheDir/classFilesOrder)
+        }.toOption
+        val impFile = Try {
+          read(codeCacheDir/"imports.list")
+        }.toOption
+        val impTrees = Try{
+          read(codeCacheDir/"importTrees.json")
+        }.toOption
+        (metadataJson, impFile, impTrees) match{
+          case (Some(metadata), Some(imp), Some(importTrees)) =>
+            val (loadedTag, classFilesList) =
+              upickle.default.read[(String, Seq[(String, String)])](metadata)
+            if (cacheTag == loadedTag){
+              val res = {
+                for((path, tag) <- classFilesList) yield {
+                  compileCacheLoad(path, tag)
+                }
+              }.flatten
+              val imports = upickle.default.read[Imports](imp)
+              val impTreeList = upickle.default.read[Seq[ImportTree]](importTrees)
+              Some((classFilesList.unzip._1, res.unzip._1, imports, impTreeList))
+            }
+            else None
+          case _ => None
         }
       }
     }
--- a/repl/src/main/scala/ammonite/repl/Util.scala
+++ b/repl/src/main/scala/ammonite/repl/Util.scala
@@ -6,6 +6,7 @@ import ammonite.ops._
 import acyclic.file
 import fansi.Attrs
 import pprint.{PPrint, PPrinter}
+import ammonite.repl.Parsers.ImportTree
 
 import scala.collection.mutable
 import scala.reflect.NameTransformer
@@ -124,7 +125,8 @@ case class Catching(handler: PartialFunction[Throwable, Res.Failing]) {
 }
 
 case class Evaluated(wrapper: Seq[Name],
-                     imports: Imports)
+                     imports: Imports,
+                     tag: String)
 
 /**
   * Represents the importing of a single name in the Ammonite REPL, of the
@@ -307,8 +309,12 @@ object Util{
   }
 
   // Type aliases for common things
+
+  type CacheDetails = (String, String)
+  //                   Wrapper HashVal
   type IvyMap = Map[(String, String, String, String), Set[String]]
   type ClassFiles = Traversable[(String, Array[Byte])]
+  type CacheOutput = (Seq[String], Seq[ClassFiles], Imports, Seq[ImportTree])
   type CompileCache = (ClassFiles, Imports)
 
 
--- a/repl/src/main/scala/ammonite/repl/interp/Compiler.scala
+++ b/repl/src/main/scala/ammonite/repl/interp/Compiler.scala
@@ -47,10 +47,6 @@ trait Compiler{
   def parse(line: String): Either[String, Seq[Global#Tree]]
   var importsLen = 0
 
-  /**
-   * Writes files to dynamicClasspath. Needed for loading cached classes.
-   */ 
-  def addToClasspath(classFiles: ClassFiles): Unit
 }
 object Compiler{
   /**
@@ -138,6 +134,30 @@ object Compiler{
       val settings = settingsX
     }
     (reporter, vd, jcp)
+  }
+
+  def writeDeep(d: VirtualDirectory,
+                path: List[String],
+                suffix: String): OutputStream = path match {
+    case head :: Nil => d.fileNamed(path.head + suffix).output
+    case head :: rest =>
+      writeDeep(
+        d.subdirectoryNamed(head).asInstanceOf[VirtualDirectory],
+        rest, suffix
+      )
+  }
+
+  /**
+    * Writes files to dynamicClasspath. Needed for loading cached classes.
+    */
+  def addToClasspath(classFiles: Traversable[(String, Array[Byte])],
+                     dynamicClasspath: VirtualDirectory): Unit = {
+    val names = classFiles.map(_._1)
+    for((name, bytes) <- classFiles){
+      val output = writeDeep(dynamicClasspath, name.split('.').toList, ".class")
+      output.write(bytes)
+      output.close()
+    }
   }
 
   def apply(classpath: Seq[java.io.File],
@@ -256,16 +276,6 @@ object Compiler{
     }
 
 
-    def writeDeep(d: VirtualDirectory,
-                  path: List[String],
-                  suffix: String): OutputStream = path match {
-      case head :: Nil => d.fileNamed(path.head + suffix).output
-      case head :: rest =>
-        writeDeep(
-          d.subdirectoryNamed(head).asInstanceOf[VirtualDirectory],
-          rest, suffix
-        )
-    }
 
     /**
      * Compiles a blob of bytes and spits of a list of classfiles
@@ -317,14 +327,6 @@ object Compiler{
       }
     }
 
-    def addToClasspath(classFiles: Traversable[(String, Array[Byte])]): Unit = {
-      val names = classFiles.map(_._1)
-      for((name, bytes) <- classFiles){
-        val output = writeDeep(dynamicClasspath, name.split('.').toList, ".class")
-        output.write(bytes)
-        output.close()
-      }
-    }
 
     def parse(line: String): Either[String, Seq[Global#Tree]]= {
       val errors = mutable.Buffer.empty[String]
--- a/repl/src/main/scala/ammonite/repl/interp/Evaluator.scala
+++ b/repl/src/main/scala/ammonite/repl/interp/Evaluator.scala
@@ -5,10 +5,12 @@ import java.lang.reflect.InvocationTargetException
 import acyclic.file
 import ammonite.repl.frontend.{SessionChanged, Session, ReplExit}
 import ammonite.repl._
+import Parsers.ImportTree
 
 import Util.{CompileCache, ClassFiles}
 
 import scala.collection.immutable.ListMap
+import scala.reflect.io.VirtualDirectory
 import scala.collection.mutable
 import scala.util.Try
 
@@ -21,6 +23,7 @@ import scala.util.Try
  */
 trait Evaluator{
   def loadClass(wrapperName: String, classFiles: ClassFiles): Res[Class[_]]
+  def evalMain(cls: Class[_]): Any
   def getCurrentLine: String
   def update(newImports: Imports): Unit
 
@@ -33,9 +36,16 @@ trait Evaluator{
   def processScriptBlock(cls: Class[_],
                          newImports: Imports,
                          wrapperName: Name,
-                         pkgName: Seq[Name]): Res[Evaluated]
+                         pkgName: Seq[Name],
+                         tag: String): Res[Evaluated]
 
   def sess: Session
+
+  def evalCachedClassFiles(cachedData: Seq[ClassFiles],
+                           pkg: String,
+                           wrapper: String,
+                           dynamicClasspath: VirtualDirectory,
+                           classFilesList: Seq[String]): Res[Seq[_]]
 
 }
 
@@ -174,7 +184,9 @@ object Evaluator{
         Timer("eval.processLine evaluatorRunPrinter 1")
         evaluatorRunPrinter(iter.foreach(printer.out))
         Timer("eval.processLine evaluatorRunPrinter end")
-        evaluationResult(Seq(Name("$sess"), indexedWrapperName), newImports)
+
+        // "" Empty string as cache tag of repl code
+        evaluationResult(Seq(Name("$sess"), indexedWrapperName), newImports, "")
       }
     }
 
@@ -182,24 +194,45 @@ object Evaluator{
     def processScriptBlock(cls: Class[_],
                            newImports: Imports,
                            wrapperName: Name,
-                           pkgName: Seq[Name]) = for {
+                           pkgName: Seq[Name],
+                           tag: String) = for {
       _ <- Catching{userCodeExceptionHandler}
     } yield {
       Timer("cachedCompileBlock")
       evalMain(cls)
       Timer("evalMain")
-      val res = evaluationResult(pkgName :+ wrapperName, newImports)
+      val res = evaluationResult(pkgName :+ wrapperName, newImports, tag)
       Timer("evaluationResult")
       res
     }
 
+    def evalCachedClassFiles(cachedData: Seq[ClassFiles],
+                             pkg: String,
+                             wrapper: String,
+                             dynamicClasspath: VirtualDirectory,
+                             classFilesList: Seq[String]): Res[Seq[_]] = {
+
+      val res = Res.map(cachedData.zipWithIndex) {
+        case (clsFiles, index) =>
+          Compiler.addToClasspath(clsFiles, dynamicClasspath)
+          eval.loadClass(classFilesList(index), clsFiles)
+      }
+
+      try {
+        for {
+          r <- res
+        } yield r.map(eval.evalMain(_))
+      }
+      catch {userCodeExceptionHandler}
+    }
 
     def update(newImports: Imports) = {
       frames.head.addImports(newImports)
     }
 
     def evaluationResult(wrapperName: Seq[Name],
-                         imports: Imports) = {
+                         imports: Imports,
+                         tag: String) = {
       Evaluated(
         wrapperName,
         Imports(
@@ -218,7 +251,8 @@ object Evaluator{
 
             id.copy(prefix = rootedPrefix)
           }
-        )
+        ),
+        tag
       )
     }
   }
--- a/repl/src/main/scala/ammonite/repl/interp/Interpreter.scala
+++ b/repl/src/main/scala/ammonite/repl/interp/Interpreter.scala
@@ -1,19 +1,24 @@
 package ammonite.repl.interp
 import ammonite.repl.tools.{IvyThing, Resolver, Resolvers}
-import java.io.File
+import java.io.{File, OutputStream}
+import java.nio.file.NotDirectoryException
+import org.apache.ivy.plugins.resolver.RepositoryResolver
 import ammonite.repl.Res
 import scala.collection.mutable
 import scala.tools.nsc.Settings
-import acyclic.file
-import fastparse.all._
 import ammonite.ops._
+import fastparse.all._
 import ammonite.repl.Parsers.ImportTree
-import pprint.{Config, PPrint}
+
 import annotation.tailrec
 import ammonite.repl._
 import ammonite.repl.frontend._
+import Util._
+import pprint.{Config, PPrinter, PPrint}
+import ammonite.terminal.Filter
 
 import scala.reflect.io.VirtualDirectory
+import scala.util.Try
 
 /**
  * A convenient bundle of all the functionality necessary
@@ -51,9 +56,15 @@ class Interpreter(prompt0: Ref[String],
   val eval = Evaluator(mainThread.getContextClassLoader, 0  )
 
   val dynamicClasspath = new VirtualDirectory("(memory)", None)
-  var compiler: Compiler = _
+  var compiler: Compiler = null
   var pressy: Pressy = _
   def evalClassloader = eval.sess.frames.head.classloader
+
+  def reInit() = {
+    if(compiler != null)
+      init()
+  }
+
   def init() = {
     Timer("Interpreter init init 0")
     // Note we not only make a copy of `settings` to pass to the compiler,
@@ -81,8 +92,6 @@ class Interpreter(prompt0: Ref[String],
   }
 
 
-  val preprocess = Preprocessor(compiler.parse)
-
   Timer("Interpreter init Preprocess")
 
 
@@ -94,8 +103,7 @@ class Interpreter(prompt0: Ref[String],
     replApi
   )
   Timer("Interpreter init eval")
-  init()
-  Timer("Interpreter init init")
+
   val argString = replArgs.zipWithIndex.map{ case (b, idx) =>
     s"""
     val ${b.name} =
@@ -117,13 +125,13 @@ class Interpreter(prompt0: Ref[String],
   var predefImports = Imports()
   for( (sourceCode, wrapperName) <- predefs) {
     val pkgName = Seq(Name("ammonite"), Name("predef"))
-    processModule0(
+
+    processModule(
       ImportHook.Source.File(wd/"Main.scala"),
       sourceCode,
       wrapperName,
       pkgName,
-      predefImports,
-      autoImport = true
+      true
     ) match{
       case Res.Success(imports) =>
         predefImports = predefImports ++ imports
@@ -140,7 +148,7 @@ class Interpreter(prompt0: Ref[String],
 
   eval.sess.save()
   Timer("Interpreter init predef 0")
-  init()
+  reInit()
   Timer("Interpreter init predef 1")
 
   val importHooks = Ref(Map[Seq[String], ImportHook](
@@ -178,12 +186,12 @@ class Interpreter(prompt0: Ref[String],
           Res.Success(Imports())
       }
     } yield {
-      init()
+      reInit()
       hookResults
     }
   }
   def resolveImportHooks(source: ImportHook.Source,
-                         stmts: Seq[String]): Res[(Imports, Seq[String])] = {
+                         stmts: Seq[String]): Res[(Imports, Seq[String], Seq[ImportTree])] = {
     val hookedStmts = mutable.Buffer.empty[String]
     val importTrees = mutable.Buffer.empty[ImportTree]
     for(stmt <- stmts) {
@@ -208,33 +216,39 @@ class Interpreter(prompt0: Ref[String],
       hookImports <- Res.map(importTrees)(resolveSingleImportHook(source, _))
     } yield {
       val imports = Imports(hookImports.flatten.flatMap(_.value))
-      (imports, hookedStmts)
+      (imports, hookedStmts, importTrees)
     }
   }
 
-  def processLine(code: String, stmts: Seq[String], fileName: String): Res[Evaluated] = for{
-    _ <- Catching { case ex =>
-      Res.Exception(ex, "Something unexpected went wrong =(")
-    }
+  def processLine(code: String, stmts: Seq[String], fileName: String): Res[Evaluated] ={
+    val preprocess = Preprocessor(compiler.parse)
+    for{
+      _ <- Catching { case ex =>
+        Res.Exception(ex, "Something unexpected went wrong =(")
+      }
 
-    (hookImports, hookedStmts) <- resolveImportHooks(
-      ImportHook.Source.File(wd/"Main.scala"), stmts
-    )
-    processed <- preprocess.transform(
-      hookedStmts,
-      eval.getCurrentLine,
-      "",
-      Seq(Name("$sess")),
-      Name("cmd" + eval.getCurrentLine),
-      eval.sess.frames.head.imports ++ hookImports,
-      prints => s"ammonite.repl.frontend.ReplBridge.repl.Internal.combinePrints($prints)"
-    )
+      (hookImports, hookedStmts, _) <- resolveImportHooks(
+        ImportHook.Source.File(wd/"<console>"),
+        stmts
+      )
+      processed <- preprocess.transform(
+        hookedStmts,
+        eval.getCurrentLine,
+        "",
+        Seq(Name("$sess")),
+        Name("cmd" + eval.getCurrentLine),
+        predefImports ++ eval.sess.frames.head.imports ++ hookImports,
+        prints => s"ammonite.repl.frontend.ReplBridge.repl.Internal.combinePrints($prints)"
+      )
+      out <- evaluateLine(
+        processed, printer,
+        fileName, Name("cmd" + eval.getCurrentLine)
+      )
+    } yield out.copy(imports = out.imports ++ hookImports)
 
-    out <- evaluateLine(
-      processed, printer,
-      fileName, Name("cmd" + eval.getCurrentLine)
-    )
-  } yield out.copy(imports = out.imports ++ hookImports)
+  }
+
+
 
 
   def withContextClassloader[T](t: => T) = {
@@ -267,9 +281,8 @@ class Interpreter(prompt0: Ref[String],
   def evaluateLine(processed: Preprocessor.Output,
                    printer: Printer,
                    fileName: String,
-                   indexedWrapperName: Name): Res[Evaluated] = {
+                   indexedWrapperName: Name): Res[Evaluated] = for{
 
-    for{
       _ <- Catching{ case e: ThreadDeath => Evaluator.interrupted(e) }
       (classFiles, newImports) <- compileClass(
         processed,
@@ -287,14 +300,14 @@ class Interpreter(prompt0: Ref[String],
 
       }
     } yield res
-  }
+
 
   def processScriptBlock(processed: Preprocessor.Output,
                          printer: Printer,
                          wrapperName: Name,
                          fileName: String,
                          pkgName: Seq[Name]) = for {
-    (cls, newImports) <- cachedCompileBlock(
+    (cls, newImports, tag) <- cachedCompileBlock(
       processed,
       printer,
       wrapperName,
@@ -302,7 +315,7 @@ class Interpreter(prompt0: Ref[String],
       pkgName,
       "scala.Iterator[String]()"
     )
-    res <- eval.processScriptBlock(cls, newImports, wrapperName, pkgName)
+    res <- eval.processScriptBlock(cls, newImports, wrapperName, pkgName, tag)
   } yield res
 
 
@@ -311,7 +324,7 @@ class Interpreter(prompt0: Ref[String],
                          wrapperName: Name,
                          fileName: String,
                          pkgName: Seq[Name],
-                         printCode: String): Res[(Class[_], Imports)] = {
+                         printCode: String): Res[(Class[_], Imports, String)] = {
 
     Timer("cachedCompileBlock 1")
 
@@ -322,7 +335,7 @@ class Interpreter(prompt0: Ref[String],
     Timer("cachedCompileBlock 2")
     val compiled = storage.compileCacheLoad(fullyQualifiedName, tag) match {
       case Some((classFiles, newImports)) =>
-        compiler.addToClasspath(classFiles)
+        Compiler.addToClasspath(classFiles, dynamicClasspath)
         Res.Success((classFiles, newImports))
       case _ =>
         val noneCalc = for {
@@ -339,32 +352,71 @@ class Interpreter(prompt0: Ref[String],
       (classFiles, newImports) <- compiled
       _ = Timer("cachedCompileBlock 4")
       cls <- eval.loadClass(fullyQualifiedName, classFiles)
-    } yield (cls, newImports)
+    } yield (cls, newImports, tag)
   }
 
   def processModule(source: ImportHook.Source,
                     code: String,
                     wrapperName: Name,
                     pkgName: Seq[Name],
-                    autoImport: Boolean) = {
+                    autoImport: Boolean): Res[Imports] = {
 
-    processModule0(source, code, wrapperName, pkgName, predefImports, autoImport)
+    val cacheTag = Util.md5Hash(Iterator(code.getBytes)).map("%02x".format(_)).mkString
+    storage.classFilesListLoad(
+      pkgName.map(_.backticked).mkString("."),
+      wrapperName.backticked,
+      cacheTag
+    ) match {
+      case None =>
+        init()
+        val res = processModule0(source, code, wrapperName, pkgName, predefImports, autoImport)
+        res match{
+         case Res.Success(data) =>
+           reInit()
+           val (imports, cachedData, importTrees) = data
+           storage.classFilesListSave(
+             pkgName.map(_.backticked).mkString("."),
+             wrapperName.backticked,
+             cachedData,
+             imports,
+             cacheTag,
+             importTrees
+           )
+           res.map(_._1)
+         case r: Res.Failing => r
+       }
+      case Some((classFilesList, cachedData, imports, importsTrees)) =>
+        importsTrees.map(resolveSingleImportHook(source, _))
+        eval.evalCachedClassFiles(
+          cachedData,
+          pkgName.map(_.backticked).mkString("."),
+          wrapperName.backticked,
+          dynamicClasspath,
+          classFilesList
+        ) match {
+          case Res.Success(_) =>
+            eval.update(imports)
+            Res.Success(imports)
+          case r: Res.Failing => r
+        }
+    }
+
   }
 
   def preprocessScript(source: ImportHook.Source, code: String) = for{
     blocks <- Preprocessor.splitScript(Interpreter.skipSheBangLine(code))
     hooked <- Res.map(blocks){case (prelude, stmts) => resolveImportHooks(source, stmts) }
-    (hookImports, hookBlocks) = hooked.unzip
-  } yield (blocks.map(_._1).zip(hookBlocks), Imports(hookImports.flatMap(_.value)))
+    (hookImports, hookBlocks, importTrees) = hooked.unzip3
+  } yield (blocks.map(_._1).zip(hookBlocks), Imports(hookImports.flatMap(_.value)), importTrees)
 
   def processModule0(source: ImportHook.Source,
                      code: String,
                      wrapperName: Name,
                      pkgName: Seq[Name],
                      startingImports: Imports,
-                     autoImport: Boolean): Res[Imports] = for{
-    (processedBlocks, hookImports) <- preprocessScript(source, code)
-    res <- processCorrectScript(
+                     autoImport: Boolean): Res[Interpreter.ProcessedData] = for{
+    (processedBlocks, hookImports, importTrees) <- preprocessScript(source, code)
+    (imports, cacheData) <- processCorrectScript(
       processedBlocks,
       startingImports ++ hookImports,
       pkgName,
@@ -379,28 +431,35 @@ class Interpreter(prompt0: Ref[String],
         ),
       autoImport
     )
-  } yield res ++ hookImports
+  } yield (imports ++ hookImports, cacheData, importTrees.flatten)
 
-  def processExec(code: String): Res[Imports] = for {
-    (processedBlocks, hookImports) <- preprocessScript(
-      ImportHook.Source.File(wd/"Main.scala"), code
-    )
-    res <- processCorrectScript(
-      processedBlocks,
-      eval.sess.frames.head.imports ++ hookImports,
-      Seq(Name("$sess")),
-      Name("cmd" + eval.getCurrentLine),
-      { (processed, wrapperIndex, indexedWrapperName) =>
-        evaluateLine(
-          processed,
-          printer,
-          s"Main$wrapperIndex.scala",
-          indexedWrapperName
-        )
-      },
-      autoImport = true
-    )
-  } yield res ++ hookImports
+
+
+  def processExec(code: String): Res[Imports] = {
+    init()
+    for {
+      (processedBlocks, hookImports, _) <- preprocessScript(
+        ImportHook.Source.File(wd/"<console>"),
+        code
+      )
+      (imports, _) <- processCorrectScript(
+        processedBlocks,
+        eval.sess.frames.head.imports ++ hookImports,
+        Seq(Name("$sess")),
+        Name("cmd" + eval.getCurrentLine),
+        { (processed, wrapperIndex, indexedWrapperName) =>
+          evaluateLine(
+            processed,
+            printer,
+            s"Main$wrapperIndex.scala",
+            indexedWrapperName
+          )
+        },
+        autoImport = true
+      )
+    } yield imports ++ hookImports
+  }
+
 
 
   def processCorrectScript(blocks: Seq[(String, Seq[String])],
@@ -409,9 +468,9 @@ class Interpreter(prompt0: Ref[String],
                            wrapperName: Name,
                            evaluate: Interpreter.EvaluateCallback,
                            autoImport: Boolean
-                          )
-                          : Res[Imports] = {
+                          ): Res[Interpreter.CacheData] = {
 
+    val preprocess = Preprocessor(compiler.parse)
     Timer("processCorrectScript 1")
     // we store the old value, because we will reassign this in the loop
     val outerScriptImportCallback = scriptImportCallback
@@ -427,12 +486,13 @@ class Interpreter(prompt0: Ref[String],
     @tailrec def loop(blocks: Seq[(String, Seq[String])],
                       scriptImports: Imports,
                       lastImports: Imports,
-                      wrapperIndex: Int): Res[Imports] = {
+                      wrapperIndex: Int,
+                      compiledData: List[CacheDetails]): Res[Interpreter.CacheData] = {
       if (blocks.isEmpty) {
         // No more blocks
         // if we have imports to pass to the upper layer we do that
         if (autoImport) outerScriptImportCallback(lastImports)
-        Res.Success(lastImports)
+        Res.Success(lastImports, compiledData)
       } else {
         Timer("processScript loop 0")
         // imports from scripts loaded from this script block will end up in this buffer
@@ -462,14 +522,26 @@ class Interpreter(prompt0: Ref[String],
           case r: Res.Exception => r
           case Res.Success(ev) =>
             val last = ev.imports ++ nestedScriptImports
-            loop(blocks.tail, scriptImports ++ last, last, wrapperIndex + 1)
-          case Res.Skip => loop(blocks.tail, scriptImports, lastImports, wrapperIndex + 1)
+            loop(
+              blocks.tail,
+              scriptImports ++ last,
+              last,
+              wrapperIndex + 1,
+              (ev.wrapper.map(_.backticked).mkString("."), ev.tag) :: compiledData
+            )
+          case Res.Skip => loop(
+            blocks.tail,
+            scriptImports,
+            lastImports,
+            wrapperIndex + 1,
+            compiledData
+          )
         }
       }
     }
     // wrapperIndex starts off as 1, so that consecutive wrappers can be named
     // Wrapper, Wrapper2, Wrapper3, Wrapper4, ...
-    try loop(blocks, startingImports, Imports(), wrapperIndex = 1 )
+    try loop(blocks, startingImports, Imports(), wrapperIndex = 1, List())
     finally scriptImportCallback = outerScriptImportCallback
   }
 
@@ -509,7 +581,7 @@ class Interpreter(prompt0: Ref[String],
 
     def cp(jar: Path): Unit = {
       handleClasspath(new java.io.File(jar.toString))
-      init()
+      reInit()
     }
     def ivy(coordinates: (String, String, String), verbose: Boolean = true): Unit = {
       val resolved = loadIvy(coordinates, verbose)
@@ -521,8 +593,7 @@ class Interpreter(prompt0: Ref[String],
 
       resolved.foreach(handleClasspath)
 
-
-      init()
+      reInit()
     }
   }
 
@@ -557,7 +628,7 @@ class Interpreter(prompt0: Ref[String],
 
       def exec(file: Path): Unit = apply(read(file))
 
-      def module(file: Path): Unit = {
+      def module(file: Path) = {
         val (pkg, wrapper) = Util.pathToPackageWrapper(file, wd)
         processModule(
           ImportHook.Source.File(wd/"Main.scala"),
@@ -570,7 +641,7 @@ class Interpreter(prompt0: Ref[String],
           case Res.Exception(t, s) => throw t
           case x => //println(x)
         }
-        init()
+        reInit()
       }
 
       object plugin extends DefaultLoadJar {
@@ -641,12 +712,12 @@ class Interpreter(prompt0: Ref[String],
 
       def pop(num: Int = 1) = {
         val res = eval.sess.pop(num)
-        init()
+        reInit()
         res
       }
       def load(name: String = "") = {
         val res = eval.sess.load(name)
-        init()
+        reInit()
         res
       }
     }
@@ -669,7 +740,7 @@ object Interpreter{
       Util.md5Hash(imports.iterator.map(_.toString.getBytes)),
       classpathHash
     ))
-    "cache" + bytes.map("%02x".format(_)).mkString //add prefix to make sure it begins with a letter
+    bytes.map("%02x".format(_)).mkString
   }
 
   def skipSheBangLine(code: String)= {
@@ -679,8 +750,10 @@ object Interpreter{
       code
   }
 
-  type EvaluateCallback = (Preprocessor.Output, Int, Name) => Res[Evaluated]
 
+  type EvaluateCallback = (Preprocessor.Output, Int, Name) => Res[Evaluated]
+  type CacheData = (Imports, Seq[CacheDetails])
+  type ProcessedData = (Imports, Seq[CacheDetails], Seq[ImportTree])
   def indexWrapperName(wrapperName: Name, wrapperIndex: Int): Name = {
     Name(wrapperName.raw + (if (wrapperIndex == 1) "" else "_" + wrapperIndex))
   }
--- a/repl/src/test/resources/scriptLevelCaching/runTimeExceptions.scala
+++ b/repl/src/test/resources/scriptLevelCaching/runTimeExceptions.scala
@@ -1,0 +1,4 @@
+import ammonite.ops._
+println(5/read(cwd/'repl/'target/'test/'resources/'scriptLevelCaching/"num.value").trim.toInt)
+rm(cwd/'repl/'target/'test/'resources/'scriptLevelCaching/"num.value")
+write(cwd/'repl/'target/'test/'resources/'scriptLevelCaching/"num.value", "0")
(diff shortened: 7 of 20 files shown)


coderabhishek:
Program flow is like:
`
script is given  => check if script can be loaded from cache =>(if yes) instatiate an interpreter object with dummy compile function and load classFiles from cache using storage.cacheFilesListLoad and load and exec them

==>(Not Found in cache) run script as usual just at each stage while coming out of compileCache keep passing the classFiles wrapperName and hash value's list comprising of all the blocks' data
In main function store the data in cache using cacheFilesList

Even load.module follows an exact same process

Interpreter takes a flag withCompiler, if the flag is false no function will be called which requires a compiler therefore only when cache is being evaled make an interpreter with flag set false
`

repl/src/main/scala/ammonite/repl/Main.scala
```diff
@@ -70,21 +80,104 @@ case class Main(predef: String = "",
     res
   }
 
+
+  def runScriptLevelCache(path: Path, args: Seq[String] = Vector.empty,
```
coderabhishek: starting point for running scripts
Tries to load from cache else calls  `runscript`

repl/src/main/scala/ammonite/repl/Main.scala
```diff
@@ -70,21 +80,104 @@ case class Main(predef: String = "",
...
+                storage: Storage,
+                cacheTag: String,
+                cacheScript: Boolean = true): Res[Seq[ImportData]] = {
     val repl = instantiateRepl()
     val (pkg, wrapper) = Util.pathToPackageWrapper(path, wd)
-    repl.interp.processModule(read(path), wrapper, pkg) match{
+    repl.interp.processModule(read(path), wrapper, pkg) match {
       case x: Res.Failing => x
-      case Res.Success(imports) =>
+      case Res.Success(data) =>
```
coderabhishek: data is list of cached classFiles and corresponding imports to load and eval classFiles

repl/src/main/scala/ammonite/repl/Storage.scala
```diff
@@ -90,6 +92,41 @@ object Storage{
       }
     }
 
+    def classFilesListSave(pkg: String,
+                           wrapper: String,
+                           dataList: List[(String, String)],
+                           tag: String): Unit = {
+      val dir = pkg + "." + wrapper + "0"
```
coderabhishek: make a file with "0" appended to wrapperName having hashValue and wrapperName of each block

lihaoyi:
Great that you got tests passing! 

Some high level feedback:

- **Would it be possible to move all the logic into the `processModule` function or some helper called by it**? It seems you are duplicating a bunch of logic between `runScriptLevelCache` and `cachedModule`, as called through `loadModule` and `load.module`. Given that all these code paths end up going through `processModule0` anyway, this would help keep the logic in one place.

- **If we did that, we could probably get rid of `runScriptLevelCache` entirely**, since `runScript` will eventuall call `processModule0` which will take care of the caching etc.. After all, the "first" script you run is no different from any other script you `load.module`; they all go through `processModule0`

- **We could probably get rid of the `withCompiler` flag**; if all the script-caching logic is included as part of `processModule0`, check like https://github.com/lihaoyi/Ammonite/pull/403/files#diff-bcdb6a0282e047eba770bc309743e114R126 become un-necessary since that code calls `processModule0`, which should do the right thing by default

That's the high-level review; your code looks great. Leaving some other feedback in-line. Now that you've got this working and tests passing, let's iterate on this diff until it's great!

integration/src/test/scala/ammonite/integration/BasicTests.scala
```diff
@@ -44,7 +46,7 @@ object BasicTests extends TestSuite{
           |cd! 'repl/'src
           |@
           |println(wd relativeTo x)""".stripMargin
-      )
+        )
```
lihaoyi: Try not to re-format irrelevant things as part of this diff. Things like [this](https://github.com/lihaoyi/Ammonite/pull/403/files#diff-2acce48a0e2b7d464bb53c0acc47e29fR81), and [this](https://github.com/lihaoyi/Ammonite/pull/403/files#diff-bcdb6a0282e047eba770bc309743e114R239) (adding indentation to the whole for-comprehension) may or may not be the "right" formatting, but they have no place in an already-very-large diff like this one. Given that this diff is already large and hard to review for correctness, you should aim to avoid all this sort of minor/irrelevant changes so we can focus on the script-level-caching. If we care enough, we can put these changes into a separate diff later.

repl/src/main/scala/ammonite/repl/interp/Interpreter.scala
```diff
@@ -454,16 +493,65 @@ class Interpreter(prompt0: Ref[String],
...
 
+      // Try loading script from cache
+      def cachedModule(path: Path) = {
+        Timer.startTime = System.nanoTime()
+
+        val code: String = read(path)
+        val (pkg, wrapper) = Util.pathToPackageWrapper(path, wd)
+        val cacheTag = "cache" + Util.md5Hash(Iterator(code.getBytes))
+          .map("%02x".format(_)).mkString
+        if(!code.contains("load(")) {
```
lihaoyi: Why do we need to check if the code contains `load(`? 

I don't think this is necessary; we should cache a module regardless of what it does. Even if we did want to do this, it would immediately break once someone called the function via `load (...)` or `load.apply (...)`

repl/src/main/scala/ammonite/repl/interp/Preprocessor.scala
```diff
@@ -81,7 +81,8 @@ object Preprocessor{
                   indexedWrapperName: String,
                   imports: Imports,
                   printerTemplate: String => String) = for{
-      Preprocessor.Expanded(code, printer) <- expandStatements(stmts, resultIndex)
+            Preprocessor.Expanded(code, printer) <- expandStatements(stmts, resultIndex)
```
lihaoyi: Don't reformat things unnecessarily

repl/src/main/scala/ammonite/repl/interp/Preprocessor.scala
```diff
@@ -186,14 +187,13 @@ object Preprocessor{
       }}
       unwrapped match{
         case Nil => Res.Skip
-        case postSplit =>
-          complete(stmts.mkString(""), wrapperIndex, postSplit)
-
+        case postSplit => complete(stmts.mkString(""), wrapperIndex, postSplit)
```
lihaoyi: Don't reformat things unnecessarily

repl/src/test/scala/ammonite/repl/TestMain.scala
```diff
@@ -3,6 +3,6 @@ package ammonite.repl
 
 object TestMain{
   def main(args: Array[String]): Unit = {
-    Main.main(args ++ Array("--home", "target/tempAmmoniteHome"))
+    Main.main(args)
```
lihaoyi: This shouldn't need to change, should it?

integration/src/test/scala/ammonite/integration/BasicTests.scala
```diff
@@ -107,5 +108,20 @@ object BasicTests extends TestSuite{
       res.out.lines.foreach { println }
       println("\n-------------------------")
     }
+
+    'scriptCache{
+      val scripts = List('scriptLevelCaching/"testLoadModule.scala",
+        'scriptLevelCaching/"scriptTwo.scala",
+        'scriptLevelCaching/"scriptOne.scala",
+        'scriptLevelCaching/"QuickSort.scala")
```
lihaoyi: Format this (and [this](https://github.com/lihaoyi/Ammonite/pull/403/files#diff-d71dddf61413a0340636e6a9eaadee32R103) and [this](https://github.com/lihaoyi/Ammonite/pull/403/files#diff-bcdb6a0282e047eba770bc309743e114R400)) consistently with other multi-line functions:

```scala
FUNCTION(
  arg0,
  arg1,
  arg2,
)
```

repl/src/main/scala/ammonite/repl/Main.scala
```diff
@@ -70,21 +80,104 @@ case class Main(predef: String = "",
...
+      case _ =>
+        if (args == Seq() || !code.contains("load(")) {
+          storage.asInstanceOf[Storage.Folder].classFilesListLoad(pkg, wrapper, cacheTag) match {
+            case Seq() =>
+              runScript(path, args, kwargs, storage, cacheTag)
+            case cachedData =>
+              val outBuffer = mutable.Buffer.empty[String]
+              val warningBuffer = mutable.Buffer.empty[String]
+              val errorBuffer = mutable.Buffer.empty[String]
+              val infoBuffer = mutable.Buffer.empty[String]
```
lihaoyi: Not sure why we're building up these buffers and throwing them away, but it doesn't matter anyway if we're getting rid of `runScriptLevelCache` as described in my high-level feedback

repl/src/main/scala/ammonite/repl/interp/Interpreter.scala
```diff
@@ -454,16 +493,65 @@ class Interpreter(prompt0: Ref[String],
...
+                }
+                blockNumber += 1
+              }
+              }
+          }
+        } else loadModule(path, storage.asInstanceOf[Storage.Folder], cacheTag)
+      }
+
+      def module(file: Path): Unit = {
+        if (!withCompiler)
```
lihaoyi: We shouldn't need this check if `loadModule` (or `processModule0`) helps us check whether a loaded script has been fully-cached every single time

repl/src/main/scala/ammonite/repl/interp/Interpreter.scala
```diff
@@ -311,14 +349,66 @@ class Interpreter(prompt0: Ref[String],
...
   }
 
+
+  def processModule(source: ImportHook.Source,
+                    code: String,
+                    wrapperName: Name,
+                    pkgName: Seq[Name]
+                   ): Res[Imports] = if(scriptCaching) {
+    val cacheTag = "cache" + Util.md5Hash(Iterator(code.getBytes)).map("%02x".format(_)).mkString
+    storage.asInstanceOf[Storage.Folder].classFilesListLoad(pkgName.map(_.backticked).mkString("."), wrapperName.backticked, cacheTag) match {
```
lihaoyi: We should need to `asInstanceOf` to check if something is a `Storage.Folder`; instead, `Storage` should has `classfilesListLoad` and `classfilesListLoad` as part of it's interface, and `Storage.InMemory` should just keep things in an in-memory `Map` or something when saved and read from that `Map` on load

repl/src/main/scala/ammonite/repl/interp/Interpreter.scala
```diff
@@ -33,10 +40,14 @@ class Interpreter(prompt0: Ref[String],
                   replArgs: Seq[Bind[_]])
   extends ImportHook.InterpreterInterface{ interp =>
 
-
   val hardcodedPredef =
     "import ammonite.repl.frontend.ReplBridge.repl.{pprintConfig, derefPPrint}"
 
+  // Should script as a whole be cached or not
+  val scriptCaching = storage match {
```
lihaoyi: This should not be necessary, as described in the comment below. `Storage.InMemory` should be able to cache things just fine, in memory.

repl/src/main/scala/ammonite/repl/interp/Interpreter.scala
```diff
@@ -103,37 +114,55 @@ class Interpreter(prompt0: Ref[String],
...
   }.mkString("\n")
 
-  val predefs = Seq(
-    (hardcodedPredef, Name("HardcodedPredef")),
-    (predef, Name("Predef")),
-    (storage.loadPredef, Name("LoadedPredef")),
-    (argString, Name("ArgsPredef"))
-  )
+  var predefImports = Imports(Nil)
+  initPredef()
```
lihaoyi: What is the purpose of moving this into a function?

repl/src/main/scala/ammonite/repl/Main.scala
```diff
@@ -71,21 +81,52 @@ case class Main(predef: String = "",
...
+                kwargs: Map[String, String],
+                storage: Storage): Res[Seq[ImportData]] = {
+    val outBuffer = mutable.Buffer.empty[String]
+    val warningBuffer = mutable.Buffer.empty[String]
+    val errorBuffer = mutable.Buffer.empty[String]
+    val infoBuffer = mutable.Buffer.empty[String]
+    val prompt = Ref("@ ")
+    val colors = Ref[Colors](Colors.Default)
+    val frontEnd = Ref[FrontEnd](AmmoniteFrontEnd(Filter.empty))
+    val printer = Printer(
```
lihaoyi: This is probably incorrect; we don't print stuff much when running scripts, but anything we do print should be printed "normally" e.g. via https://github.com/lihaoyi/Ammonite/blob/master/repl/src/main/scala/ammonite/repl/Repl.scala#L31-L36, and not aggregated into a bunch of random buffers to be thrown away

repl/src/main/scala/ammonite/repl/Main.scala
```diff
@@ -95,23 +136,28 @@ case class Main(predef: String = "",
...
-              case Res.Exception(e: ArgParseException, s) =>
-                e.setStackTrace(Array())
-                e.cause.setStackTrace(e.cause.getStackTrace.takeWhile( frame =>
-                  frame.getClassName != "ammonite.repl.ScriptInit$" ||
-                  frame.getMethodName != "parseScriptArg"
-                ))
-                Res.Exception(e, s)
-              case x => x
-            }
+            case Res.Success(_) => Res.
```
lihaoyi: Don't reformat things unnecessarily; it makes code difficult to review

repl/src/main/scala/ammonite/repl/interp/Interpreter.scala
```diff
@@ -311,14 +349,66 @@ class Interpreter(prompt0: Ref[String],
...
+    } yield (cls, newImports, tag)
+
   }
 
-  def processModule(source: ImportHook.Source, code: String, wrapperName: Name, pkgName: Seq[Name]) = {
 
-    processModule0(source, code, wrapperName, pkgName, predefImports)
+  def evalCachedClassFiles(cachedData: Seq[CompileCache], pkg: String, wrapper: String): Unit = {
+
+    def evalMain(cls: Class[_]) =
```
lihaoyi: This is already a method in the `Evaluator` class; perhaps we can move it out into the `Evaluator` companion object so this method can all it? Rather than copy-pasting

repl/src/main/scala/ammonite/repl/interp/Interpreter.scala
```diff
@@ -581,7 +687,7 @@ class Interpreter(prompt0: Ref[String],
       Interpreter.this.compiler.search(target)
     }
     def compiler = Interpreter.this.compiler.compiler
-    def newCompiler() = init()
+    def newCompiler() = null
```
lihaoyi: Not sure what you are doing here, but returning `null` is probably not the right thing to do

lihaoyi:
The unit tests aren't testing the right thing. Our goal isn't to get the `compilationCount` to zero, but instead it is to ensure that the *compiler is never initialized*.

Realistically, what you should do is:

- Introduce a new `compilerInitialized` boolean on the `Interpreter` that starts as `false` and gets set to `true` when/every-time the compiler is initialized
- Move the tests from integration tests into "normal" unit tests, in the `repl/` project
- Extract the body of `runScript` into `runScriptInternal`, with `runScriptInternal` `private[ammonite]` since it's only to be used for tests and not part of the public API
- Make `runScriptInternal` return a `Res[(Seq[ImportData], Boolean)]` with the boolean coming from the `compilerInitialized
- Make `runScript` call `runScriptInternal`, and discard the boolean.
- Make your unit tests instantiate the REPL and call scripts through the normal `Main()` call, except calling `runScriptInternal` rather than `runScript`, and validating that the second time (?) the same script is run, the `compilerInitialized: Boolean` that gets returned is `false`

repl/src/main/scala/ammonite/repl/interp/Interpreter.scala
```diff
@@ -311,14 +349,66 @@ class Interpreter(prompt0: Ref[String],
...
+    def getBlockNumber = blockNumber match {
+      case 1 => ""
+      case _ => "_" + blockNumber.toString
+    }
+
+    // Iterate through cached blocks and execute classFiles
+    // blockNumber keeps track of blockIndex
+    cachedData.foreach { d =>
+      for {
+        cls <- eval.loadClass(pkg + "." + wrapper + getBlockNumber, d._1)
```
lihaoyi: What happens if `loadClass` or `evalMain` fail? You should probably use `Res.map` on this, and make `evalCachedClassFiles` return a `Res[_]` to represent the possibility of failure

repl/src/main/scala/ammonite/repl/interp/Interpreter.scala
```diff
@@ -414,23 +509,31 @@ class Interpreter(prompt0: Ref[String],
...
+            wrapperIndex + 1,
+            compiledData
+          )
         }
       }
     }
     // wrapperIndex starts off as 1, so that consecutive wrappers can be named
     // Wrapper, Wrapper2, Wrapper3, Wrapper4, ...
-    try loop(blocks, startingImports, Imports(), wrapperIndex = 1 )
+    try loop(blocks, startingImports, Imports(Nil), wrapperIndex = 1, List[(String, String)]())
```
lihaoyi: This can be simply `List()`; the type is inferred

repl/src/main/scala/ammonite/repl/interp/Interpreter.scala
```diff
@@ -311,14 +349,66 @@ class Interpreter(prompt0: Ref[String],
...
+         case Res.Success(data) =>
+           val Imports(imports) = data._1
+           val files = for (d <- data._2) yield (d._1, d._2)
+           if (imports.count(_.toName == "main") == 0)
+             storage.asInstanceOf[Storage.Folder].classFilesListSave(pkgName.map(_.backticked).mkString("."), wrapperName.backticked, files, cacheTag)
+           res.map(_._1)
+         case r => res.asInstanceOf[Res[Imports]]
+       }
+      case cachedData =>
+        evalCachedClassFiles(cachedData, pkgName.map(_.backticked).mkString("."), wrapperName.backticked)
```
lihaoyi: If `evalCachedClassFiles` returns a `Res[_]`, we should propagate that `Res` to be the result of this function

repl/src/main/scala/ammonite/repl/interp/Interpreter.scala
```diff
@@ -311,14 +349,66 @@ class Interpreter(prompt0: Ref[String],
...
+        init()
+        val res = processModule0(source, code, wrapperName, pkgName, predefImports)
+       res match{
+         case Res.Success(data) =>
+           val Imports(imports) = data._1
+           val files = for (d <- data._2) yield (d._1, d._2)
+           if (imports.count(_.toName == "main") == 0)
+             storage.asInstanceOf[Storage.Folder].classFilesListSave(pkgName.map(_.backticked).mkString("."), wrapperName.backticked, files, cacheTag)
+           res.map(_._1)
+         case r => res.asInstanceOf[Res[Imports]]
```
lihaoyi: This should be just `case r => r`; shouldn't need `res` nor the `asInstanceOf`

repl/src/main/scala/ammonite/repl/Main.scala
```diff
@@ -71,21 +81,52 @@ case class Main(predef: String = "",
...
+    )
     val (pkg, wrapper) = Util.pathToPackageWrapper(path, wd)
-    repl.interp.processModule(ImportHook.Source.File(path), read(path), wrapper, pkg) match{
+    interp.processModule(ImportHook.Source.File(path), read(path), wrapper, pkg) match {
       case x: Res.Failing => x
-      case Res.Success(imports) =>
-        repl.interp.init()
-        imports.value.find(_.toName == "main") match {
+      case Res.Success(processedData) =>
+        val Imports(imports) = processedData
```
lihaoyi: Rather than pattern match, extracting `imports`, and then wrapping it back in a `Res.Success`, you should just return `processedData` directly below

repl/src/main/scala/ammonite/repl/interp/Interpreter.scala
```diff
@@ -51,18 +60,24 @@ class Interpreter(prompt0: Ref[String],
   val eval = Evaluator(mainThread.getContextClassLoader, 0  )
 
   val dynamicClasspath = new VirtualDirectory("(memory)", None)
-  var compiler: Compiler = _
+  var compiler: Compiler = null
   var pressy: Pressy = _
   def evalClassloader = eval.sess.frames.head.classloader
+
+  def updateClassPath = {
```
lihaoyi: Why does this have to be a global `var`? What's wrong with using `eval.sess.frames.head.classpath` directly?

integration/src/test/scala/ammonite/integration/BasicTests.scala
lihaoyi: @coderabhishek bump

repl/src/main/scala/ammonite/repl/Main.scala
lihaoyi: @coderabhishek bump, this is all dead code, can you clean it up?

repl/src/main/scala/ammonite/repl/Main.scala
lihaoyi: @coderabhishek bump

repl/src/main/scala/ammonite/repl/interp/Interpreter.scala
```diff
@@ -258,35 +310,36 @@ class Interpreter(prompt0: Ref[String],
...
-      processed,
-      printer,
-      wrapperName,
-      fileName,
-      pkgName,
-      "scala.Iterator[String]()"
-    )
-    res <- eval.processScriptBlock(cls, newImports, wrapperName, pkgName)
-  } yield res
+      (cls, newImports, tag) <- cachedCompileBlock(
```
lihaoyi: @coderabhishek avoid reformatting things unnecessarily

repl/src/main/scala/ammonite/repl/interp/Interpreter.scala
```diff
@@ -94,46 +106,76 @@ class Interpreter(prompt0: Ref[String],
...
-        predefImports = predefImports ++ imports
-      case Res.Failure(ex, msg) =>
-        ex match{
-          case Some(e) => throw new RuntimeException("Error during Predef: " + msg, e)
-          case None => throw new RuntimeException("Error during Predef: " + msg)
-        }
 
-      case Res.Exception(ex, msg) =>
-        throw new RuntimeException("Error during Predef: " + msg, ex)
+  def writeDeep(d: VirtualDirectory,
```
lihaoyi: Don't just copy and paste things! If you want to re-use the code that is currently in the `Compiler` class, you should move it to the companion object or to a separate file.

repl/src/main/scala/ammonite/repl/interp/Compiler.scala
```diff
@@ -296,8 +295,8 @@ object Compiler{
 
       run.compileFiles(List(singleFile))
 
-      val outputFiles = enumerateVdFiles(vd).toVector
 
+      val outputFiles = enumerateVdFiles(vd).toVector
```
lihaoyi: Formatting

repl/src/main/scala/ammonite/repl/interp/Interpreter.scala
```diff
@@ -198,15 +247,19 @@ class Interpreter(prompt0: Ref[String],
       "",
       Seq(Name("$sess")),
       Name("cmd" + eval.getCurrentLine),
-      eval.sess.frames.head.imports ++ hookImports,
+      predefImports ++ eval.sess.frames.head.imports ++ hookImports,
       prints => s"ammonite.repl.frontend.ReplBridge.repl.Internal.combinePrints($prints)"
     )
-
+//    _ <- Res.Success(init())
```
lihaoyi: Cleanup

repl/src/main/scala/ammonite/repl/interp/Interpreter.scala
```diff
@@ -388,12 +517,15 @@ class Interpreter(prompt0: Ref[String],
     @tailrec def loop(blocks: Seq[(String, Seq[String])],
                       scriptImports: Imports,
                       lastImports: Imports,
-                      wrapperIndex: Int): Res[Imports] = {
+                      wrapperIndex: Int,
+                      compiledData: List[CacheDetails]
+                     ): Res[(Imports, List[CacheDetails])] = {
       if (blocks.isEmpty) {
         // No more blocks
+        // No more blocks
```
lihaoyi: Cleanup

repl/src/main/scala/ammonite/repl/interp/Interpreter.scala
```diff
@@ -414,23 +546,31 @@ class Interpreter(prompt0: Ref[String],
             scriptImports,
             _ => "scala.Iterator[String]()"
           )
-
```
lihaoyi: Formatting

repl/src/main/scala/ammonite/repl/interp/Preprocessor.scala
```diff
@@ -194,14 +195,13 @@ object Preprocessor{
...
       }
     }
 
     def complete(code: String, resultIndex: String, postSplit: Seq[String]) = {
       val reParsed = postSplit.map(p => (parse(p), p))
+
```
lihaoyi: Formatting

repl/src/main/scala/ammonite/repl/interp/Preprocessor.scala
```diff
@@ -301,7 +301,8 @@ ${importBlock(imports)}
 
 object ${indexedWrapperName.backticked}{\n"""
 
-    val bottomWrapper = s"""\ndef $$main() = { $printCode }
+
+    val bottomWrapper = s"""\ndef $$main(): Iterator[String] = { $printCode }
```
lihaoyi: Formatting

repl/src/test/scala/ammonite/repl/session/EulerTests.scala
```diff
@@ -660,7 +660,7 @@ object EulerTests extends TestSuite{
         r: Int = 45228
       """)
     }
-    // Seems to crash the compiler
+//     Seems to crash the compiler
```
lihaoyi: Formatting

repl/src/test/scala/ammonite/repl/TestRepl.scala
```diff
@@ -73,6 +73,8 @@ class TestRepl {
         assert(Parsers.split(incomplete.mkString("\n")) == None)
       }
 
+//      println("+++++++++++++++++++++++++++++++++++=" + commandText + "++++++++++")
+
```
lihaoyi: Cleanup

shell/src/main/resources/ammonite/shell/example-predef-bare.scala
```diff
@@ -2,10 +2,12 @@ import ammonite.ops._
 val scalaVersion = scala.util.Properties.versionNumberString.split("\\.").dropRight(1).mkString(".")
 val ammVersion = ammonite.Constants.version
 load.cp(cwd/'shell/'target/s"scala-$scalaVersion"/s"ammonite-shell_$scalaVersion-$ammVersion.jar")
+
```
lihaoyi: Are these necessary?

repl/src/test/resources/scriptLevelCaching/predef.scala
```diff
@@ -0,0 +1,42 @@
+import ammonite.repl.frontend.ReplBridge.repl
```
lihaoyi: This file should not be necessary; if you need this code-string, you can read it off of `ammonite.repl.Main.defaultPredefString`. We shouldn't copy-paste things

repl/src/main/scala/ammonite/repl/interp/Interpreter.scala
```diff
@@ -94,46 +106,76 @@ class Interpreter(prompt0: Ref[String],
     replApi
   )
   Timer("Interpreter init eval")
-  init()
-  Timer("Interpreter init init")
+
+//  if(withCompiler) init()
+//  Timer("Interpreter init init")
+
```
lihaoyi: Cleanup

repl/src/main/scala/ammonite/repl/interp/Interpreter.scala
```diff
@@ -311,14 +364,82 @@ class Interpreter(prompt0: Ref[String],
...
+          wrapperName.backticked) match {
+          case Res.Success(_) =>
+            Res.Success(imports)
+          case r => r.asInstanceOf[Res[Imports]]
+
+        }
+    }
   }
 
+
```
lihaoyi: Formatting

repl/src/main/scala/ammonite/repl/Storage.scala
```diff
@@ -22,6 +23,14 @@ trait Storage{
...
   def compileCacheSave(path: String, tag: String, data: CompileCache): Unit
   def compileCacheLoad(path: String, tag: String): Option[CompileCache]
+  def classFilesListSave(pkg: String,
+                         wrapper: String,
+                         dataList: List[(String, String)],
+                         imports: Imports,
+                         tag: String): Unit
+  def classFilesListLoad(pkg: String,
+                         wrapper: String,
+                         cacheTag: String): (Seq[CompileCache], Imports)
```
lihaoyi: We should not overload empty sequences to mean the cached value was not found; this should really return `Option[(Seq[CompileCache], Imports)]`

repl/src/main/scala/ammonite/repl/interp/Interpreter.scala
```diff
@@ -311,14 +364,82 @@ class Interpreter(prompt0: Ref[String],
...
+  }
+
+
+  def processModule(source: ImportHook.Source,
+                    code: String,
+                    wrapperName: Name,
+                    pkgName: Seq[Name]
+                   ): Res[Imports] = {
+    val cacheTag = "cache" + Util.md5Hash(Iterator(code.getBytes)).map("%02x".format(_)).mkString
```
lihaoyi: Shouldn't need the `cache` prefix here

repl/src/main/scala/ammonite/repl/interp/Interpreter.scala
```diff
@@ -388,12 +517,15 @@ class Interpreter(prompt0: Ref[String],
     @tailrec def loop(blocks: Seq[(String, Seq[String])],
                       scriptImports: Imports,
                       lastImports: Imports,
-                      wrapperIndex: Int): Res[Imports] = {
+                      wrapperIndex: Int,
+                      compiledData: List[CacheDetails]
+                     ): Res[(Imports, List[CacheDetails])] = {
```
lihaoyi: `(Imports, List[CacheDetails])` should be its own case class, since we appear to be using it everywhere. Put it in `Util.scala` with all the other such case classes

lihaoyi:
Trying this out locally, it appears that the cache isn't working: even if I run the same hello-world script multiple times, it thinks there's a cache-miss which results in the compiler being initialized, slowing things down:

```scala
haoyi-mbp:Ammonite haoyi$ cat Hello.scala
val hello = "Hello!"
println(hello)
```
```diff
haoyi-mbp:Ammonite haoyi$ git diff
diff --git a/repl/src/main/scala/ammonite/repl/Main.scala b/repl/src/main/scala/ammonite/repl/Main.scala
index ba696d9..2fe6ed1 100644
--- a/repl/src/main/scala/ammonite/repl/Main.scala
+++ b/repl/src/main/scala/ammonite/repl/Main.scala
@@ -111,6 +111,7 @@ case class Main(predef: String = "",
       case x: Res.Failing =>
         x
       case Res.Success(processedData) =>
+        pprint.log(repl.interp.compiler)
         val Imports(imports) = processedData

         repl.interp.updateClassPath
diff --git a/repl/src/main/scala/ammonite/repl/interp/Interpreter.scala b/repl/src/main/scala/ammonite/repl/interp/Interpreter.scala
index 5dc223f..8cead90 100644
--- a/repl/src/main/scala/ammonite/repl/interp/Interpreter.scala
+++ b/repl/src/main/scala/ammonite/repl/interp/Interpreter.scala
@@ -408,6 +408,7 @@ class Interpreter(prompt0: Ref[String],
       wrapperName.backticked,
       cacheTag) match {
       case (Seq(), Imports(Nil)) =>
+        println("MISS")
         init()
         val res = processModule0(source, code, wrapperName, pkgName, predefImports)
         res match{
@@ -426,6 +427,7 @@ class Interpreter(prompt0: Ref[String],
            r.asInstanceOf[Res[Imports]]
        }
       case (cachedData, imports) =>
+        println("HIT")
         evalCachedClassFiles(cachedData,
           pkgName.map(_.backticked).mkString("."),
           wrapperName.backticked) match {

```
```bash
haoyi-mbp:Ammonite haoyi$ time repl/target/scala-2.11/ammonite-repl-0.6.3-SNAPSHOT-2.11.8 Hello.scala
MISS
MISS
MISS
MISS
Hello!
runScript:114 repl.interp.compiler	ammonite.repl.interp.Compiler$$anon$4@3be50b2c

real	0m3.892s
user	0m14.301s
sys	0m0.553s
haoyi-mbp:Ammonite haoyi$ time repl/target/scala-2.11/ammonite-repl-0.6.3-SNAPSHOT-2.11.8 Hello.scala
HIT
HIT
MISS
HIT
Hello!
runScript:114 repl.interp.compiler	ammonite.repl.interp.Compiler$$anon$4@2f19f2aa

real	0m1.750s
user	0m3.768s
sys	0m0.264s
```

This only occur *with* my current predef; with *no* predef, it is fast and there are no cache misses. The problematic predef is:

```scala
/*load.ivy("com.lihaoyi" %% "ammonite-shell" % ammonite.Constants.version)
@
val shellSession = ammonite.shell.ShellSession()
import shellSession._
import ammonite.shell.PPrints._
import ammonite.ops._
import ammonite.shell._
ammonite.shell.Configure(repl, wd)*/
```

The whole thing is commented out; I suspect this is your code which checks to see if a cache entry is found:

```scala
case (Seq(), Imports(Nil)) =>
```

Which sees that there are no imports and no blocks, and then assumes it needs to initialize the compiler to start doing things

lihaoyi:
The following ~/.ammonite/predef.scala` seems to fail as well with your diff, even before caching takes place

```scala
haoyi-mbp:Ammonite haoyi$ cat ~/.ammonite/predef.scala
load.ivy("com.lihaoyi" %% "scalatags" % "0.5.4")
@
import scalatags.Text.all._
println(div(123).render)
haoyi-mbp:Ammonite haoyi$ time repl/target/scala-2.11/ammonite-repl-0.6.3-SNAPSHOT-2.11.8 Hello.scala
HIT
HIT
MISS
:: loading settings :: url = jar:file:/Users/haoyi/Dropbox%20(Personal)/Workspace/Ammonite/repl/target/scala-2.11/ammonite-repl-0.6.3-SNAPSHOT-2.11.8!/org/apache/ivy/core/settings/ivysettings.xml
:: resolving dependencies :: com.lihaoyi#scalatags_2.11-caller;working
	confs: [default]
	found com.lihaoyi#scalatags_2.11;0.5.4 in central
	found org.scala-lang#scala-library;2.11.4 in m2
	found com.lihaoyi#acyclic_2.11;0.1.2 in central
	found org.scala-lang#scala-compiler;2.11.0 in central
	found org.scala-lang.modules#scala-xml_2.11;1.0.1 in chain-resolver
	found org.scala-lang.modules#scala-parser-combinators_2.11;1.0.1 in chain-resolver
	found org.scala-lang#scala-reflect;2.11.4 in central

:: problems summary ::
:::: WARNINGS
	configuration not found 'compile' in org.scala-lang#scala-library;2.11.4: ignoring

	it was required from org.scala-lang#scala-reflect;2.11.4


:: USE VERBOSE OR DEBUG MESSAGE LEVEL FOR MORE DETAILS
LoadedPredef.scala:3: not found: value scalatags
import scalatags.Text.all._
       ^
LoadedPredef.scala:4: not found: value div
;val res_1 = println(div(123).render)
                     ^
Exception in thread "main" java.lang.RuntimeException: Error during Predef: Compilation Failed
	at ammonite.repl.interp.Interpreter$$anonfun$8.apply(Interpreter.scala:144)
	at ammonite.repl.interp.Interpreter$$anonfun$8.apply(Interpreter.scala:133)
	at scala.collection.TraversableLike$WithFilter$$anonfun$foreach$1.apply(TraversableLike.scala:733)
	at scala.collection.immutable.List.foreach(List.scala:381)
	at scala.collection.TraversableLike$WithFilter.foreach(TraversableLike.scala:732)
	at ammonite.repl.interp.Interpreter.<init>(Interpreter.scala:133)
	at ammonite.repl.Repl.<init>(Repl.scala:40)
	at ammonite.repl.Main.instantiateRepl(Main.scala:68)
	at ammonite.repl.Main.runScript(Main.scala:109)
	at ammonite.repl.Main$$anonfun$main$2.apply(Main.scala:297)
	at ammonite.repl.Main$$anonfun$main$2.apply(Main.scala:280)
	at scala.Option.foreach(Option.scala:257)
	at ammonite.repl.Main$.main(Main.scala:280)
	at ammonite.repl.Main.main(Main.scala)

real	0m2.464s
user	0m6.465s
sys	0m0.364s
```

repl/src/main/scala/ammonite/repl/Main.scala
```diff
@@ -102,16 +131,21 @@ case class Main(predef: String = "",
...
+            case Res.Exception(e:
+              ArgParseException, s) =>
+              e.setStackTrace(Array())
+              e.cause.setStackTrace(e.
+                cause.getStackTrace.takeWhile(frame =>
+                frame.getClassName != "ammonite.repl.ScriptInit$" ||
+                  frame.getMethodName !=
+                    "parseScriptArg"
+              ))
+              Res.Exception(e
```
lihaoyi: Formatting

integration/src/test/scala/ammonite/integration/BasicTests.scala
```diff
@@ -77,9 +77,7 @@ object BasicTests extends TestSuite{
       // tests a bit more tightly; currently the error just goes to stdout
       // and there's no way to inspect/validate it =/
       'tooFew{
-        val errorMsg = intercept[ShelloutException]{
-          exec('basic/"Args.scala", "3")
-        }.result.err.string
+        val errorMsg = exec('basic/"Args.scala", "3").err.string
```
lihaoyi: This is fine for debugging, but unless we explicitly want to change the behavior we should not leave this in the final codebase

repl/src/main/scala/ammonite/repl/Main.scala
```diff
@@ -77,15 +80,17 @@ case class Main(predef: String = "",
     */
   def runScript(path: Path,
                 args: Seq[String],
-                kwargs: Map[String, String]): Res[Imports] = {
+                kwargs: Map[String, String],
+                storage: Storage): Res[Seq[ImportData]] = {
```
lihaoyi: You should be able to use the `storageBackend` from the case class here, and not need to pass it in yourself again

repl/src/main/scala/ammonite/repl/Main.scala
```diff
@@ -105,13 +110,15 @@ case class Main(predef: String = "",
               case Res.Success(_) => Res.Success(imports)
               case Res.Exception(e: ArgParseException, s) =>
                 e.setStackTrace(Array())
-                e.cause.setStackTrace(e.cause.getStackTrace.takeWhile( frame =>
+                e.cause.setStackTrace(e.
```
lihaoyi: Avoid unnecessary re-formatting

repl/src/main/scala/ammonite/repl/Main.scala
lihaoyi: Also, why are we changing the signature from `Res[Imports]` to `Res[Seq[ImportData]]`? That should not be necessary

repl/src/main/scala/ammonite/repl/Main.scala
```diff
@@ -258,7 +265,10 @@ object Main{
       (fileToExecute, codeToExecute) match{
         case (None, None) => println("Loading..."); main.run()
         case (Some(path), None) =>
-          main.runScript(path, passThroughArgs, kwargs.toMap) match{
+          main.runScript(path,
+            passThroughArgs,
+            kwargs.toMap,
+            main.storageBackend) match{
```
lihaoyi: Closing paren on a new line, aligned with the first character of `main .runScript(`

lihaoyi: Closing paren on a new line, aligned with the first character of `main .runScript(`

repl/src/main/scala/ammonite/repl/Main.scala
```diff
@@ -258,7 +265,10 @@ object Main{
       (fileToExecute, codeToExecute) match{
         case (None, None) => println("Loading..."); main.run()
         case (Some(path), None) =>
-          main.runScript(path, passThroughArgs, kwargs.toMap) match{
+          main.runScript(path,
```
lihaoyi: `path` on a new line, aligned with other arguments

repl/src/main/scala/ammonite/repl/Storage.scala
```diff
@@ -42,6 +50,9 @@ object Storage{
     }
 
     var compileCache: mutable.Map[String, (String, CompileCache)] = mutable.Map.empty
+    var classFilesListcache: mutable.Map[String,
+      (String, List[(String, String)], Imports)] = mutable.Map.empty
```
lihaoyi: Should be formatted as 

```scala
var classFilesListcache = mutable.Map.empty[String, (String, List[(String, String)], Imports)]
```

lihaoyi: Also, why is this a `var`?

repl/src/main/scala/ammonite/repl/Storage.scala
```diff
@@ -51,17 +62,46 @@ object Storage{
...
+            Some((res, imports))
+          }
+          else None
+      }
+    }
   }
 
 
-  class Folder(val dir: Path) extends Storage{
+  case class Folder(val dir: Path) extends Storage{
```
lihaoyi: Why is this becoming a `case class`?

repl/src/main/scala/ammonite/repl/interp/Evaluator.scala
```diff
@@ -97,7 +98,7 @@ object Evaluator{
 
       def pop(num: Int = 1) = {
         var next = eval.frames
-        for(i <- 0 until num){
+        for (i <- 0 until num){
```
lihaoyi: Unnecessary reformatting

repl/src/main/scala/ammonite/repl/interp/Interpreter.scala
lihaoyi: Could we call this 
```scala
def reInit()
```

so it's obvious it's related to `init()`?

repl/src/main/scala/ammonite/repl/interp/Interpreter.scala
```diff
@@ -94,46 +103,52 @@ class Interpreter(prompt0: Ref[String],
...
       ammonite.repl.frontend.ReplBridge.repl.replArgs($idx).value.asInstanceOf[${b.typeTag.tpe}]
     """
   }.mkString("\n")
 
-  val predefs = Seq(
-    (hardcodedPredef, Name("HardcodedPredef")),
-    (predef, Name("Predef")),
-    (storage.loadPredef, Name("LoadedPredef")),
-    (argString, Name("ArgsPredef"))
-  )
```
lihaoyi: The changes to this block and the one below it seem unnecessary; if the only thing you wanted to do is add `if !sourceCode.isEmpty` to the for loop, and change `processModule0` to `processModule`, you can do so without changing this whole block.

Also, you should not need to do the `if !sourceCode.isEmpty` check. If your caching is working properly, it would be cached on first run and not actually make things faster. All it does is mask deeper problems, like the problems with a comment-only predef I found earlier

Also

repl/src/main/scala/ammonite/repl/interp/Interpreter.scala
```diff
@@ -94,46 +103,52 @@ class Interpreter(prompt0: Ref[String],
...
+                case None => throw new RuntimeException("Error during Predef: " + msg)
+              }
+
+            case Res.Exception(ex, msg) =>
+              throw new RuntimeException("Error during Predef: " + msg, ex)
+          }
     }
-  }
+
+  updateClassPath
```
lihaoyi: Does this need to be above the `eval.sess.save()`, instead of below it like the original `init()` call is? If not, we should leave it in the same place

repl/src/main/scala/ammonite/repl/interp/Interpreter.scala
```diff
@@ -142,7 +157,8 @@ class Interpreter(prompt0: Ref[String],
     "ivy" -> ImportHook.Ivy
   ))
 
-  def resolveImportHooks(source: ImportHook.Source, stmts: Seq[String]): Res[(Imports, Seq[String])] = {
+  def resolveImportHooks(source: ImportHook.Source,
+                         stmts: Seq[String]): Res[(Imports, Seq[String])] = {
```
lihaoyi: Unnecessary reformatting

repl/src/main/scala/ammonite/repl/interp/Interpreter.scala
```diff
@@ -163,6 +179,7 @@ class Interpreter(prompt0: Ref[String],
       }
     }
 
+
```
lihaoyi: Unnecessary reformatting

repl/src/main/scala/ammonite/repl/interp/Interpreter.scala
```diff
@@ -186,27 +203,38 @@ class Interpreter(prompt0: Ref[String],
...
   }
 
-  def processLine(code: String, stmts: Seq[String], fileName: String): Res[Evaluated] = for{
-    _ <- Catching { case ex =>
-      Res.Exception(ex, "Something unexpected went wrong =(")
-    }
+  def processLine(code: String,
+                  stmts: Seq[String],
+                  fileName: String,
+                  preprocess: Preprocessor = Preprocessor(compiler.parse)
```
lihaoyi: Why is `preprocess` a default argument? Is someone going to pass in a custom version? If not it should be a normal `val` defined in the method body

lihaoyi: In fact, what's the purpose of moving it all the way in here to begin with? Could we leave it in the old position and just make it's initialization `lazy`?

repl/src/main/scala/ammonite/repl/interp/Preprocessor.scala
```diff
@@ -90,11 +90,12 @@ object Preprocessor{
                   imports: Imports,
                   printerTemplate: String => String) = for{
       Preprocessor.Expanded(code, printer) <- expandStatements(stmts, resultIndex)
+
```
lihaoyi: Unnecessary reformatting

repl/src/main/scala/ammonite/repl/interp/Interpreter.scala
lihaoyi: bump @coderabhishek

repl/src/main/scala/ammonite/repl/interp/Interpreter.scala
```diff
@@ -517,14 +629,15 @@ class Interpreter(prompt0: Ref[String],
 
       def exec(file: Path): Unit = apply(read(file))
 
-      def module(file: Path): Unit = {
+
+      def module(file: Path) = {
         val (pkg, wrapper) = Util.pathToPackageWrapper(file, wd)
-        processModule(ImportHook.Source.File(wd/"<console>"), read(file), wrapper, pkg) match{
+        processModule(ImportHook.Source.File(wd / "<console>"), read(file), wrapper, pkg) match {
```
lihaoyi: Unnecessary reformatting

repl/src/main/scala/ammonite/repl/interp/Interpreter.scala
```diff
@@ -517,14 +629,15 @@ class Interpreter(prompt0: Ref[String],
 
       def exec(file: Path): Unit = apply(read(file))
 
-      def module(file: Path): Unit = {
+
```
lihaoyi: Unnecessary reformatting

repl/src/main/scala/ammonite/repl/interp/Preprocessor.scala
```diff
@@ -301,6 +301,7 @@ ${importBlock(imports)}
 
 object ${indexedWrapperName.backticked}{\n"""
 
+
```
lihaoyi: Unnecessary reformatting

repl/src/main/scala/ammonite/repl/interp/Preprocessor.scala
```diff
@@ -196,7 +197,6 @@ object Preprocessor{
         case Nil => Res.Skip
         case postSplit =>
           complete(stmts.mkString(""), wrapperIndex, postSplit)
-
```
lihaoyi: Formatting

repl/src/main/scala/ammonite/repl/interp/Interpreter.scala
```diff
@@ -311,14 +334,81 @@ class Interpreter(prompt0: Ref[String],
...
+               cacheTag)
+           res.map(_._1)
+         case r =>
+           r.asInstanceOf[Res[Imports]]
+       }
+      case Some((cachedData, imports)) =>
+        evalCachedClassFiles(cachedData,
+          pkgName.map(_.backticked).mkString("."),
+          wrapperName.backticked) match {
+          case Res.Success(_) =>
```
lihaoyi: Indentation

repl/src/main/scala/ammonite/repl/Storage.scala
```diff
@@ -90,6 +130,53 @@ object Storage{
...
+    def classFilesListSave(pkg: String,
+                           wrapper: String,
+                           dataList: List[(String, String)],
+                           imports: Imports,
+                           tag: String): Unit = {
+
+      val dir = pkg.replace("/", "$div") + "." + wrapper.replace("/", "$div") + "0"
+      val codeCacheDir = compileCacheDir/dir
+        if(!exists(codeCacheDir)) {
+        mkdir(codeCacheDir)
```
lihaoyi: Indentation

repl/src/main/scala/ammonite/repl/interp/Preprocessor.scala
```diff
@@ -91,10 +91,10 @@ object Preprocessor{
                   printerTemplate: String => String) = for{
       Preprocessor.Expanded(code, printer) <- expandStatements(stmts, resultIndex)
       (wrappedCode, importsLength) = wrapCode(
-        pkgName, indexedWrapperName, leadingSpaces + code,
-        printerTemplate(printer.mkString(", ")),
-        imports
-      )
+      pkgName, indexedWrapperName, leadingSpaces + code,
```
lihaoyi: bad indentation

repl/src/main/scala/ammonite/repl/Main.scala
```diff
@@ -268,6 +280,7 @@ object Main{
               ex.setStackTrace(trace.take(i))
               throw ex
             case Res.Success(_) =>
+
```
lihaoyi: Formatting...

repl/src/test/scala/ammonite/repl/ScriptTests.scala
```diff
@@ -280,25 +279,48 @@ object ScriptTests extends TestSuite{
...
+          java.nio.file.Files.createTempDirectory("ammonite-tester")
+        )
 
         'blocks{
           val cases = Seq("OneBlock.scala" -> 2, "TwoBlocks.scala" -> 3, "ThreeBlocks.scala" -> 4)
           for((fileName, expected) <- cases){
             val storage = Storage.InMemory()
             val interp = createTestInterp(storage)
             val n0 = storage.compileCache.size
-
```
lihaoyi: Formatting...

repl/src/main/scala/ammonite/repl/interp/Interpreter.scala
```diff
@@ -311,14 +334,81 @@ class Interpreter(prompt0: Ref[String],
...
+
+    // Iterate through cached blocks and execute classFiles
+    // blockNumber keeps track of blockIndex
+
+    Res[Unit](Try {
+      cachedData.foreach{ d =>
+         for {
+          cls <- eval.loadClass(pkg + "." + wrapper + getBlockNumber, d._1)
+        } yield {
+           Compiler.addToClasspath(d._1, dynamicClasspath)
```
lihaoyi: Formatting

repl/src/main/scala/ammonite/repl/interp/Interpreter.scala
coderabhishek: It is exceeding 100 chars otherwise

repl/src/main/scala/ammonite/repl/interp/Interpreter.scala
coderabhishek: I think this code is ok, can you point me exact formatting mistake?

lihaoyi: Why are you adding three extra blank lines here? It has nothing to do with your change, and is more lines than we should need anyway: 1 or 2, which we already had before

repl/src/main/scala/ammonite/repl/interp/Interpreter.scala
```diff
@@ -345,34 +433,42 @@ class Interpreter(prompt0: Ref[String],
...
-          s"Main$wrapperIndex.scala",
-          indexedWrapperName
-        )
-      }
-    )
-  } yield res ++ hookImports
+
+
+  def processExec(code: String): Res[Imports] = {
+    init()
```
coderabhishek: For this code should I put `init` inside `_  <- Res.Success(init())` in for comprehension that will just add one line, instead of changing whole function?

lihaoyi: This is fine

repl/src/main/scala/ammonite/repl/interp/Interpreter.scala
lihaoyi: ok

repl/src/main/scala/ammonite/repl/interp/Interpreter.scala
coderabhishek: Ok I have changed this. Will get updated in next commit

repl/src/main/scala/ammonite/repl/interp/Interpreter.scala
coderabhishek: Done, Only meant the missing line at the end, is that right?

repl/src/main/scala/ammonite/repl/Util.scala
```diff
@@ -124,7 +124,8 @@ case class Catching(handler: PartialFunction[Throwable, Res.Failing]) {
 }
 
 case class Evaluated(wrapper: Seq[Name],
-                     imports: Imports)
+                     imports: Imports,
+                     tag: String = "")
```
lihaoyi: Shouldn't use a default argument here; explicitly provide it in the places where it's necessary

repl/src/test/scala/ammonite/repl/ScriptTests.scala
```diff
@@ -271,7 +271,7 @@ object ScriptTests extends TestSuite{
       }
       'caching{
 
-        def createTestInterp(storage: Storage) = new Interpreter(
+        def createTestInterp(storage: Storage, predefS: String = "") = new Interpreter(
```
lihaoyi: Get rid of this `predefS` argument and just use `ammonite.repl.Main.defaultPredefString` directly

repl/src/main/scala/ammonite/repl/interp/Interpreter.scala
```diff
@@ -311,12 +328,70 @@ class Interpreter(prompt0: Ref[String],
...
+       }
+      case Some((cachedData, imports)) =>
+        evalCachedClassFiles(
+          cachedData,
+          pkgName.map(_.backticked).mkString("."),
+          wrapperName.backticked
+        ) match {
+          case Res.Success(_) =>
+            Res.Success(imports)
+          case r => r.asInstanceOf[Res[Imports]]
```
lihaoyi: When you changed the snippet 10 lines above, can you change this one too?

repl/src/main/scala/ammonite/repl/Util.scala
```diff
@@ -324,6 +328,7 @@ object Util{
 object Timer{
   var current = 0L
   var show = false
+  var startTime = System.nanoTime()
```
lihaoyi: What is this?

repl/src/main/scala/ammonite/repl/Storage.scala
```diff
@@ -90,6 +130,54 @@ object Storage{
...
+      if(!exists(codeCacheDir)) None
+      else {
+        val metadataJson = Try {
+          read(codeCacheDir/classFilesOrder)
+        }.toOption
+        val impFile = Try {
+          read(codeCacheDir/"imports.list")
+        }.toOption
+        val (loadedTag, classFilesList) =
+          upickle.default.read[(String, Seq[(String, String)])](metadataJson.getOrElse(""))
```
lihaoyi: Why is there a `getOrElse("")` here? That will just cause `read` to throw and exception. What's the point?

repl/src/main/scala/ammonite/repl/Storage.scala
```diff
@@ -90,6 +130,54 @@ object Storage{
...
+
+        if (cacheTag == loadedTag){
+          val res = {
+            for((path, tag) <- classFilesList) yield {
+              compileCacheLoad(path, tag)
+            }
+          }.flatten
+          val imports = upickle.default.read[Imports](impFile.getOrElse(""))
+          Some((res.unzip._1, imports))
+        }
```
lihaoyi: This if-statement is copy-pasted from above. Can you do something about it?

repl/src/main/scala/ammonite/repl/interp/Interpreter.scala
```diff
@@ -423,14 +509,26 @@ class Interpreter(prompt0: Ref[String],
...
+            wrapperIndex + 1,
+            compiledData
+          )
         }
       }
     }
     // wrapperIndex starts off as 1, so that consecutive wrappers can be named
     // Wrapper, Wrapper2, Wrapper3, Wrapper4, ...
-    try loop(blocks, startingImports, Imports(), wrapperIndex = 1 )
+    try loop(blocks, startingImports, Imports(Nil), wrapperIndex = 1, List())
```
lihaoyi: Why the additional `Nil` here?

repl/src/main/scala/ammonite/repl/interp/Interpreter.scala
```diff
@@ -627,7 +723,7 @@ object Interpreter{
       Util.md5Hash(imports.iterator.map(_.toString.getBytes)),
       classpathHash
     ))
-    "cache" + bytes.map("%02x".format(_)).mkString //add prefix to make sure it begins with a letter
+    bytes.map("%02x".format(_)).mkString //add prefix to make sure it begins with a letter
```
lihaoyi: This comment is out of date

repl/src/main/scala/ammonite/repl/interp/Interpreter.scala
```diff
@@ -311,12 +328,70 @@ class Interpreter(prompt0: Ref[String],
...
+                           wrapper: String): Res[Seq[Unit]] = {
+
+    def getBlockNumber(blockNumber: Int) = blockNumber match {
+      case 1 => ""
+      case _ => "_" + blockNumber.toString
+    }
+
+      val cls = Res.map(cachedData.zipWithIndex)(
+        (data: (ClassFiles, Int)) => {
+          val (clsFiles, index) = data
```
lihaoyi: This should be

```scala
case (clsFiles, index) =>
```

without the curlies

repl/src/main/scala/ammonite/repl/interp/Interpreter.scala
```diff
@@ -117,7 +125,7 @@ class Interpreter(prompt0: Ref[String],
   var predefImports = Imports()
   for( (sourceCode, wrapperName) <- predefs) {
     val pkgName = Seq(Name("ammonite"), Name("predef"))
-    processModule0(ImportHook.Source.File(wd/"<console>"), sourceCode, wrapperName, pkgName, predefImports) match{
```
lihaoyi: Why the change from `<console>` to `Main.scala`?

repl/src/main/scala/ammonite/repl/interp/Interpreter.scala
```diff
@@ -311,12 +328,70 @@ class Interpreter(prompt0: Ref[String],
...
       _ = Timer("cachedCompileBlock 4")
       cls <- eval.loadClass(fullyQualifiedName, classFiles)
-    } yield (cls, newImports)
+    } yield (cls, newImports, tag)
+  }
+
+
+  def evalCachedClassFiles(cachedData: Seq[ClassFiles],
+                           pkg: String,
+                           wrapper: String): Res[Seq[Unit]] = {
```
lihaoyi: This should be a `Res[Unit]`, not `Res[Seq[Unit]]`

repl/src/main/scala/ammonite/repl/interp/Interpreter.scala
```diff
@@ -311,12 +328,70 @@ class Interpreter(prompt0: Ref[String],
...
+    }
+
+      val cls = Res.map(cachedData.zipWithIndex)(
+        (data: (ClassFiles, Int)) => {
+          val (clsFiles, index) = data
+          Compiler.addToClasspath(clsFiles, dynamicClasspath)
+          eval.loadClass(pkg + "." + wrapper + getBlockNumber(index + 1), clsFiles)
+        }
+      )
+      cls.map(_.map(eval.evalMain(_)))
```
lihaoyi: You should use a `for`-comprehension instead of `.map` on the `Res`. 

The name `cls` is also misleading, since this isn't a `Class`, it is a `res`.

Furthermore, this is incorrect, as an exception inside `evalMain` will get thrown and not caught. You need to use a `_ <- Catching{userCodeExceptionHandler}` inside your for-comprehension to catch that.

Lastly, this stuff which has to do with evaluating compiled class files should live in `Evaluator`, not in `Interpreter`

repl/src/main/scala/ammonite/repl/interp/Interpreter.scala
```diff
@@ -311,12 +328,70 @@ class Interpreter(prompt0: Ref[String],
...
+             pkgName.map(_.backticked).mkString("."),
+             wrapperName.backticked,
+             files,
+             imports,
+             cacheTag
+           )
+           res.map(_._1)
+         case r: Res.Failing => r
+       }
+      case Some((cachedData, imports)) =>
```
lihaoyi: You are failing to cache import hooks, and thus any necessary magic imports like `import $file.foo` to load secondary files or Ivy dependencies will not run on cached scripts, which will cause failures.

repl/src/main/scala/ammonite/repl/interp/Interpreter.scala
```diff
@@ -258,23 +274,24 @@ class Interpreter(prompt0: Ref[String],
         )
 
       }
+    // empty quotes to serve as a dummy cacheTag for repl commands which are not to be cached
```
lihaoyi: Stale comment

repl/src/main/scala/ammonite/repl/interp/Interpreter.scala
```diff
@@ -311,12 +328,70 @@ class Interpreter(prompt0: Ref[String],
...
+    storage.classFilesListLoad(
+      pkgName.map(_.backticked).mkString("."),
+      wrapperName.backticked,
+      cacheTag
+    ) match {
+      case None =>
+        init()
+        val res = processModule0(source, code, wrapperName, pkgName, predefImports)
+        res match{
+         case Res.Success(data) =>
```
lihaoyi: This should be 

```scala
case Res.Success((imports, files)) => 
```

lihaoyi:
Looks like we're not done yet. 

You are missing two sets of unit tests, given the potential failures I am seeing in your code:

- Tests than runtime exceptions within cached scripts are properly caught and returned in a `Res.Failing` result
- Tests that magic imports such as `import $file.Foo` or `import $ivy.`com.lihaoyi::scalatags:0.5.4` work in cached scripts.

repl/src/main/scala/ammonite/repl/Util.scala
coderabhishek: This I was using for testing, if someone wants to check time elapsed between two points in program flow, one can set the `startTime` and print by subtracting from the `System.Nanotime` at the end. I will remove this if you think it won't be necessary in future otherwise I will put a comment explaining its  use

repl/src/main/scala/ammonite/repl/interp/Interpreter.scala
coderabhishek: Removed

repl/src/test/scala/ammonite/repl/ScriptTests.scala
coderabhishek: In some cases it creates interp with empty predef to test

repl/src/main/scala/ammonite/repl/Util.scala
lihaoyi: Let's remove this for now

repl/src/test/scala/ammonite/repl/ScriptTests.scala
lihaoyi: Oh ok that's fine then, can we remove the default argument and pass in the `predefS` manually for those cases?

Also, let's just call it `predef` instead of `predefS`

repl/src/main/scala/ammonite/repl/interp/Evaluator.scala
```diff
@@ -170,25 +172,28 @@ object Evaluator{
         // Exhaust the printer iterator now, before exiting the `Catching`
         // block, so any exceptions thrown get properly caught and handled
 
-        val iter = evalMain(cls).asInstanceOf[Iterator[String]]
+        val iter = cls.getDeclaredMethod("$main").invoke(null).asInstanceOf[Iterator[String]]
```
lihaoyi: I am not convinced that this is doing anything. Can you revert it?

repl/src/main/scala/ammonite/repl/Storage.scala
```diff
@@ -90,6 +133,63 @@ object Storage{
...
+      val dir = pkg.replace("/", "$div") + "." + wrapper.replace("/", "$div") + "0"
+      val codeCacheDir = compileCacheDir/dir
+      if(!exists(codeCacheDir)) {
+        mkdir(codeCacheDir)
+        write(
+          codeCacheDir/classFilesOrder,
+          upickle.default.write((tag, dataList.reverse), indent = 4)
+        )
+        write(codeCacheDir/"imports.list", upickle.default.write(imports))
+        write(codeCacheDir/"importTrees.list", upickle.default.write(importTreesList))
```
coderabhishek: Store the importTree also and load it and pass each importTree to ResolveSingleImportHook, to load imports

repl/src/test/scala/ammonite/repl/ScriptTests.scala
```diff
@@ -299,6 +303,55 @@ object ScriptTests extends TestSuite{
...
+          )
+          val res = intercept[java.lang.ArithmeticException]{
+            interp2.replApi.load.module(
+              resourcesPath/'scriptLevelCaching/"runTimeExceptions.scala"
+            )
+          }
+          assert(interp2.compiler == null &&
+            res.toString == "java.lang.ArithmeticException:/by zero")
+        }
+
```
coderabhishek: runTimeExceptions.scala reads a number from file num.val and prints `5/num.val`, first it is 1 and then changed to zero by the script, so when next time it reads it causes a run-time exception

repl/src/test/resources/scriptLevelCaching/runTimeExceptions.scala
```diff
@@ -0,0 +1,4 @@
+import ammonite.ops._
+println(5/read(cwd/'repl/'src/'test/'resources/'scriptLevelCaching/"num.value").trim.toInt)
```
lihaoyi: You should make this read and write from `repl/target/scriptCachingTests/`, not the `src/` folder which is for source code

repl/src/main/scala/ammonite/repl/Storage.scala
lihaoyi: These should both be `.json` files since they contain JSON, and you should pass in `indent = 4` to make the pretty-printed and human readable

repl/src/main/scala/ammonite/repl/Storage.scala
```diff
@@ -90,6 +133,63 @@ object Storage{
       }
     }
 
+    def classFilesListSave(pkg: String,
+                           wrapper: String,
+                           dataList: Seq[(String, String)],
+                           imports: Imports,
+                           tag: String,
+                           importTreesList: Seq[ImportTree]): Unit = {
+      val dir = pkg.replace("/", "$div") + "." + wrapper.replace("/", "$div") + "0"
```
lihaoyi: As discussed, we should not randomly prefix things with `0`: these should go in a separate folder from the `cache/compile/`, e.g. `cache/scriptLevel/`

repl/src/main/scala/ammonite/repl/Storage.scala
```diff
@@ -90,6 +133,63 @@ object Storage{
...
+      val codeCacheDir = compileCacheDir/dir
+      if(!exists(codeCacheDir)) {
```
lihaoyi: This `exists` check is not necessary; `mkdir` does it for you already

repl/src/test/scala/ammonite/repl/ScriptTests.scala
```diff
@@ -299,6 +303,55 @@ object ScriptTests extends TestSuite{
...
+            assert(interp2.compiler == null)
+            interp2.replApi.load.module(resourcesPath/script)
+            assert(interp2.compiler == null)
+          }
+
+          'testOne - check('scriptLevelCaching/"scriptTwo.scala")
+          'testTwo - check('scriptLevelCaching/"scriptOne.scala")
+          'testThree - check('scriptLevelCaching/"QuickSort.scala")
+          'testLoadModule - check('scriptLevelCaching/"testLoadModule.scala")
+          'testFileImport - check('scriptLevelCaching/"testFileImport.scala")
```
lihaoyi: We should have another test here to use `import $ivy` to verify that that keeps working inside cached scripts

repl/src/main/scala/ammonite/repl/interp/Evaluator.scala
coderabhishek: I don't know why but it is necessary, without that weird `BoxedUnit`error starts coming for repl cmds

repl/src/test/scala/ammonite/repl/ScriptTests.scala
coderabhishek: Added!

lihaoyi:
Looks good to me!

PR #406, opened 2016-07-03
Fix Small Errors in Implement Script Caches Commit

coderabhishek:
Imports cache file changed to ".json" instead of ".list"


lihaoyi:
@coderabhishek Can you make this PR against master?

PR #408, opened 2016-07-06
Change extension of Ammonite files from .scala to .sc

coderabhishek:
This diff makes Ammonite run `.sc` scripts instead of `.scala` scripts to avoid confusion with regular scala files


lihaoyi:
Looks good, thanks!

PR #409, opened 2016-07-08
Make integration tests pass on windows

coderabhishek:
All integration tests are passing on windows except those which use ivy cause ivy-settings for windows are different


lihaoyi:
This looks good

PR #411, opened 2016-07-14
Make most of  the amm tests and integration tests pass on Windows

coderabhishek:
All integration and Amm tests except those which use `ivy`, or Syntax Error messages are passing. There are few tests in Amm/importHookTests having few windows path problems.

Syntax Error tests are failing on windows due to fastparse showing lesser number of characters as `\r\n` consume one extra character in the hard coded ` show 10 chars` thing in fastparse.

Ivy fails on windows as ivy settings for windows are different.

Major changes:
1) Change `splitScript` function in Preprocessor.scala to update `ncomment` variable with appropriate newlines and remove extra `\r` on windows. 
2) Update `\n` everywhere with `System.lineSeparator` defined as `newLine` in `Util.scala`
3) Modify test cases wherever necessary and guard those which can't pass now
4) Normalize newlines wherever reading from files or a hard-coded string is there

### DIFF
--- a/amm/src/main/scala/ammonite/Main.scala
+++ b/amm/src/main/scala/ammonite/Main.scala
@@ -9,6 +9,7 @@ import ammonite.main.{Defaults, Repl, Router}
 import ammonite.main.Router.{ArgSig, EntryPoint}
 import ammonite.util.Name.backtickWrap
 import ammonite.util._
+import ammonite.util.Util.newLine
 
 
 
@@ -60,7 +61,7 @@ case class Main(predef: String = "",
     new Repl(
       inputStream, outputStream, errorStream,
       storage = storageBackend,
-      predef = augmentedPredef + "\n" + predef,
+      predef = augmentedPredef + newLine + predef,
       wd = wd,
       welcomeBanner = welcomeBanner,
       replArgs = replArgs,
@@ -144,7 +145,7 @@ object Main{
             |
             |You can also use `--` as a shorthand for `-x main`, to pass arguments
             |to the main method
-          """.stripMargin.replace("\n", "\n" + " " * 8)
+          """.stripMargin.replace("\n", newLine + " " * 8)
         )
       arg[String]("<args>...")
         .optional()
--- a/amm/src/main/scala/ammonite/interp/ImportHook.scala
+++ b/amm/src/main/scala/ammonite/interp/ImportHook.scala
@@ -108,7 +108,7 @@ object ImportHook{
                 ))
 
                 Result.Source(
-                  read(filePath),
+                  Util.normalizeNewlines(read(filePath)),
                   wrapper,
                   pkg,
                   ImportHook.Source.File(filePath),
--- a/amm/src/main/scala/ammonite/interp/Preprocessor.scala
+++ b/amm/src/main/scala/ammonite/interp/Preprocessor.scala
@@ -3,6 +3,7 @@ package ammonite.interp
 import acyclic.file
 import ammonite._
 import ammonite.util._
+import ammonite.util.Util.{windowsPlatform, newLine, normalizeNewlines}
 import fastparse.all._
 
 import scala.reflect.internal.Flags
@@ -43,12 +44,14 @@ object Preprocessor{
   def errMsg(msg: String, code: String, expected: String, idx: Int): String = {
     val locationString = {
       val (first, last) = code.splitAt(idx)
-      val lastSnippet = last.split('\n').headOption.getOrElse("")
-      val firstSnippet = first.reverse.split('\n').lift(0).getOrElse("").reverse
-      firstSnippet + lastSnippet + "\n" + (" " * firstSnippet.length) + "^"
+      val lastSnippet = last.split(newLine).headOption.getOrElse("")
+      val firstSnippet = first.reverse
+                              .split(newLine.reverse)
+                              .lift(0).getOrElse("").reverse
+      firstSnippet + lastSnippet + newLine + (" " * firstSnippet.length) + "^"
     }
 
-    s"Syntax Error: $msg\n$locationString"
+    s"Syntax Error: $msg${newLine}$locationString"
   }
 
   /**
@@ -68,10 +71,22 @@ object Preprocessor{
 
         // comment holds comments or empty lines above the code which is not caught along with code
         for( (comment, code) <- s.value){
-          val ncomment = comment + "\n"*offset
 
-          // 1 is added as Separator parser eats up the '\n' following @
-          offset = offset + comment.count(_ == '\n') + code.map(_.count(_ == '\n')).sum + 1
+          //ncomment has required number of newLines appended based on OS and offset
+          //since fastparse has hardcoded `\n`s, while parsing strings with `\r\n`s it
+          //gives out one extra `\r` after '@' i.e. block change
+          //which needs to be removed to get correct line number (It adds up one extra line)
+          //thats why the `comment.substring(1)` thing is necessary
+          val ncomment =
+            if(windowsPlatform && !blocks.isEmpty && !comment.isEmpty){
+              comment.substring(1) + newLine * offset
+            }else{
+              comment + newLine * offset
+            }
+
+          // 1 is added as Separator parser eats up the newLine char following @
+          offset = offset + (comment.split(newLine, -1).length - 1) +
+            code.map(_.split(newLine, -1).length - 1).sum + 1
           blocks.append((ncomment, code))
         }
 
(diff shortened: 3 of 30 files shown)


amm/src/main/scala/ammonite/Main.scala
```diff
@@ -60,7 +60,7 @@ case class Main(predef: String = "",
     new Repl(
       inputStream, outputStream, errorStream,
       storage = storageBackend,
-      predef = augmentedPredef + "\n" + predef,
+      predef = augmentedPredef + System.lineSeparator() + predef,
```
lihaoyi: Could we provide an alias to `System.lineSeparator()` from something shorter? e.g. `newLine` or something in the Utils.scala

amm/src/main/scala/ammonite/interp/ImportHook.scala
```diff
@@ -108,7 +108,7 @@ object ImportHook{
                 ))
 
                 Result.Source(
-                  read(filePath),
+                  read(filePath).replace("\r", "").replace("\n", System.lineSeparator()),
```
lihaoyi: We're doing a bunch of ad-hoc `.replace`s all over the place; let's aggregate them all into a helper function `normalizeNewlines(s: String)` and use that everywhere

integration/src/test/scala/ammonite/integration/TestUtils.scala
```diff
@@ -2,11 +2,16 @@ package ammonite.integration
 
 import ammonite.ops._
 import ImplicitWd._
+import ammonite.util.Util
 
 /**
   * Created by haoyi on 6/4/16.
   */
 object TestUtils {
+  val newLine = Util.windowsPlatform match{
```
lihaoyi: As far as I can tell this is currently unnecessary; it's only used in one place and you surrounded that with an `if (Util.windowsPlatform)` anyway

amm/src/main/scala/ammonite/interp/Preprocessor.scala
```diff
@@ -68,10 +70,21 @@ object Preprocessor{
 
         // comment holds comments or empty lines above the code which is not caught along with code
         for( (comment, code) <- s.value){
-          val ncomment = comment + "\n"*offset
 
-          // 1 is added as Separator parser eats up the '\n' following @
-          offset = offset + comment.count(_ == '\n') + code.map(_.count(_ == '\n')).sum + 1
+          //ncomment has required number of newLines appended based on OS and offset
+          //fastparse behaves differently for "\n" and "\r\n"
+          // thats why the `comment.substring(1)` thing is necessary
```
lihaoyi: This comment is an insufficient explanation of what is going on. You will need to elaborate on it to make it understandable

lihaoyi:
Those are the only things I noticed for now; when you update it I'd keep an eye on the oracleJDK7/Scala2.11.x builds; they both failed, and I'm not sure if it was random or whether it's a sign of a consistent failure on those platforms. We'll see when you update

amm/src/main/scala/ammonite/interp/Preprocessor.scala
```diff
@@ -293,12 +309,12 @@ object Preprocessor{
 package ${pkgName.map(_.backticked).mkString(".")}
 ${importBlock(imports)}
 
-object ${indexedWrapperName.backticked}{\n"""
+object ${indexedWrapperName.backticked}{\n""".replace("\n", newLine)
```
lihaoyi: You should be able to directly interpolate `$newLine` into the string here, and below, rather than leaving the `\n` inside the string and then calling `.replace` to swap it out

integration/src/test/scala/ammonite/integration/LineNumberTests.scala
```diff
@@ -13,36 +14,48 @@ object LineNumberTests extends TestSuite{
   val tests = this{
 
     def checkErrorMessage(file: RelPath, expected: String) = {
-      val e = intercept[ShelloutException]{ 
+      val e = intercept[ShelloutException]{
         exec(file)
-      }.result.err.string.replace("\r", "").replace("\n", System.lineSeparator())
-      assert(e.contains(expected.replace("\n", System.lineSeparator())))
+      }.result.err.string
+      assert(e.contains(expected.replace("\n", Util.newLine)))
```
lihaoyi: Should we be using `normalizeNewlines` here as well?

integration/src/test/scala/ammonite/integration/LineNumberTests.scala
```diff
@@ -13,36 +14,48 @@ object LineNumberTests extends TestSuite{
...
 
-    'errorTest - checkErrorMessage(
-      file = 'lineNumbers/"ErrorLineNumberTest.sc",
-      expected =
-        """Syntax Error: ("}" | `case`):5:24 ...")\n  }\n\n  d"
-          |    printlnqs(unsorted))
-          |                       ^""".stripMargin
-    )
+    'errorTest - {
+      if(!Util.windowsPlatform) {
```
lihaoyi: Can you add a comment saying why these are skipped? I know it's probably something to do with fastparse error messages, as we discussed, but we should write it down so 6 months later we won't forget it when reading the code

amm/src/main/scala/ammonite/interp/Preprocessor.scala
```diff
@@ -68,10 +71,23 @@ object Preprocessor{
...
+          //gives out one extra `\r` after '@' i.e. block change
+          //which needs to be removed to get correct line number (It adds up one extra line)
+          //thats why the `comment.substring(1)` thing is necessary
+          val ncomment = windowsPlatform match {
+            case true =>
+              if(!blocks.isEmpty && !comment.isEmpty)
+                comment.substring(1) + newLine * offset
+              else comment + newLine * offset
+            case false => comment + newLine * offset
+          }
```
lihaoyi: This should probably be re-written as 

```scala
val ncomment = 
  if(windowsPlatform && !blocks.isEmpty && !comment.isEmpty){
    comment.substring(1) + newLine * offset  
  }else{
    comment + newLine * offset
  }
 ```

**Comment by lihaoyi** [comment]
This looks good. I left some final notes; take care of those, squash your commits, and update the PR title/description to have a nice detailed summary of what changes you made, which tests remain disabled (it's not just the ivy tests!) and why they are disabled. I'll merge it after that

File: amm/src/main/scala/ammonite/interp/Preprocessor.scala (hunk 5 above)
Comment (coderabhishek) [inline amm/src/main/scala/ammonite/interp/Preprocessor.scala]: Since this string (actually whole file) is written on a linux machine the formatting due `"""` puts `\n`s after each line of in string. like 
`"""
This line

next line
""" is "\nThisLine\n\nnext line\n"

File: integration/src/test/scala/ammonite/integration/LineNumberTests.scala (hunk 6 above)
Comment (coderabhishek) [inline integration/src/test/scala/ammonite/integration/LineNumberTests.scala]: I have changed for this instance but there are numerous other instances in test expected outputs where `.replace("\n", Util.newLine` is used as strings written on linux appear different as I mentioned above. I think replace will be ok and natural for those places like this
`
          s"""The following arguments failed to be parsed:
             |(i: Int) failed to parse input "foo" with $exMsg
             |expected arguments: (i: Int, s: String, path: ammonite.ops.Path)"""
            .stripMargin
            .replace("\n", Util.newLine)
`
If you say I will change them also

Comment (lihaoyi) [inline integration/src/test/scala/ammonite/integration/LineNumberTests.scala]: Yes let's change all of them to `normalizeNewlines` to be consistent.

File: amm/src/main/scala/ammonite/interp/Preprocessor.scala [hunk 9]
```diff
@@ -289,16 +304,16 @@ object Preprocessor{
...
 
-object ${indexedWrapperName.backticked}{\n"""
+object ${indexedWrapperName.backticked}{\n""")
 
-    val bottomWrapper = s"""\ndef $$main() = { $printCode }
+    val bottomWrapper = normalizeNewlines(s"""\ndef $$main() = { $printCode }
   override def toString = "${indexedWrapperName.raw}"
 }
-"""
+""")
```
coderabhishek: Need to normalize even as even if I remove `\n`s since this string is written on a linux machine the formatting due to `"""` will add `\n`s like 
`
"""
A

B
"""
`
will become "\nA\n\nB\n", so anyway normalize is necessary

lihaoyi: I do not understand your explanation; what's wrong with "\nA\n\nB\n"?

coderabhishek: On windows also it will remain '\n' only, though it does not affect the line numbers or anything else. I thought it would be better to avoid wrong newline chars on windows as `top(bottom)Wrapper` gets appended as it is to the code while code already has `\r\n` on windows.

lihaoyi:
Looks good to me, thanks!

Issue #292, opened 2015-12-08
Ammonite's dependencies conflict with others on classpath


bblfish:
I was trying to run code [from my branch of upickle](https://github.com/bblfish/upickle-pprint/commit/597ba2573228293f02dbf70ac64a2ac79197bf07) [related to issue 122](https://github.com/lihaoyi/upickle-pprint/pull/122) using Ammonite started from sbt from  inside the `upickleJVM` project. 

I tried what the docs suggested to place in `global.sbt` and the following

```
libraryDependencies += "com.lihaoyi" % "ammonite-repl" % "0.5.1" cross CrossVersion.full

initialCommands in (console) := """ammonite.repl.Main.run("")"""
```

I I then run `compile:console` and end up in the **amm** console.

But it looks like I cannot create a `Js.Unserializable("hi")` there. The class is not found even though it is in the code.

But when  I `^D` from amm back to the  normal scala console started by sbt, I can create an instance of `Js.Unserializable`

lihaoyi:
The basic problem here is that we have no isolation between Ammonite and the stuff we run inside it. That means basically anything in this list

https://github.com/lihaoyi/Ammonite/blob/master/build.sbt#L107-L115

Will have some kind of weird mis-behavior if you try to develop using Ammonite, or even if you're just using a different version of that library in the same project. Possible solutions include:

- JarJar https://github.com/Tapad/sbt-jarjar, which IIRC does automatic-vendoring
- Classloader isolation

fommil:
oh, don't you set the `usejavacp=false` flag? e.g. https://github.com/ensime/ensime-server/blob/master/core/src/main/scala/org/ensime/core/Analyzer.scala#L66

lihaoyi:
Yeah I do, and then I walk the classpath myself and gather up the jars anyway =P

fommil:
ah, right, so you're intentionally adding ammonite onto the user's classpath. I see.

lihaoyi:
Yeah, well, that's not the problem. The problem is that I want to add the *user's* classpath to *Ammonite's*. After all, I want to interact with it in the REPL! However, Ammonite has its own dependencies, and when it conflicts with the user's it's a pain. Hence the possible solutions above

lihaoyi:
Since nobody else has complained about this I'm going to close this for now

PR #412, opened 2016-07-18
Add support for class loader isolation

alexarchambault:
This PR allows to isolate parts of Ammonite from user sessions.

It splits the `amm` module into `amm-api` (things available to users), `amm-tools` (extra things available to users, less essential), and `amm` (Ammonite itself, like the current `amm` module, with implementations for what's in `amm-api`).

As is, it can be tested by launching Ammonite via coursier, like
```
$ sbt published/publishLocal
$ coursier launch \
      -i ammonite \
      -I ammonite:com.lihaoyi:ammonite-tools_2.11.8:0.6.3-SNAPSHOT \
      com.lihaoyi:ammonite_2.11.8:0.6.3-SNAPSHOT
```
Launched this way, coursier sets up a first base class loader, with the user-facing things (`amm-api` and `amm-tools` modules, with their dependencies), then a second one with the former as parent for `amm` itself. The code in the last commit of this PR is able to find the first class loader, and use it as a base for the Ammonite session class loaders. Thus not exposing `amm` itself in the session, only `amm-api` and `amm-tools`.

Launched in other ways, there's no isolation like before.

There would be ways to have the generated assembly be fine with isolation, that I'll explain if you're fine with the idea. With this launcher, integration tests could be written for isolation.


alexarchambault:
The first 12 commits mainly split existing things, in things that will stay in `amm` and things that will move to `amm-api` and `amm-tools` (all the moves are done in the `Add ammonite-api...` and `Add ammonite-tools...` commits).

alexarchambault:
In particular, this allows not to have upickle, fastparse / scalaparse, scalaj-http, in the Ammonite session.

lihaoyi:
Thanks for the PR!

Honestly, I cannot follow what this PR is doing. There are a lot of good things in there: the splitting up of the `amm/` module is one I have been thinking of for a while. 

But it's mixed in with a bunch of peripheral stuff like renaming `ImportHook` to `ImportHooks`, or changing the style of imports from `import Thing; Thing.foo(...)` to `import Thing.foo; foo(...)`, which most definitely aren't necessary to what it's trying to do.

I also do not know how this works, or doesn't work. I have no idea what the coursier CLI tool does, and any explanation of what this PR does relating to how to do stuff using the coursier executable has no impact on me. It is not clear from reading the code.

I think the idea of classloader isolation is useful and wanted, it needs to be discussed in isolation. If you're interested in pushing this forward, I'd suggest starting with one of the thing that's in this PR:

- Moving things around (???)
- Breaking up `amm/` into sub-components (It's not clear to me that `tools` and `api` is the most useful split)

And get them into master one by one, properly reviewed. Then when we get to the classloader isolation bit, it should be easier to discuss without being distracted by irrelevant changes.

alexarchambault:
About the move things around, as said above, they split things into chunks that later stay in `amm`, and chunks that are moved to `amm-tools` or `amm-api`. E.g. about `ImportHook` / `ImportHooks`, `ImportHook` goes to `amm-api` whereas the implems in `ImportHooks` stay in `amm`.

(I'll have time to discuss that again and the other points later.)

alexarchambault:
I'll submit the "Move things around" commits as a separate PRs, although they may not really make sense taken in isolation - the last 3 commits here are the main justification for them.

About the way this does class loader isolation, the idea is to replace the current class loader hierarchy:
```
 .------------------------.
 |        - CL -          |
 | amm                    |
 | + its dependencies     |
 .________________________.
             |
             |
             V
.--------------------------.
|  SpecialClassLoader      |
|                          |
|  user code...            |
|                          |
|                          |
.__________________________.
             |
             |
             V
.--------------------------.
|  SpecialClassLoader      |
|                          |
|  user code...            |
|                          |
|                          |
.__________________________.
             |
             |
             V
            ...
```
with something like:
```

.------------------------.
|        - CL 1 -        |
| amm-api                |
| amm-tools              |
| + their dependencies   |
.________________________.
        |          '--------------------------.
        |                                     |
        V                                     V
.--------------------------.     .--------------------------.
|        - CL 2 -          |     |  SpecialClassLoader      |
| amm                      |     |                          |
| + its dependencies       |     |  user code...            |
| (but all that's already  |     |                          |
|  in CL 1)                |     |                          |
.__________________________.     .__________________________.
                                              |
                                              |
                                              V
                                 .--------------------------.
                                 |  SpecialClassLoader      |
                                 |                          |
                                 |  user code...            |
                                 |                          |
                                 |                          |
                                 .__________________________.
                                              |
                                              |
                                              V
                                             ...
```

Shared things like `ReplAPI`, `ImportHook`, `Evaluator`, ..., are defined in `amm-api`. 

Some of these things are just traits / interfaces (like `ImportHook` or `Evaluator`). Implementations of those are in `amm` loaded by `CL 2` (like the `Evaluator` returned by `Evaluator.apply`). These instances can be used from the Ammonite session as they are instances of classes from `CL 1` too.

lihaoyi:
Ok interesting.

I'd say it's probably (?) not worth sending a PR for the splitting-up yet, mainly because I'm not convinced that this is the best way to implement isolation of user code from Ammonite dependencies. But that's open for discussion.

Honestly, the integration with coursier and it's getIsolationTargets seems pretty special-cased and fragile. What is the advantage of doing it this way, over e.g. using jarjar to auto-vendor the packages we do not want to expose? That will work in this case and others without needing such a tight coupling with the enclosing runtime environment

alexarchambault:
About splitting or not (and fragileness), not splitting, and then vendoring a few packages, is the most fragile way to go. You have to be sure that what is vendored is only used internally and does not leak in any way via non vendored classes. Else that would break using these non vendored classes from the repl.

You could split `amm` (into `amm` and `amm-api` say), and then vendor what's in `amm` (and the dependencies it doesn't share with `amm-api`). That's fragile too, as before, as if you don't vendor a dependency 1 of `amm` and do vendor a dependency 2 of the former, dependency 1 would be broken from the repl too.

Only a careful handling of class loaders and dependencies prevents this (like coursier automatically does, above or in my [ammonium fork](https://github.com/alexarchambault/ammonium), or like you could do too).

About the use of coursier above, it's only because I've been too lazy to change the assembly generated by Ammonite. It can also generate a standalone launcher that handles the right class loader setup, with its `bootstrap` command. Something like
```
coursier bootstrap \
  -i ammonite \
  -I ammonite:com.lihaoyi:ammonite-tools_$SCALA_VERSION:$AMMONITE_VERSION \
  com.lihaoyi:ammonite_$SCALA_VERSION:$AMMONITE_VERSION
```
should work in the next version of it (for now it requires arguments like `-M ammonite.Main -d "\${user.home}/.ammonite/jars"` too, which are noise, and needs extra things for a really standalone jar, for reasons that are too long to explain here).

lihaoyi:
Hey @alexarchambault, sorry for taking long to reply.

Looking at the current state of the project, I don't think it'll be worth merging this change right now. My main considerations are:

- I don't think the benefit (fixing #292) sufficiently outweighs the cost of complexity to the codebase (in this diff, and the further classloader-work we'll need to do to make it work in all cases). In particular, I honestly haven't heard any other complaints since @bblfish's one report long ago, which leads me to de-prioritize it as a not-that-important issue

- I'm planning on ripping apart the `amm` package, but along different lines from what this diff does (interactive, compilation, runtime subsystems) and this split into `amm`/`amm-api` will get in the way of that 

I'm sorry as it's clear you put in a lot of work into this, but that's my decision for now. Happy to continue the conversation if you think there's more discussion worth having, or if there are more compelling use cases that solving #292 would allow that I'm not seeing right now

PR #413, opened 2016-07-20
Make all amm and integration pass on windows except the Syntax Error ones

coderabhishek:
Major Changes:
1) To make the extended ascii character things defined in `scalaz` and similar other problems with `shapeless` we need to set encoding to `utf-8` on windows, for which I have set environment var on appveyor     `JAVA_TOOL_OPTIONS: -Dfile.encoding=UTF8`

2) There are no other major changes except one liners like 

```
private def jarSignature(url: URL) = {
-    val path = Path(url.getFile, root)
+    val path = Path(file.Paths.get(url.toURI()).toFile(), root)
```

3) Few ivy tests are behaving unusual on appveyor but thats just caching problem so I have made `appveyor.yml` itself a dependency for `.ivy2` cache so any major change in appveyor will redownload ivy. Though I am not sure if this change should merge into master.
4) I have introduced a `sanitizePath` function in `Util.scala` to remove windows or linux restricted chars from file names and replace them by `$foo`.
5) Rest is just uncommenting `if(!Util.windowsPlatform)` guard etc.

### DIFF
--- a/amm/src/main/scala/ammonite/interp/ClassLoaders.scala
+++ b/amm/src/main/scala/ammonite/interp/ClassLoaders.scala
@@ -118,7 +118,7 @@ class SpecialClassLoader(parent: ClassLoader, parentSignature: Seq[(Path, Long)]
   }
 
   private def jarSignature(url: URL) = {
-    val path = Path(url.getFile, root)
+    val path = Path(file.Paths.get(url.toURI()).toFile(), root)
     path -> path.mtime.toMillis
   }
 
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,6 +1,11 @@
 # Adapted from https://github.com/sbt/sbt-native-packager/blob/master/appveyor.yml
 version: '{build}'
 os: Windows Server 2012
+
+#  To make the extended ascii characters work on Windows
+#  we need to set string encoding for jvm to utf-8 on windows
+environment:
+        JAVA_TOOL_OPTIONS: -Dfile.encoding=UTF8
 install:
   - ps: |
       Add-Type -AssemblyName System.IO.Compression.FileSystem
@@ -25,5 +30,5 @@ test_script:
 #  - sbt ++2.10.6 ops/test
 cache:
   - C:\sbt\
-  - C:\Users\appveyor\.ivy2
+  - C:\Users\appveyor\.ivy2 -> appveyor.yml
   - C:\Users\appveyor\.sbt
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -30,5 +30,6 @@ test_script:
 #  - sbt ++2.10.6 ops/test
 cache:
   - C:\sbt\
+  # Just to make sure ivy cache is updated 'cause otherwise few tests misbehave
   - C:\Users\appveyor\.ivy2 -> appveyor.yml
   - C:\Users\appveyor\.sbt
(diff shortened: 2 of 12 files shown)


amm/src/main/scala/ammonite/interp/ClassLoaders.scala
```diff
@@ -118,7 +118,7 @@ class SpecialClassLoader(parent: ClassLoader, parentSignature: Seq[(Path, Long)]
   }
 
   private def jarSignature(url: URL) = {
-    val path = Path(url.getFile, root)
```
lihaoyi: What is the purpose of this change? You obviously thought it was interesting enough to mention in the description but you didn't actually say why it was necessary

amm/src/main/scala/ammonite/util/Util.scala
```diff
@@ -58,6 +58,19 @@ object Util{
   //irrespective of the OS on which script was written
   def normalizeNewlines(s: String) = s.replace("\r", "").replace("\n", newLine)
 
+  def sanitizePath(s: String) =
```
lihaoyi: Can we implement this using `NameTransformer.encode` or `decode`? We use those in a few places like https://github.com/lihaoyi/Ammonite/blob/9e249f6d5fd97c56121f11ca77b59071e4a72e25/amm/src/main/scala/ammonite/util/Model.scala#L117, and it's the way the Scala compiler currently does it, so if possible it would nice to be consistent with them

integration/src/test/scala/ammonite/integration/ErrorTruncationTests.scala
```diff
@@ -18,7 +18,7 @@ object ErrorTruncationTests extends TestSuite{
     val e = fansi.Str(
       intercept[ShelloutException]{ exec(file) }.result.err.string
     ).plainText
-    assert(e == expected)
```
lihaoyi: Why is this change necessary? The goal of this test suite is to enforce that these outputs are short and do not contain unnecessary stack traces, and this change would invalidate that. If you want to make it less strict on windows, gate your change with a `Util.windowsPlatform` to allow it to pass less-strictly on windows while still being strict on linux

appveyor.yml
```diff
@@ -1,6 +1,8 @@
 # Adapted from https://github.com/sbt/sbt-native-packager/blob/master/appveyor.yml
 version: '{build}'
 os: Windows Server 2012
+environment:
+        JAVA_TOOL_OPTIONS: -Dfile.encoding=UTF8
```
lihaoyi: This deserves a comment as to why it's necessary

lihaoyi:
That is all for now!

integration/src/test/scala/ammonite/integration/ErrorTruncationTests.scala
lihaoyi: Also, when you do make this change, leave a comment explaining why it is necessary. This sort of special-casing is entirely non-obvious and when you come back a year later it's easy to not know why it was done

amm/src/main/scala/ammonite/interp/ClassLoaders.scala
coderabhishek: Actually `url.getFile` doesn't converts urls back to OS specific path.
http://stackoverflow.com/questions/6164448/convert-url-to-normal-windows-filename-java

Also `url.getFile` is now deprecated

integration/src/test/scala/ammonite/integration/ErrorTruncationTests.scala
coderabhishek: I think the new change should be fine. Is it good enough?

lihaoyi: ok

amm/src/main/scala/ammonite/interp/ClassLoaders.scala
lihaoyi: ok

lihaoyi:
Looks good to me!

Issue #418, opened 2016-07-28
amm fails if ran from /var or /bin


13h3r:
```
alexey@Alexeys-MacBook-Pro /var » cat > /Users/alexey/temp/C.sc                                                                    1 ↵
println("It works")
alexey@Alexeys-MacBook-Pro /var » amm /Users/alexey/temp/C.sc
It works
Exception in thread "main" java.lang.ClassNotFoundException: $file.`..`.`..`.Users.alexey.temp.C$$routes$
    at java.net.URLClassLoader.findClass(URLClassLoader.java:381)
    at ammonite.interp.SpecialClassLoader.findClass(ClassLoaders.scala:113)
    at java.lang.ClassLoader.loadClass(ClassLoader.java:424)
    at java.lang.ClassLoader.loadClass(ClassLoader.java:357)
    at ammonite.main.Scripts$$anonfun$runScript$2.apply(Scripts.scala:52)
    at ammonite.main.Scripts$$anonfun$runScript$2.apply(Scripts.scala:22)
    at ammonite.util.Res$Success.map(Res.scala:61)
    at ammonite.main.Scripts$.runScript(Scripts.scala:22)
    at ammonite.Main.runScript(Main.scala:82)
    at ammonite.Main$$anonfun$main$1$$anonfun$apply$1.apply(Main.scala:193)
    at ammonite.Main$.ammonite$Main$$ifContinually$1(Main.scala:173)
    at ammonite.Main$$anonfun$main$1.apply(Main.scala:175)
    at ammonite.Main$$anonfun$main$1.apply(Main.scala:175)
    at scala.Option.foreach(Option.scala:257)
    at ammonite.Main$.main(Main.scala:175)
    at ammonite.Main.main(Main.scala)
```

lihaoyi:
Should be fixed by https://github.com/lihaoyi/Ammonite/pull/420

PR #420, opened 2016-07-30
Make sure scripts with symbols in path name work

coderabhishek:
This PR is intended as a patch for https://github.com/lihaoyi/Ammonite/issues/418.
Before this PR any script having special chars in path name fails because while loadingClasses wrapperNames are encoded whereas while findClass is given non-encoded Name as argument. 
This PR fixes this issue.

### DIFF
--- a/integration/src/test/scala/ammonite/integration/BasicTests.scala
+++ b/integration/src/test/scala/ammonite/integration/BasicTests.scala
@@ -25,6 +25,33 @@ object BasicTests extends TestSuite{
       assert(evaled.out.trim == "Hello World")
     }
 
+    //make sure scripts with symbols in path names work fine
+    'scriptWithSymbols {
+      if (!Util.windowsPlatform){
+        val dirAddr =
+          cwd/'target/'test/'resources/'ammonite/'integration/'basic
+        val weirdScriptName = "script%#.@*+叉燒.sc"
+        val scriptAddr = dirAddr/weirdScriptName
+        rm(scriptAddr)
+        write(scriptAddr, """println("Script Worked!!")""")
+        val evaled = %%bash(
+          executable,
+          scriptAddr
+          )
+        assert(evaled.out.trim == "Script Worked!!" && evaled.err.string.isEmpty)
+      }
+    }
+
+    'scriptInSomeOtherDir{
+      val scriptAddr = tmp.dir()/"script.sc"
+      rm(scriptAddr)
+      write(scriptAddr, """println("Worked!!")""")
+      val evaled = %% bash(
+        executable,
+        scriptAddr
+        )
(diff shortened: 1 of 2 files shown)


i..name.sc
```diff
@@ -0,0 +1 @@
+println("Done")
```
lihaoyi: This file looks un-used; can we kill it?

integration/src/test/resources/ammonite/integration/basic/script..run.sc
```diff
@@ -0,0 +1 @@
+println("Script Worked!!")
```
lihaoyi: If we're going to test encoding of symbols, we might as well go the whole way and name the file `script%#.@*+叉燒.sc` and make sure all that gets properly handled

lihaoyi:
I have verified that this fixes the problem, but could you add a unit test that replicates the original problem? i.e. run a script that lives outside of the working directory of the shell used to start the script? Ammonite-Ops lets you pass in a working directory when it runs external processes via `%%`.

integration/src/test/scala/ammonite/integration/BasicTests.scala
```diff
@@ -25,6 +25,27 @@ object BasicTests extends TestSuite{
...
+          cwd/'integration/'src/'test/'resources/'ammonite/'integration/'basic/"script%#.@*+叉燒.sc"
+        rm(scriptAddr)
+        write(scriptAddr, """println("Script Worked!!")""")
+        val evaled = exec('basic / "script%#.@*+叉燒.sc")
+        assert(evaled.out.trim == "Script Worked!!" && evaled.err.string.isEmpty)
+      }
+    }
+
+    'scriptInSomeOtherDir{
+      rm(home/'bin/"script.sc")
```
lihaoyi: Can use use a a `tmp.dir` instead of people's `home`?

lihaoyi:
Just one tweak; when you're done squash it and I'll merge it

integration/src/test/scala/ammonite/integration/BasicTests.scala
```diff
@@ -25,6 +25,27 @@ object BasicTests extends TestSuite{
       assert(evaled.out.trim == "Hello World")
     }
 
+    //make sure scripts with symbols in path names work fine
+    'scriptWithSymbols {
+      if (!Util.windowsPlatform){
+        val scriptAddr =
+          cwd/'integration/'target/'test/'resources/'ammonite/'integration/'basic/"script%#.@*+叉燒.sc"
+        rm(scriptAddr)
```
coderabhishek: I have changed scriptAddr from src to target
I think this one should be better.

lihaoyi: Sounds good

PR #422, opened 2016-08-01
Support for custom java options via JAVA_OPTS env var

### LINKED ISSUE #419: Custom java options (JAVA_OPTS env var support)
As a Finagle user, I need to pass some java options to the framework, but the `amm` command (which is a bash wrapper that invokes the JVM) does not seem to accept a way to pass custom java options to the JVM.

Historically one would have a JAVA_OPTS environmental variable containing all the Java options, and the bash wrapper would add this string as an argument when the JVM is invoked.

I believe Ammonite should support JAVA_OPTS or an equivalent mechanism.

sscarduzio:
Fixes #419


lihaoyi:
Looks good, thanks!

Issue #417, opened 2016-07-28
$file import fails for scripts in nested folders if you leave out the `.sc` when running script


lihaoyi:
```scala
haoyi-mbp:~ haoyi$ amm A
Exception in thread "main" java.lang.StringIndexOutOfBoundsException: String index out of range: 0
	at java.lang.String.charAt(String.java:646)
	at ammonite.util.Name.<init>(Model.scala:120)
	at ammonite.util.Util$.pathToPackageWrapper(Util.scala:23)
	at ammonite.main.Scripts$.runScript(Scripts.scala:20)
	at ammonite.Main.runScript(Main.scala:82)
	at ammonite.Main$$anonfun$main$1$$anonfun$apply$1.apply(Main.scala:193)
	at ammonite.Main$.ammonite$Main$$ifContinually$1(Main.scala:173)
	at ammonite.Main$$anonfun$main$1.apply(Main.scala:175)
	at ammonite.Main$$anonfun$main$1.apply(Main.scala:175)
	at scala.Option.foreach(Option.scala:257)
	at ammonite.Main$.main(Main.scala:175)
	at ammonite.Main.main(Main.scala)
```

A file `A` doesn't need to exist, nor does `A.sc`. Nevertheless, we should fail with a better error message

lihaoyi:
Fixed by #425

PR #425, opened 2016-08-03
Make scripts without extension throw correct exceptions

coderabhishek:
This patch is a fix for https://github.com/lihaoyi/Ammonite/issues/417.
It avoids the case where `pathToPackage` throws unwanted exception when it can not find '.' in file name.


lihaoyi:
Looks good!

Could you also update the [unstable changelog](http://www.lihaoyi.com/Ammonite/#UnstableChanglog) as part of this diff? I'd like to get us in a habit of doing it so that when the time comes to publish a stable release, we can just C&P it over to the normal changelog and be done with it

lihaoyi:
Green enough for me, thanks!

PR #435, opened 2016-08-12
Add regression tests for removal of scalac/fastparse dependency for c…

coderabhishek:
…ached scripts

### DIFF
--- a/integration/src/test/scala/ammonite/integration/BasicTests.scala
+++ b/integration/src/test/scala/ammonite/integration/BasicTests.scala
@@ -20,6 +20,13 @@ object BasicTests extends TestSuite{
   println("StandaloneTests")
   val tests = TestSuite {
 
+    def execWithJAVA_OPTSset(name: RelPath, home: Path) = %%bash(
+      executable,
+      replStandaloneResources/name,
+      "-h",
+      home,
+      JAVA_OPTS = "-verbose:class"
+      )
     'hello{
       val evaled = exec('basic/"Hello.sc")
       assert(evaled.out.trim == "Hello World")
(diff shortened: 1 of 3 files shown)


integration/src/test/scala/ammonite/integration/BasicTests.scala
```diff
@@ -41,6 +41,46 @@ object BasicTests extends TestSuite{
...
+      assert(evaled1.out.trim.contains("fastparse"))
+
+      val evaled2 = %%bash(
+        executable,
+        replStandaloneResources/'basic/"Print.sc",
+        "-h",
+        "target/tempAmmoniteHome",
+        JAVA_OPTS = "-verbose:class"
+        )
+      assert(!evaled2.out.trim.contains("fasparse"))
```
lihaoyi: There's a typo here and it will always return true

integration/src/test/scala/ammonite/integration/BasicTests.scala
```diff
@@ -41,6 +41,46 @@ object BasicTests extends TestSuite{
...
+        executable,
+        replStandaloneResources/'basic/"Print.sc",
+        "-h",
+        "target/tempAmmoniteHome",
+        JAVA_OPTS = "-verbose:class"
+      )
+      rm(cwd/'target/'tempAmmoniteHome)
+      val count1 = substrCount(evaled1.out.trim, "scala.tools.nsc")
+      val count2 = substrCount(evaled2.out.trim, "scala.tools.nsc")
+      assert(count1 > count2)
```
lihaoyi: Can we add some concrete limit to the number of classes? e.g. `count1 > 100 && count2 < 10` or something similar

lihaoyi:
Looks great, thanks!

PR #440, opened 2016-08-19
Add interp.args field for storing script arguments

benwbooth:
Stores the unmodified script arguments to InterpAPI.args field. I tested it and it works:

```
#!/usr/bin/env amm
println(interp.args)
```

```
$ ./test.sc 1 2 3
Vector(1, 2, 3)
```

The only thing I'm not sure about is if args really contains all the arguments or just the non-keyword positional arguments. It seems like it should be right, since `passThroughArgs` from the `main` method is being passed into `runScript` as `args` which is what I'm using. Does `passThroughArgs` contain all the unprocessed arguments?


lihaoyi:
> Does `passThroughArgs` contain all the unprocessed arguments?

You should test it and tell me 😛

lihaoyi:
@benwbooth This also needs a unit or integration test if it wants to land

lihaoyi:
Closing due to inactivity

Issue #433, opened 2016-08-10
Don't cache script dependencies with version *-SNAPSHOT


yakticus:
When using scripting, I would like to be able to do this:
```
import $ivy.`com.mycompany::somelib:0.1-SNAPSHOT`
```
…where `somelib` comes from my local maven/ivy repo (published as a snapshot from my local dev environment). Ammonite caches `somelib` version `0.1-SNAPSHOT` under `~/.ivy2/cache` and thus if I re-build somelib, the changes are not picked up by scripts importing it.

The proposed change is to add an exception so that versions ending in `-SNAPSHOT` are not cached. Documentation should be updated to mention this and the fact that using `*-SNAPSHOT` versions could slow down loading.

Ammonite has its own cache of what was resolved, under `~/.ammonite/cache/0.7.4/ivycache.json`, which can be found in the code here:

https://github.com/lihaoyi/Ammonite/blob/master/amm/runtime/src/main/scala/ammonite/runtime/Storage.scala#L238

The trick would be to make sure that for `*-SNAPSHOT` versions, that the resolver cache points to `~/.ivy2/local` instead of `~/.ivy2/cache`

yakticus:
merged https://github.com/lihaoyi/Ammonite/pull/451

PR #451, opened 2016-08-29
fix for #433 - ivy pulls in changes to SNAPSHOT

yakticus:
See commit log for details. 

NOTE: I removed cache ivy resolvers because they don't seem to do anything.

Q: Should the docs be updated for this issue?

### DIFF
--- a/integration/src/test/scala/ammonite/integration/BasicTests.scala
+++ b/integration/src/test/scala/ammonite/integration/BasicTests.scala
@@ -138,6 +138,33 @@ object BasicTests extends TestSuite{
 
       assert(errorMsg.contains("IvyThing$IvyResolutionException"))
     }
+    'testIvySnapshotNoCache{
+
+      val dummyScala = intTestResources/"some-dummy-library"/'src/'main/'scala/'dummy/"Dummy.scala"
+
+      def publishJarAndRunScript(theThing: String) = {
+        // 1. edit code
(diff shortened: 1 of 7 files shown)


integration/src/test/scala/ammonite/integration/BasicTests.scala
```diff
@@ -138,6 +138,33 @@ object BasicTests extends TestSuite{
 
       assert(errorMsg.contains("IvyThing$IvyResolutionException"))
     }
+    'testIvySnapshotNoCache{
+
+      val dummyScala = intTestResources/"some-dummy-library"/'src/'main/'scala/'dummy/"Dummy.scala"
```
lihaoyi: Could we move this into `integration/target`? That will make sure we don't accidentally check it in since it'll be gitignored

lihaoyi:
One nit, and the CI build seems to be complaining about your lines being too long

lihaoyi:
@yakticus seems to have conflicted with #456 since I merged it; could you squash it onto master?

yakticus:
the failing test, `ammonite.session.ProjectTests.load.ivy.resolvers` passes on my machine

lihaoyi:
That resolvers failure is probably just a flake. Tested the snapshot/non-snapshot workflows locally manually and they behave as expected. Thanks @yakticus!

PR #452, opened 2016-08-29
Make output silent with -s flag

coderabhishek:
Passing -s flag will make program output free from any noisy string not printed by the program itself eg. "Compiling ScriptName" printed when script is compiled instead of using from cache.

```
abhishek@abhishek-Inspiron-3542:~/Ammonite$ ./a i.sc
Compiling Main.sc
Compiling Main.sc
Compiling Main.sc
Compiling Main.sc
Compiling Main.sc
Compiling i.sc
Hello

abhishek@abhishek-Inspiron-3542:~/Ammonite$ 
```

```
abhishek@abhishek-Inspiron-3542:~/Ammonite$ rm ~/.ammonite/cache/ -rf
abhishek@abhishek-Inspiron-3542:~/Ammonite$ ./a i.sc -s
Hello

```

Besides this -s also makes ivy silent

### DIFF
--- a/amm/interp/src/main/scala/ammonite/runtime/Interpreter.scala
+++ b/amm/interp/src/main/scala/ammonite/runtime/Interpreter.scala
@@ -33,7 +33,7 @@ class Interpreter(val printer: Printer,
                   // the REPL before it starts
                   extraBridges: Interpreter => Seq[(String, String, AnyRef)],
                   val wd: Path,
-                  verboseIvy: Boolean = true)
+                  verboseOutput: Boolean = true)
   extends ImportHook.InterpreterInterface{ interp =>
 
 
@@ -365,6 +365,13 @@ class Interpreter(val printer: Printer,
       tag
     ) match {
       case None =>
+        (source, verboseOutput) match {
+          case (ImportHook.Source.File(fName), true) =>
+            printer.out("Compiling " + fName.last + "\n")
+          case (ImportHook.Source.URL(url), true) => 
+            printer.out("Compiling " + url + "\n")
+          case _ =>
+        }
         init()
         val res = processModule0(
           source, code, wrapperName, pkgName,
@@ -581,7 +588,7 @@ class Interpreter(val printer: Printer,
         val resolved = ammonite.runtime.tools.IvyThing(
           () => interpApi.resolvers(),
           printer,
-          verboseIvy
+          verboseOutput
         ).resolveArtifact(
           groupId,
           artifactId,
@@ -605,7 +612,7 @@ class Interpreter(val printer: Printer,
     lazy val ivyThing = ammonite.runtime.tools.IvyThing(
       () => interpApi.resolvers(),
       printer,
-      verboseIvy
+      verboseOutput
     )
 
     def handleClasspath(jar: File): Unit
@@ -708,7 +715,7 @@ object Interpreter{
     Name(wrapperName.raw + (if (wrapperIndex == 1) "" else "_" + wrapperIndex))
   }
 
-  def initPrinters(output: OutputStream, error: OutputStream) = {
+  def initPrinters(output: OutputStream, error: OutputStream, verboseOutput: Boolean) = {
     val colors = Ref[Colors](Colors.Default)
     val printStream = new PrintStream(output, true)
     val errorPrintStream = new PrintStream(error, true)
--- a/integration/src/test/scala/ammonite/integration/BasicTests.scala
+++ b/integration/src/test/scala/ammonite/integration/BasicTests.scala
@@ -29,7 +29,7 @@ object BasicTests extends TestSuite{
       JAVA_OPTS = "-verbose:class"
       )
     'hello{
-      val evaled = exec('basic/"Hello.sc")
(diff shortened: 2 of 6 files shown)


amm/interp/src/main/scala/ammonite/runtime/Interpreter.scala
```diff
@@ -359,6 +359,7 @@ class Interpreter(val printer: Printer,
       tag
     ) match {
       case None =>
+        printer.out("Compiling Script\n")
```
lihaoyi: Could we make this print out the name of the script that's being compiled?

amm/interp/src/main/scala/ammonite/runtime/Interpreter.scala
```diff
@@ -699,9 +700,17 @@ object Interpreter{
...
+  def initPrinters(output: OutputStream, error: OutputStream, verboseOutput: Boolean) = {
     val colors = Ref[Colors](Colors.Default)
-    val printStream = new PrintStream(output, true)
+    val printStream = verboseOutput match{
+      case true => new PrintStream(output, true)
+      //make a dummy printstream in case of non-verbose output
+      case false => new PrintStream(new OutputStream(){
+        @Override
+        def write(b: Int){
+        }
```
lihaoyi: Skip the `@Override` annotation (not necessary in Scala) and let's define write as `= ()` since the `(b: Int){}` syntax is getting deprecated

integration/src/test/scala/ammonite/integration/BasicTests.scala
```diff
@@ -30,7 +30,7 @@ object BasicTests extends TestSuite{
       )
     'hello{
       val evaled = exec('basic/"Hello.sc")
-      assert(evaled.out.trim == "Hello World")
+      assert(evaled.out.trim.replace("Compiling Script\n", "") == "Hello World")
```
lihaoyi: For this, and the other tests that are broken, can we pass in `-s` rather than doing a string-replace check?

lihaoyi:
Some nits but looks good. Tests are still failing

amm/interp/src/main/scala/ammonite/runtime/Interpreter.scala
```diff
@@ -699,9 +700,15 @@ object Interpreter{
...
+  def initPrinters(output: OutputStream, error: OutputStream, verboseOutput: Boolean) = {
     val colors = Ref[Colors](Colors.Default)
-    val printStream = new PrintStream(output, true)
+    val printStream = verboseOutput match{
+      case true => new PrintStream(output, true)
+      //make a dummy printstream in case of non-verbose output
+      case false => new PrintStream(new OutputStream(){
+        def write(i: Int) = ()
+      })
+    }
```
lihaoyi: @coderabhishek by swapping out the entire output stream with a dummy in the case of `-s`, what else are you affecting apart from the `Compiling` message??

amm/interp/src/main/scala/ammonite/runtime/Interpreter.scala
```diff
@@ -359,6 +359,7 @@ class Interpreter(val printer: Printer,
       tag
     ) match {
       case None =>
+        printer.out("Compiling " + wrapperName.backticked + "\n")
```
lihaoyi: Let's pull the name of the file out of the `source` and use that, including the `.scala`.

amm/interp/src/main/scala/ammonite/runtime/Interpreter.scala
coderabhishek: It should make any printer.out thing go silent. Which will not affect errors and warnings but make anything printed by our program silent.
Though printlns will still be printed but I think we should not use them for printing.

lihaoyi: Rather than rewiring `printer.out`, could we instead put an if-statement around the `Compiling...` statement? This would cause some funny behavior when in the REPL (e.g. stuff wouldn't get echoed back properly after running a statement) whereas if we put the if around `Compiling...`, the only change would be that we don't see `Compiling...` when loading a script from the Ammonite REPL, which is probably less surprising

coderabhishek: ok but that'd require such a guard for anything being printed in future. I'll change it that way.

lihaoyi: Yeah that's probably fine; I don't think a blanket "shut off all stdout" is something people would expect, so instead having it "shut off some subset of auxiliary output we deem unimportant" might be easier to understand. Then we can turn it on case by case without worrying about weird interactions

lihaoyi:
@coderabhishek can you squash or rebase this? For some reason you keep getting existing commits in the PR history which makes it impossible to revew

lihaoyi:
also, a lot of your tests are failing

coderabhishek:
Yeah I will fix the PR. This was not final. I just pushed to check. I will msg you on gitter once it is done.

Issue #448, opened 2016-08-26
deprecate `cwd` and rename it to `pwd`


yakticus:
Rename cwd -> pwd so that it is more consistent w/ bash and less confusing with `wd`

yakticus:
merged: https://github.com/lihaoyi/Ammonite/pull/456

PR #456, opened 2016-08-31
#448 fix - deprecated cwd and added pwd. find/replace cwd -> pwd

yakticus:


lihaoyi:
This is failing because the `build.sc` file is using a hardcoded, old version of Ammonite to run itself. Thus you should leave those as `cwd` for now and after we publish another stable version we can go swap them over

lihaoyi:
Failures are spurious, this looks good, thanks @yakticus 
!

PR #457, opened 2016-09-01
support multiple --predef-file options

tdyas:
Supporting multiple --predef-file options makes it easier to have a shared project-specific predef and a personal predef for a Scala project.

Note: I matched the existing behavior of `ammonite.runtime.Storage.Folder.loadPredef` to silently ignore file not found errors. I'm skeptical of that behavior, but it is preexisting so will match it.


tdyas:
Failure is do a rate limit on some API being exceeded.

lihaoyi:
Failures look like flakes. Looks good, thanks @tdyas!

PR #464, opened 2016-09-07
Fix #432 (Can't cd into directories' symlinks)

### LINKED ISSUE #432: Cannot chdir into symlinked directory
On Mac OSX, amm 0.7 shell cannot execute `cd! 'abcd` if `abcd` is a symlink to a directory.

To reproduce:
```
$ amm
@ mkdir! 'test
@ %ln ("-s", "test", "test1")
@ cd! 'test1
```

Expected behavior: `cd!` should be successful. The symlinked directory `test1` exists and is accessible to the user. The behavior of `cd! 'test1` should be the same as in `bash`.

Actual result: Exception, current directory remains unchanged.
```
java.nio.file.NotDirectoryException: /private/tmp/test1
  ammonite.shell.ShellSession$$anonfun$1.apply(ShellSession.scala:22)
  ammonite.shell.ShellSession$$anonfun$1.apply(ShellSession.scala:21)
  ammonite.ops.Callable1.$bang(Extensions.scala:42)
  $sess.cmd104$.<init>(cmd104.sc:1)
  $sess.cmd104$.<clinit>(cmd104.sc:1)
```

On Mac OS X, `/tmp` is a symlink, so this problem prevents me from doing `cd! root/'tmp` on Mac OS X.

Another point is that the built-in `ln` cannot seem to produce symlinks, only hard links (is that right?). That's why I am calling the external `ln` in this example shown above.

tkroman:
This is a proposal to fix #432 via the mechanism of resolving 'real paths' + following symlinks (as described by the `j.n.f.Path`'s javadoc (https://docs.oracle.com/javase/8/docs/api/java/nio/file/Path.html#toRealPath-java.nio.file.LinkOption...-)). 

Also fixes the bug which caused the `stat.full` on the `root` variable to fail.

Test included.

### DIFF
--- a/shell/src/main/scala/ammonite/shell/ShellSession.scala
+++ b/shell/src/main/scala/ammonite/shell/ShellSession.scala
@@ -5,27 +5,37 @@ import java.nio.file.attribute.PosixFilePermission
 
 import ammonite.ops._
 import ammonite.repl.FrontEndUtils
-import pprint.{Config, PPrinter, PPrint}
+import pprint.{Config, PPrinter}
+
+import scala.util.Try
 
 case class ShellSession() extends OpsAPI {
   var wd0 = pwd
   /**
-   * The current working directory of the shell, that will get picked up by
-   * any ammonite.ops commands you use
-   */
+    * The current working directory of the shell, that will get picked up by
+    * any ammonite.ops commands you use
+    */
   implicit def wd = wd0
   /**
-   * Change the working directory `wd`; if the provided path is relative it
-   * gets appended on to the current `wd`, if it's absolute it replaces.
-   */
+    * Change the working directory `wd`; if the provided path is relative it
+    * gets appended on to the current `wd`, if it's absolute it replaces.
+    */
   val cd = (arg: Path) => {
-    if (!stat(arg).isDir) throw new NotDirectoryException(arg.toString)
-    else {
-      wd0 = arg
-      wd0
-    }
+    /*
+     * `arg` should be a directory or a symlink to a directory
+     * `realPath will be None if that is not the case,
+     * otherwise - Some(realPath)
+     */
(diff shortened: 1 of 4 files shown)


lihaoyi:
The test failures look like flakes; this should be fine

shell/src/main/scala/ammonite/shell/ShellSession.scala
```diff
@@ -5,27 +5,37 @@ import java.nio.file.attribute.PosixFilePermission
...
+     * `realPath will be None if that is not the case,
+     * otherwise - Some(realPath)
+     */
+    val realPath = Option(arg)
+      .filter(_.isDir)
+      .orElse(Try(Path(arg.toNIO.toRealPath())).toOption.filter(_.isDir))
 
+    realPath match {
+      case None => throw new NotDirectoryException(arg.toString)
+      case Some(path) => wd0 = path; wd0
```
winitzki: I think instead of `wd0 = path` this should be `wd0 = arg` because then you can actually change into a symlinked directory and not into the directory the symlink points to.

As it is now, `tmp` is symlinked on a Mac OS X, and you can't do something like this:
```
cd! root/'tmp
ls! up/home // "up" does not point to root
```

tkroman: I thought about that, but decided to go with `path` as `wd` because the consequences of using a symlink as a `wd` were unclear to me. This way we ensure that at least the `wd` state is consistent with the link following process. But if you are sure that allowing `wd` to be a symlink doesn't break anything, it would be nice to have a consistent `cd` (as the shells do).

winitzki: Yes, I'll make a new PR, and move the discussion there.

PR #466, opened 2016-09-08
[Issue 432] Follow links when executing `cd`

winitzki:
- Adding a method `Path#followLinks` that follows symlinks transitively (if the Path is a symlink).

- Fixing `cd` so that you can go to a symlinked directory.

This fixes issue https://github.com/lihaoyi/Ammonite/issues/432


winitzki:
The test failure is only in Scala 2.10.5 + openJDK 7 and seems to be unrelated to my changes...

lihaoyi:
Test failure is just a flake; this looks good to me I think

lihaoyi:
Ended up preferring the implementation from #464, which should also solve this problem I think

PR #469, opened 2016-09-08
add "Path#tryFollowLinks"; "cd" will now change to symlinked directory

winitzki:
I think `cd! symlinkedDir` should actually change into a symlinked directory and not into the directory the symlink points to.

As it is now, `tmp` is symlinked on a Mac OS X, and you can't do something like this:
```
cd! root/'tmp
ls! up/'home // "up" does not point to root
```

This PR makes `cd` change to the symlinked directory (if it exists).

This PR also adds `Path#tryFollowLinks : Option[Path]`, which will return the link target (if it exists).

### DIFF
--- a/ops/src/test/scala/test/ammonite/ops/PathTests.scala
+++ b/ops/src/test/scala/test/ammonite/ops/PathTests.scala
@@ -340,39 +340,37 @@ object PathTests extends TestSuite{
         }
       }
       'symlinks{
+
+        val names = Seq('test123, 'test124, 'test125, 'test126)
+        val twd = tmp.dir()
+
         'nestedSymlinks{
           if(Unix()) {
-            val names = Seq('test123, 'test124, 'test125, 'test126)
-            names.foreach(p => rm ! pwd / p)
-            mkdir ! pwd / 'test123
-            ln.s(pwd / 'test123, pwd / 'test124)
-            ln.s(pwd / 'test124, pwd / 'test125)
-            ln.s(pwd / 'test125, pwd / 'test126)
-
-            assert((pwd / 'test126).tryFollowLinks.get == pwd / 'test123)
-
-            names.foreach(p => rm ! pwd / p)
-            names.foreach(p => assert(!exists(pwd / p)))
+            names.foreach(p => rm ! twd/p)
+            mkdir ! twd/'test123
+            ln.s(twd/'test123, twd/'test124)
+            ln.s(twd/'test124, twd/'test125)
+            ln.s(twd/'test125, twd/'test126)
+            assert((twd/'test126).tryFollowLinks.get == (twd/'test123).tryFollowLinks.get)
+            names.foreach(p => rm ! twd/p)
+            names.foreach(p => assert(!exists(twd/p)))
           }
-
         }
+
         'danglingSymlink{
           if(Unix()) {
-            val names = Seq('test123, 'test124, 'test125, 'test126)
-            names.foreach(p => rm ! pwd / p)
-            mkdir ! pwd / 'test123
-            ln.s(pwd / 'test123, pwd / 'test124)
-            ln.s(pwd / 'test124, pwd / 'test125)
-            ln.s(pwd / 'test125, pwd / 'test126)
-
-            rm ! pwd / 'test123
-
-            assert( (pwd / 'test126).tryFollowLinks.isEmpty)
-
-            names.foreach(p => rm ! pwd / p)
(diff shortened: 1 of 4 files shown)


winitzki:
Comment by @tkroman:

I thought about that, but decided to go with path as wd because the consequences of using a symlink as a wd were unclear to me. This way we ensure that at least the wd state is consistent with the link following process. But if you are sure that allowing wd to be a symlink doesn't break anything, it would be nice to have a consistent cd (as the shells do).

shell/src/main/scala/ammonite/shell/ShellSession.scala
```diff
@@ -28,11 +28,11 @@ case class ShellSession() extends OpsAPI {
      */
     val realPath = Option(arg)
       .filter(_.isDir)
-      .orElse(Try(Path(arg.toNIO.toRealPath())).toOption.filter(_.isDir))
+      .orElse(Try(arg.followLinks).toOption.filter(_.isDir))
```
winitzki: In what circumstances will NIO.toRealPath() throw an exception?

tkroman: The javadoc explains that pretty clearly :)
E.g. if a symlink's target is removed, there will likely be an exception.

shell/src/test/scala/ammonite/shell/SessionTests.scala
```diff
@@ -81,12 +81,55 @@ object SessionTests extends TestSuite{
...
+
+        @ assert(wd == originalWd/'test126)
+
+        @ cd! originalWd
+
+        @ assert(wd == originalWd)
+
+        @ rm! 'test123
+
+        @ assert( (wd / 'test126).tryFollowLinks == None )
```
winitzki: Ideally, here I want to try `cd! wd/'test126` and catch an exception. But I don't know how to specify catching an exception in this test format.

lihaoyi:
This seems reasonable; I don't have time to properly review it yet, but I promise I'll get to it soon. Thanks for your patience!

ops/src/test/scala/test/ammonite/ops/PathTests.scala
```diff
@@ -340,6 +339,42 @@ object PathTests extends TestSuite{
           }
         }
       }
+      'symlinks{
+        'nestedSymlinks{
+          if(Unix()) {
+            val names = Seq('test123, 'test124, 'test125, 'test126)
+            names.foreach(p => rm ! pwd / p)
+            mkdir ! pwd / 'test123
```
lihaoyi: Could this stuff all go in a temp dir?

ops/src/test/scala/test/ammonite/ops/PathTests.scala
```diff
@@ -340,6 +339,42 @@ object PathTests extends TestSuite{
...
+
+            names.foreach(p => rm ! pwd / p)
+            names.foreach(p => assert(!exists(pwd / p)))
+          }
+
+        }
+        'danglingSymlink{
+          if(Unix()) {
+            val names = Seq('test123, 'test124, 'test125, 'test126)
+            names.foreach(p => rm ! pwd / p)
```
lihaoyi: This stuff too; `tmp.dir` to avoid polluting working dir

ops/src/test/scala/test/ammonite/ops/PathTests.scala
```diff
@@ -340,6 +339,42 @@ object PathTests extends TestSuite{
...
+            rm ! pwd / 'test123
+
+            assert( (pwd / 'test126).tryFollowLinks.isEmpty)
+
+            names.foreach(p => rm ! pwd / p)
+            names.foreach(p => assert(!exists(pwd / p)))
+          }
+
+        }
+      }
```
lihaoyi: Could we add some simple tests for `mkdir`, `write`, `read`, `stat`/`stat.full`, `cp`, and `ls` inside a sym-linked directory? I'm 99% they will work, but we should verify

lihaoyi:
Implementation looks good; tests could use a bit of tweaking. Test failures look like flakes. That's all.

winitzki:
When time permits I'll refactor the tests to use temporary directory and to add other tests inside a symlinked directory.

lihaoyi:
Looks great, thanks!

lihaoyi:
This is now available as the unstable version http://www.lihaoyi.com/Ammonite/#UnstableVersions COMMIT-8af3ceb. Thanks @winitzki!

Issue #430, opened 2016-08-08
REPL exits on NPE on tab-complete


nmcb:
Open en type:

```scala
> amm
Welcome to the Ammonite Repl 0.7.0
(Scala 2.11.8 Java 1.8.0_92)
@ object X {
    import y
    def y(z: Int)
```

Then TAB-complete ```import y```, the ammonite REPL will exit with an:

```
Exception in thread "main" java.lang.NullPointerException
def at ammonite.interp.Pressy$Run.prefixed(Pressy.scala:135)
...
```

tkroman:
@lihaoyi this one may be closed now.

lihaoyi:
@tkroman cool, thanks!

PR #475, opened 2016-09-23
Fix + test for #430

tkroman:
#430 describes a case when we ask for completion on an invalid input and `ammonite.runtime.Pressy.Run#prefixed` throws NPE and the whole REPL goes down.

The fix is basically to lift the `prefixed` call into a Try context that is already there.


lihaoyi:
Looks good to me, thanks!

PR #477, opened 2016-09-24
[WIP]Attempt to fix #397

### LINKED ISSUE #397: Misbehaviour when piping into the Ammonite REPL
I noticed this when attempting,
```
# yolo ...
curl -s https://raw.githubusercontent.com/milessabin/shapeless/master/scripts/try-shapeless.sh | bash
```
The REPL is downloaded and launched, however garbage appears on at the REPL prompt.

The following is a smaller reproduction,
```
haoyi-mbp:~ haoyi$ echo "println(1)" | ~/amm
Loading...
Welcome to the Ammonite Repl 0.5.8
(Scala 2.11.8 Java 1.8.0_25)
haoyi-haoyi@ println(1)
1

haoyi-haoyi@ ￿￿￿￿￿￿￿￿￿￿￿￿￿￿￿￿￿￿￿￿￿￿￿￿￿￿￿￿￿￿￿￿￿￿...
```
See also #367.

tkroman:
#397 was about accepting pipe's input.
The problem as I saw it was in the fact that `-1` designating EOF of Java's input streams was accepted by an incorrect filter. 
There was this weird thing, first stating that exit is done by encountering a `-1` in a stream supplying _ints_, then designating a pattern of `(-1).toChar.toString` instead (which was a single string of `0xffff` char since `Char[acter]`s are unsigned), then (@ Filter.scala:48) did `prefix.map(_.toInt)`, which resulted in `0xffff` again, which, of course, didn't match `-1` (int) in `LazyList#dropPrefix`.

I propose an implementation which avoids this casting, which makes an original example work (both on mac OS and Linux, but that's kinda irrelevant since the problem is not in the piping mechanisms after all).

No unit tests because I'm not sure this is the final solution and I'd like to know what you think about this. If/when you give it a green light, I guess a simple test not involving the piping simulation could be added to some `FilterTest` since this isn't really a buffer reading problem.

### DIFF
--- a/terminal/src/main/scala/ammonite/terminal/filters/BasicFilters.scala
+++ b/terminal/src/main/scala/ammonite/terminal/filters/BasicFilters.scala
@@ -100,7 +100,7 @@ object BasicFilters {
       }
     },
     // java.io.Reader.read() produces -1 on EOF
-    action((-1).toChar.toString)(_ => Exit)
+    wrap(ti => ti.ts.inputs.dropPrefix(Seq(-1)).map(_ => Exit))
   )
   def clearFilter: Filter =
     action(Ctrl('l')){case TS(rest, b, c, _) => ClearScreen(TS(rest, b, c))}


tkroman:
On a side note, WDYT about implementing some super-duper-silent-mode that will allow for proper piping? With the fix proposed in this PR, we already can do `echo '1 + 1' | amm`, but there's a lot of 'garbage' printed alongside the computation result (Scala version, 'Bye!' at the end etc.).
Allowing for `echo '1 + 1' | amm -v0`, where `v0` is verbosity level under which only the actual result is printed would make for a nice featur.

lihaoyi:
Sorry I haven't looked at this! Will do so soon

lihaoyi:
This looks great; thanks for the fix! I'm happy with it as is. Adding `println`s in manually seems to validate your theory that the original filter isn't correctly picking up the -1 "end of line" indicator, while your modified version does. The reason it didn't come up with the interactive use case is that `Ctrl-D` had a separate filter that *did* work; this filter was just broken from the start.

I'm personally happy just merging it as-is without a test: I wouldn't mind having an integration test, but a lot of the other terminal-level stuff isn't really tested anyway.

With regard to have a `-v0` mode that didn't have *any* surrounding UI, I think that's a great idea and feel free to send a PR. 

My original intention was that `amm -c "..."` would serve that purpose, but I can imagine the use case where the code you want to run comes from a pipe rather than from some bash variable you can splice into the shell command

PR #478, opened 2016-09-28
add maven central version badge to readme

emanresusername:
because bookmarking http://www.lihaoyi.com/Ammonite/#Changelog didn't scroll to the right place in my browser and then i have to scroll up slightly. __I'M HAVING SERIOUS FIRST WORLD PROBLEMS HERE!!!__


lihaoyi:
Looks good

PR #480, opened 2016-10-04
`mkdir -p` to make instructions work in any order (tiny change)

tekacs:
There's no need to error on failing to create ~/.ammonite if it already exists, such as for those who have already downloaded Ammonite.


lihaoyi:
Looks good, thanks!

PR #486, opened 2016-10-13
Allow configuration of SSHD console classloader

tjdett:
This allows the server to be used in environments with hot-reloading, like the Play Framework development server.


lihaoyi:
Looks good

PR #492, opened 2016-10-28
upgrade org.apache.sshd sshd-core from 0.14 to 1.3.0 to fix openssh errors

axb21:
As of version 7.0, OpenSSH comes with DSA host key authentication [disabled by default](http://www.openssh.com/legacy.html). Ammonite's `sshd` subproject was using `org.apache.sshd` `sshd-core` version 0.14, which by default offered the DSA host key authentication. Thus, sshing to an embedded ammonite shell in a running application would not work with newer versions of the OpenSSH client, failing with an error like `Unable to negotiate with 127.0.0.1 port 2222: no matching host key type found. Their offer: ssh-dss`. You can replicate this by checking out the latest version of Ammonite from the official repo, running `sbt sshd/test:run`, following the instructions, and then trying to ssh to port 2222 on localhost using an up-to-date version of the OpenSSH ssh client.

Since I was in there, I upgraded `sshd-core` to version 1.3.0. Some files were moved to different packages, so I changed some imports and code in `SshTestingUtils.scala` and `SshServer.scala`. I then added some extra code to `SshServer.scala` to use the RSA host key authentication algorithm. There are a few extra details in my commit message. I verified that at least in my own environment, these changes eliminate the `...Thei offer: ssh-dss` error message.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -255,7 +255,7 @@ lazy val sshd = project
       crossVersion := CrossVersion.full,
       name := "ammonite-sshd",
       libraryDependencies ++= Seq(
-        "org.apache.sshd" % "sshd-core" % "0.14.0",
+        "org.apache.sshd" % "sshd-core" % "1.3.0",
         //-- test --//
         // slf4j-nop makes sshd server use logger that writes into the void
         "org.slf4j" % "slf4j-nop" % "1.7.12" % "test",
--- a/sshd/src/main/scala/ammonite/sshd/SshServer.scala
+++ b/sshd/src/main/scala/ammonite/sshd/SshServer.scala
@@ -6,12 +6,15 @@ import java.util.Collections
 import ammonite.ops.Path
 import org.apache.sshd.agent.SshAgentFactory
 import org.apache.sshd.common._
+import org.apache.sshd.common.session.Session
 import org.apache.sshd.common.file.FileSystemFactory
 import org.apache.sshd.common.session.ConnectionService
+import org.apache.sshd.common.config.keys.KeyUtils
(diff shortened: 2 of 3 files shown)


axb21:
Travis CI is failing with java version errors on unit tests like this:

```
[info] ------------------Starting Suite ammonite.sshd.SshServerTests------------------
java.lang.UnsupportedClassVersionError: org/apache/sshd/server/auth/password/PasswordAuthenticator : Unsupported major.minor version 52.0
	at java.lang.ClassLoader.defineClass1(Native Method)
	at java.lang.ClassLoader.defineClass(ClassLoader.java:800)
	at java.security.SecureClassLoader.defineClass(SecureClassLoader.java:142)
	at java.net.URLClassLoader.defineClass(URLClassLoader.java:449)
	at java.net.URLClassLoader.access$100(URLClassLoader.java:71)
	at java.net.URLClassLoader$1.run(URLClassLoader.java:361)
	at java.net.URLClassLoader$1.run(URLClassLoader.java:355)
	at java.security.AccessController.doPrivileged(Native Method)
	at java.net.URLClassLoader.findClass(URLClassLoader.java:354)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:425)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:358)
	at ammonite.sshd.SshTestingUtils$.testSshServer(SshTestingUtils.scala:58)
```

I don't know enough about this codebase to diagnose what is going on here, but it appears that somehow the test suite is being compiled with one JDK version but then run on a different JDK version? Oddly, I ran the unit tests before committing this code and did not encounter errors like these; they all passed.

Issue #493, opened 2016-10-28
sourcecode.File() does not return full path to current script


eikek:
Hello,
is it possible to obtain the path to the currently running script? The `$0` bash equivalent. I hope I didn't miss it in the configuration, but so far I couldn't find anything related. It's not super important; I'm now using a simple wrapper around `amm` to export an env var which I then use in my scripts.

``` shell
#!/usr/bin/env sh
export AMM_SCRIPT="$1"
exec java -jar -Xmx500m -XX:+UseG1GC $JAVA_OPTS ~/bin/_amm "$@"
```

Maybe ammonite provides this already?

Thank you for this great tool!

lihaoyi:
`sourcecode.File()` should work but doesn't

```scala
$ cat foo.sc
println(sourcecode.File())
$ amm foo.sc
foo.sc
```

It appears to currently give the file name, when it should be giving the entire absolute path.

I think there problem is here:

- https://github.com/lihaoyi/Ammonite/blob/master/amm/interp/src/main/scala/ammonite/runtime/Compiler.scala#L288

Instead of just the file name, we should be passing in the full path to the file. 

If you have some time, clone the repo and poke around and see what happens and maybe you'll be able to fix it =)

eikek:
Thanks for the pointer! I'll take a look next time I have time.

lihaoyi:
Should be fixed by $497

timcharper:
This bug continue to exist in 1.5.1 and 1.7.1

PR #497, opened 2016-11-03
pass full path to script being compiled

eikek:
Hi,
this is my stab at #493. I took your advice and tracked the calls to this point in `Interpreter.scala`. But… I'm not quite sure if that is the correct place; or all of them. Also, I left the case for `ImportHook.Source.Url` unchanged, but maybe it is sensible to have a full url here, too? I thought an url as return value for `File()` looks strange. And to be honest,  I don't really understand what the url case means. 

I tested with the result of `amm/test:assembly`.


lihaoyi:
Looks good to me; I don't really know what the `Url` case should do either, it was an experiment that hasn't really panned out. Fine to leave it unchanged for now.

Thanks!

timcharper:
This bug exists still in version 1.5.x and 1.7.1; did it regress? Perhaps there should be an integration test?

jasonfharris:
Yep, still looks like this is a problem in version 2.1.4 as well.

Issue #503, opened 2016-11-09
Importing from a directory below doesn't work


wbazant:
I'm trying to use the $file import like documented on the main page, but it breaks.

Am I doing something wrong? Is there a workaround?

```
Welcome to the Ammonite Repl 0.7.9
(Scala 2.11.8 Java 1.7.0_79)
@ import $file.`..`.SomeFile
ammonite.ops.PathError$InvalidSegment: [..] is not a valid path segment. Ammonite-Ops does not allow [..] as a path segment If you are dealing with path-strings coming from external sources, use the Path(...) or RelPath(...) constructor calls to convert them. If you want to use the `..` segment manually to represent going up one level in the path, use the `up` segment from `ammonite.ops.up` e.g. an external path foo/bar/../baz translates into 'foo/'bar/up/'baz.
  ammonite.ops.BasePath$.fail$1(Path.scala:76)
  ammonite.ops.BasePath$.checkSegment(Path.scala:102)
  ammonite.ops.RelPath$.StringPath(Path.scala:202)
  ammonite.runtime.ImportHook$$anonfun$3.apply(ImportHook.scala:66)
  ammonite.runtime.ImportHook$$anonfun$3.apply(ImportHook.scala:66)
  scala.collection.TraversableLike$$anonfun$map$1.apply(TraversableLike.scala:234)
  scala.collection.TraversableLike$$anonfun$map$1.apply(TraversableLike.scala:234)
  scala.collection.mutable.ResizableArray$class.foreach(ResizableArray.scala:59)
  scala.collection.mutable.ArrayBuffer.foreach(ArrayBuffer.scala:48)
  scala.collection.TraversableLike$class.map(TraversableLike.scala:234)
  scala.collection.AbstractTraversable.map(Traversable.scala:104)
  ammonite.runtime.ImportHook$.resolveFiles(ImportHook.scala:66)
  ammonite.runtime.ImportHook$SourceHook.handle(ImportHook.scala:92)
  ammonite.runtime.Interpreter$$anonfun$resolveSingleImportHook$3.apply(Interpreter.scala:151)
```

tkroman:
@wbazant, did you try following the hint in the error message? I mean this:

> If you want to use the `..` segment manually to represent going up one level in the path, use the `up` segment from `ammonite.ops.up`

wbazant:
I tried things like
 ```
import $file.`ammonite.ops.up`.SomeFile
```
but it wasn't really valid syntax.

[...]

PR #504, opened 2016-11-10
Makes `..` work with $file imports as described in docs

tkroman:
This will solve problems like the one in #503. Also, with this PR, both `^` and `..` will work as 'up' directives, so there's a possibility that instead of this PR we should just have the docs fixed (http://www.lihaoyi.com/Ammonite/#MagicImports uses `..` now).


lihaoyi:
Thanks for the PR! This took a bit of time, but decided to go with https://github.com/lihaoyi/Ammonite/pull/606 instead. `..` requires backticks, which I think is really noisy and verbose, despite the way that `..` nicely lines up with a file-paths `..`s. `^`, while unusual, is short and concise. 

Other platforms like Python use pretty arbitrary syntax for this e.g `import ..foo` to import things one level up. So I don't think standardizing on the unusual `^` syntax will be much of a problem, and best to have one way of importing parent-directory scripts rather than two

PR #506, opened 2016-11-14
Fixes NPE/broken REPL on session.load() without save()

### LINKED ISSUE #505: NPE after repl.sess.load()
I was looking for the `:reset` alternative from Scala REPL in Ammonite REPL and thought that I've found it in `repl.sess.load()`. Unfortunately, calling `repl.sess.load()` without previous `repl.sess.save()` seems to unload all the necessary stuff for the REPL itself making it useless.

```scala
./amm
Loading...
Welcome to the Ammonite Repl 0.8.0
(Scala 2.11.8 Java 1.8.0_102)
@ repl.sess.load() 
res0: ammonite.repl.SessionChanged = 
Removed Imports: Set(
  'Resolver,
  'FilterMapArrays,
  'derefPPrint,
  'IvyConstructor,
  'show,
  'typeOf,
  'IvyThing,
  'tprintColors,
  'ArtifactIdExt,
  'RegexContextMaker,
  'pathScoptRead,
  'browse,
  'GrepResult,
  'Pipeable,
  'ChainableConversions,
  'repl,
  'tail,
  'main,
...
@ 1 
java.lang.NullPointerException
@ "foo" 
java.lang.NullPointerException
```

tkroman:
Fixes #505.
The reason was that the call to `Session#save` during the session object doesn't actually help - the first eval frame is empty, since it will be updated only during the `Interpreter` construction. So we have to ensure that right after the interpreter construction (when the predef imports are all set) the call to save has effect. This PR does that by calling `save` via the `bridges->'ammonite.repl.ReplBridge'->sess->save()` handle.


amm/repl/src/main/scala/ammonite/repl/Repl.scala
```diff
@@ -59,14 +55,16 @@ class Repl(input: InputStream,
...
         replArgs
-
       )
       Seq(("ammonite.repl.ReplBridge", "repl", replApi))
     },
     wd
   )
 
-
+  interp.bridges.find(_._1 == "ammonite.repl.ReplBridge") match {
```
savulchik: consider Seq.collectFirst and then Option.foreach

lihaoyi:
Tweaked the comment a bit and merged this as 4032eeea0a01cda43e26b6bc54f5a15f05e7830c. Thanks for the investigation and the fix, and sorry for taking so long!

PR #513, opened 2016-11-25
Use Console for SSHD stream redirection

tjdett:
Previously, logging into an SSH session redirected `System.out` to the console. This caused problems with system behaviour, along with not being safe for multiple sessions.

While Scala's Console object only redirects `println` and not `System.out.println`, it is thread-local, and so behaves in a more sensible way when SSH sessions are active.


lihaoyi:
This looks good. A pity it doesn't work with `System.out.println`, but it appears it's impossible to redirect `System.out.println`. So all the Java undisciplined-println-spam won't be redirected by default, but if someone really wants that behavior I guess they can redirect things themselves just fine

PR #516, opened 2016-11-30
When stripping shebang, we should handle multiline shebangs. Added test case.

Grogs:


Grogs:
Updated to preserve line numbering after stripping shebang. 

I couldn't figure out how to add an automated test for it. Manually tested. 😢 

 `check.session` lets me see there was a compilation error, but not which line it was on.

lihaoyi:
Looks good, thanks!

Issue #245, opened 2015-10-02
Show newly available namespaces when loading dependencies


soc:
There are dependencies where the names and the packages provided differ, sometimes substantially.

In my experience, it's annoying to add a dependency and then having to search which packages have actually become available.

It would be cool if amm provided some helpful hints when loading a dependency, like:

```
loaded foo-bar, new available packages: org.foo.bar, org.foo.internal
```

lihaoyi:
Closing this as a wishlist item; if anyone wants to take a crack at it, go ahead

PR #517, opened 2016-12-01
Show newly available packages when loading dependencies, addresses #245

Grogs:
Addresses #245. Manually tested, raising PR to see which tests it breaks.

Limitation: Only works for dependencies loaded via `interp.load.ivy`. Imports via `import $ivy....` don't display the output. Not sure why, I haven't looked into the mechanics of how $ivy differs from interp.load.

Example output:

```
@ interp.load.ivy("org.scalaj" %% "scalaj-http" % "2.2.0")
Loaded scalaj-http_2.11 2.2.0, with packages: scalaj.http
```

```
@ interp.load.ivy("org.typelevel" %% "cats-core" % "0.8.0")
Loaded cats-core_2.11 0.8.0, with packages: cats.syntax, cats.data, cats, cats.arrow, cats.instances, cats.functor
```

### DIFF
--- a/amm/interp/src/main/scala-2.10/ammonite/interp/CompilerCompatibility.scala
+++ b/amm/interp/src/main/scala-2.10/ammonite/interp/CompilerCompatibility.scala
@@ -1,4 +1,4 @@
-package ammonite.runtime
+package ammonite.interp
 
 import scala.tools.nsc.Global
 import scala.tools.nsc.interactive.{ Global => InteractiveGlobal }
--- a/amm/interp/src/main/scala-2.10_2.11/ammonite/interp/GlobalInitCompat.scala
+++ b/amm/interp/src/main/scala-2.10_2.11/ammonite/interp/GlobalInitCompat.scala
@@ -1,4 +1,4 @@
-package ammonite.runtime
+package ammonite.interp
 
 import ammonite.util.ImportData
 
@@ -15,6 +15,7 @@ import scala.tools.nsc.util._
 import scala.reflect.io._
 import scala.tools.nsc
 import scala.tools.nsc.backend.JavaPlatform
+import ammonite.runtime.Classpath
 object GlobalInitCompat {
   def initInteractiveGlobal(settings: Settings,
                             reporter: AbstractReporter,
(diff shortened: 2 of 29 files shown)


Grogs:
Updated to handle 'import $ivy...` in the same way; print newly loaded packages if verbose is true.

```
@ interp.load.ivy("com.lihaoyi" %% "fansi" % "0.2.3")
Loaded fansi_2.11 0.2.3, with packages: fansi
```

```
@ import $ivy.`com.lihaoyi::scalatags:0.6.1`
Loaded scalatags_2.11 0.6.1, with packages: scalatags.generic, scalatags.stylesheet, scalatags, scalatags.text
```

Grogs:
Now it's printing this on `import $ivy` it's putting output in my scripts, which I don't like. And I can't turn it off unless I switch to interp.load.ivy.

Ideally I'd like a flag which tell me if ammonite is running in interactive mode, so this can be printed in the repl, but not scripts.

@lihaoyi, please let me know your thoughts on this change and/or the original feature request. I'll focus on other things if there's no interest in this feature. If it is something that may get merged, I'm happy to look into how to implement a flag to see if we're in interactive mode.

lihaoyi:
I think this is a good idea. I agree with you that we should show it in the REPL and not in scripts.

At first glance it looks to me like the best way to implement a flag would be to pass a `isScript: Boolean` down from `preprocessScript` or `processLine` down into `ImportHook#handle` and from there into `Interpreter#loadIvy`.

At first glance `verbose` isn't quite what we want for this - we're using that to toggle verbosity of things in scripts, whereas this is scripts vs REPL.

As for explicit `load.ivy` method calls, I don't actually know what the ideal thing to do there would be. Maybe hardcode add an extra `showPkgs: Boolean = false` flag and let people pass in `true` if they really care? I don't think `load.ivy` will get used that much overall now that `import $ivy` is available

lihaoyi:
Getting the following trying to run this

```scala
[error] /Users/lihaoyi/Dropbox/Workspace/Ammonite/amm/src/test/scala/ammonite/CachingTests.scala:123: not enough arguments for method ivy: (coordinates: (String, String, String), interactive: Boolean, verbose: Boolean)Unit.
[error] Unspecified value parameter interactive.
[error]       interp.interpApi.load.ivy("com.lihaoyi" %% "scalatags" % "0.6.2")
[error]                                ^
[warn] one warning found
[error] one error found
[error] (amm/test:compileIncremental) Compilation failed
[error] Total time: 34 s, completed 15 Dec, 2016 4:29:22 PM
```

Make sure you rebase/merge onto master and get things building and let me know when I should take a look at this again

Grogs:
Yeah, it's not ready yet, I'll comment when it's ready for another look. I was pushing WIP commits so I could check CI.

lihaoyi:
Closing due to inactivity

PR #518, opened 2016-12-06
Suggestion for hanging sentence

vitoc:
I just want to indicate that this sentence is incomplete and offer a possible completion to the sentence :) You probably have something else in mind.

p.s. Still looking at Ammonite more :)


lihaoyi:
Nice catch, thanks!

PR #521, opened 2016-12-08
Fix #507 Suppress stack trace for parse error exceptions

### LINKED ISSUE #507: Fail to parse multiline anonymous function
I'm following examples of twitter school using your console (https://twitter.github.io/scala_school/basics.html)
The following example fails to be parsed, though it works properly in _sbt console_

```
@ { i: Int =>
    println("hello world");
    i * 2
  }
```
with the following error:

> fastparse.core.ParseError: found "(\"hello world\");\n  i", expected Semis | End | BacktickId | PlainId | "with" at index 19
>   println("hello world");
>          ^
>   fastparse.core.Parsed$class.get(Parsing.scala:32)
>   fastparse.core.Parsed$Failure.get(Parsing.scala:74)
>   ammonite.runtime.Interpreter$$anonfun$processLine$2$$anonfun$apply$9$$anonfun$15.apply(Interpreter.scala:219)
>   ammonite.runtime.Interpreter$$anonfun$processLine$2$$anonfun$apply$9$$anonfun$15.apply(Interpreter.scala:217)
>   scala.collection.TraversableLike$$anonfun$flatMap$1.apply(TraversableLike.scala:241)
>   scala.collection.TraversableLike$$anonfun$flatMap$1.apply(TraversableLike.scala:241)
>   scala.collection.mutable.ResizableArray$class.foreach(ResizableArray.scala:59)
>   scala.collection.mutable.ArrayBuffer.foreach(ArrayBuffer.scala:48)
>   scala.collection.TraversableLike$class.flatMap(TraversableLike.scala:241)
>   scala.collection.AbstractTraversable.flatMap(Traversable.scala:104)
>   ammonite.runtime.Interpreter$$anonfun$processLine$2$$anonfun$apply$9.apply(Interpreter.scala:217)
>   ammonite.runtime.Interpreter$$anonfun$processLine$2$$anonfun$apply$9.apply(Interpreter.scala:213)
>   ammonite.util.Res$Success.map(Res.scala:62)
>   ammonite.runtime.Interpreter$$anonfun$processLine$2.apply(Interpreter.scala:213)
>   ammonite.runtime.Interpreter$$anonfun$processLine$2.apply(Interpreter.scala:209)
>   ammonite.util.Catching.flatMap(Res.scala:108)
>   ammonite.runtime.Interpreter.processLine(Interpreter.scala:209)
>   ammonite.repl.Repl$$anonfun$action$4$$anonfun$apply$2.apply(Repl.scala:88)
>   ammonite.repl.Repl$$anonfun$action$4$$anonfun$apply$2.apply(Repl.scala:87)
>   ammonite.repl.Scoped$$anonfun$flatMap$1.apply(Signaller.scala:44)
>   ammonite.repl.Signaller.apply(Signaller.scala:29)
>   ammonite.repl.Scoped$class.flatMap(Signaller.scala:44)
>   ammonite.repl.Signaller.flatMap(Signaller.scala:11)
>   ammonite.repl.Repl$$anonfun$action$4.apply(Repl.scala:87)
>   ammonite.repl.Repl$$anonfun$action$4.apply(Repl.scala:74)
>   ammonite.util.Res$Success.flatMap(Res.scala:57)
>   ammonite.repl.Repl.action(Repl.scala:74)
>   ammonite.repl.Repl.loop$1(Repl.scala:98)
>   ammonite.repl.Repl.run(Repl.scala:117)
>   ammonite.Main.run(Main.scala:108)
>   ammonite.Main$$anonfun$main$1$$anonfun$apply$1.apply(Main.scala:249)
>   ammonite.Main$.ammonite$Main$$ifContinually$1(Main.scala:228)
>   ammonite.Main$$anonfun$main$1.apply(Main.scala:230)
>   ammonite.Main$$anonfun$main$1.apply(Main.scala:230)
>   scala.Option.foreach(Option.scala:257)
>   ammonite.Main$.main(Main.scala:230)
>   ammonite.Main.main(Main.scala:-1)
> Something unexpected went wrong =(

shanethehat:
As mentioned in https://github.com/lihaoyi/Ammonite/issues/507, parse errors are output with a stack trace due to them being treated as exceptions. This PR adds a distinct case for parse errors to treat them as failures instead.

### DIFF
--- a/amm/interp/src/main/scala/ammonite/interp/Interpreter.scala
+++ b/amm/interp/src/main/scala/ammonite/interp/Interpreter.scala
@@ -12,6 +12,7 @@ import annotation.tailrec
 import ammonite.util.ImportTree
 import ammonite.util.Util.{CacheDetails, newLine, normalizeNewlines}
 import ammonite.util._
+import fastparse.core
 
 import scala.reflect.io.VirtualDirectory
 
@@ -207,8 +208,9 @@ class Interpreter(val printer: Printer,
   def processLine(code: String, stmts: Seq[String], fileName: String): Res[Evaluated] = {
     val preprocess = Preprocessor(compiler.parse)
     for{
-      _ <- Catching { case ex =>
-        Res.Exception(ex, "Something unexpected went wrong =(")
+      _ <- Catching {
+        case ex: core.ParseError[_, _] => Res.Failure(Some(ex), ex.getMessage)
+        case ex => Res.Exception(ex, "Something unexpected went wrong =(")
       }
 
       (hookImports, hookedStmts, _) <- resolveImportHooks(


lihaoyi:
While this works, it is a bit of a bandaid and I don't think this is quite the right solution here. 

The real problem is here:

```scala
      unwrappedStmts = hookedStmts.flatMap{x =>
        Parsers.unwrapBlock(x) match {
          case Some(contents) => Parsers.split(contents).get.get.value
          case None => Seq(x)
        }
      }
```

That `.get.get` should never fail, and the fact that it does is really is unexpected error. The fact that `.get.get` is blowing up means the `Parsers.unwrapBlock` parser and `Parsers.split` parser are not as in-synced as they should be; we should either bring them in sync or tweak the code so we don't have to call them one-after-the-other and hope they're in sync..

shanethehat:
Ah, okay. So ideally the code in the issue ticket shouldn't fail at all, or are we expecting the parser to issue a Failure rather than an exception?

lihaoyi:
I don't actually know. I (or someone) would have to think a bit about how this is "meant" to behave, given:

- Top-level things wrapped in `{...}` are meant to be unwrapped `...`, before evaluating

- The `...` inside the `{}` that gets evaluated after being unwrapped should evaluate as if it was placed into the REPL alone, without the `{}`s, except being submitted as one chunk of code rather than separate commands

- If you want to actually-evaluate `{...}`, you need to enter `{{...}}` which then gets unwrapped-only-once to `{...}` which then gets evaluated

Given that, it's not immediately clear what should happen to the given snippet of code:

```scala
@ { i: Int =>
    println("hello world");
    i * 2
  }
```

FWIW, I'm not even sure if those guidelines are even self consistent 😛 But if you have any thoughts or ideas, let's hear them...

shanethehat:
Possibly a dumb question, but why are top level things wrapped in braces meant to be unwrapped?

lihaoyi:
Unwrapping is the replacement for `:paste`/Ctrl-D in the normal Scala REPL, to allow you to enter multiple things on different lines that should be in the same block (e.g. class + companion object). Ammonite has no special commands at all, so we can't just copy the `:paste` syntax from the normal REPL, and I thought (?) this was a reasonable workaround that shouldn't be too unintuitive for people

shanethehat:
So I think that it is expected for `unwrapBlock` and `split` to be out of sync here, because they are performing different tasks on different code. What is maybe unexpected is that the `.get.get` is not anticipating failure, preferring to perform an unsafe operation and let the exception handling deal with the problem.

I think it might be nicer to change the second `get` to some kind of match, and let the interpreter fail fast if any of the hooked statements can't be split. I'm happy to play around with that as an approach if you think it's worthwhile?

lihaoyi:
I've thought about this a bit, and here's what I've come up with:

There are two plausible ways you can translate an input such as

```scala
@ { x: Int => x + 1}
```

Into a Scala compilation unit. Translation (1) is

```scala
object cmd1{ x: Int => x + 1}
```

Translation (2) is

```scala
object cmd1{ {x: Int => x + 1} }
```

The first case corresponds to logic where we strip the curlies while leaving the innards un-touched, while the second case corresponds to logic where we fail to strip the curlies if the thing inside the curlies "doesn't make sense" embedded in the enclosing object.

Of these two translations, I think the second makes more sense: the first case will basically never be what anyone wants, since the `x: Int => ...` will be treated as a self-type, which is basically never useful.

To make this happen, the correct solution is probably to change `BlockUnwrapper` from

```scala
P( "{" ~ Block.! ~ "}" ~ End)
```

to

```scala
P( "{" ~ WL.! ~ StatementBlock(Separator) ~ WL ~ "}" ~ End)
```

To bring it in sync with what the `Parsers.split` function ends up doing. That would also let us avoid the double parsing (parsing with `Parsers.unwrapBlock` followed by `Parsers.split`) and ensure that the unwrap-and-split always fails or passes as a whole. That will mean that a curlies will only get unwrapped if the inside is multiple statements and not a single lambda, but that's probably OK and in line with what people would expect

WDYT?

lihaoyi:
Closing due to inactivity

PR #525, opened 2016-12-10
Fixed #515

### LINKED ISSUE #515: Can't instantiate an inner class in 2.12 version of Ammonite
```scala
Welcome to the Ammonite Repl version-SNAPSHOT
(Scala 2.12.0 Java 1.8.0_112)
@ import $ivy.`org.pegdown:pegdown:1.6.0`
:: loading settings :: url = jar:file:/Users/lihaoyi/.coursier/cache/v1/https/repo1.maven.org/maven2/org/apache/ivy/ivy/2.4.0/ivy-2.4.0.jar!/org/apache/ivy/core/settings/ivysettings.xml
:: resolving dependencies :: org.pegdown#pegdown-caller;working
	confs: [default]
	found org.pegdown#pegdown;1.6.0 in central
	found org.parboiled#parboiled-java;1.1.7 in central
	found org.parboiled#parboiled-core;1.1.7 in central
	found org.ow2.asm#asm;5.0.3 in central
	found org.ow2.asm#asm-tree;5.0.3 in central
	found org.ow2.asm#asm-analysis;5.0.3 in central
	found org.ow2.asm#asm-util;5.0.3 in central
import $ivy.$
@ org.pegdown.ast.SimpleNode.Type.HRule
Class file for org.pegdown.ast.SimpleNode$Type not found
Compilation Failed
```

Seems to be a 2.12.0-specific problem, as 2.11.8 still works great. I bet something's still wrong with the classpath code again...

tkroman:
Without going into much depth: classes weren't being found because the classpath used for the lookup was incomplete (actually, consisted only of default JRE jars). That, in turn, was caused by the fact that Ammonite's own overrides didn't affect this case: there's so much ad-hoc / anonymous-fashion instantiations of `Global`/`InteractiveGlobal` all over the compiler internals - it's really hard to track the flow/object dependency graph. 
So the easiest way would be to just ensure that we have a correct `platform` instance. Note that this should also take care of the `loaders` instance (in the interactive global), but in case that's not true: if something similar to this pops up, we may know where to look.


lihaoyi:
Great find! I had removed that `JavaPlatform` thing back in https://github.com/lihaoyi/Ammonite/commit/057440bbf854df78767db1264098e4784f86ab60#diff-83c4b7ec0c41486199bf2d9c0eefc09eR74 for the 2.12 sources, because the tests still seemed to pass, but I didn't realize that some edge cases would break and not get caught by the test suite. 

Could you add a small regression test to https://github.com/lihaoyi/Ammonite/blob/master/amm/src/test/scala/ammonite/session/ProjectTests.scala, with a comment referring to issue #515? No need anything fancy to actually exercise the library, just the minimal case to ensure this doesn't regress in future when we inevitably make further changes. Then I'll merge it and kick off a new stable release for people to start using

tkroman:
@lihaoyi done.

lihaoyi:
Looks great, thanks!

Issue #514, opened 2016-11-25
tweak Ctrl-k behavior (request)


emanresusername:
I'm used to `zsh` so i might be biased, but i almost always want `Ctrl-k` to only delete the current line, instead of deleting **__everything__** after the cursor

say i were to hit `Ctrl-k` in the following scenario, the `|` represents the cursor
```scala
{
|import com.duplicate.thing
import com.duplicate.thing
val lots = """
and
lots
of
lines
after
that
you
want
to
keep
"""
}
```
i'd prefer the result to simply remove the duplicate line, instead of the current behavior which would remove everything after the cursor and leave me with just
```scala
{
```

lihaoyi:
Go ahead and send a PR; many of these commands were ported from Bash, which does not have multiple lines at all, so it makes sense to tweak them to be more useful in a multi-line context

emanresusername:
haven't dug into the code base at all yet, any pointers on where to start?

lihaoyi:
Probably here https://github.com/lihaoyi/Ammonite/blob/3ccb6b18cd5a18ce63070b2179884995125aaa57/terminal/src/main/scala/ammonite/terminal/filters/ReadlineFilters.scala#L151

lihaoyi:
Fixed by https://github.com/lihaoyi/Ammonite/pull/529, should be out in the next unstable build

PR #529, opened 2016-12-11
Ctrl-u/Ctrl-k cut to the beginning/end of line

emanresusername:
#514 
Didn't want to completely scrap the old behavior. I intended this change to re-map the old behavior to `Ctrl-Shift-k` (i.e. `Ctrl-K` (uppercase)) and replace the old `Ctrl-k` with zsh-style "delete to next newline" behavior. 
Wasn't sure how to test locally to confirm that simply upcasing the `k` would have that desired effect though?

### DIFF
--- a/terminal/src/main/scala/ammonite/terminal/filters/ReadlineFilters.scala
+++ b/terminal/src/main/scala/ammonite/terminal/filters/ReadlineFilters.scala
@@ -21,15 +21,15 @@ object ReadlineFilters {
 
   // Backspace  <- delete char
   // Del        -> delete char
-  // Ctrl-u     <- delete all
-  // Ctrl-k     -> delete all
+  // Ctrl-u     <- delete characters on line before cursor
+  // Ctrl-k     -> delete up to newline (or only newline)
   // Alt-d      -> delete word
   // Ctrl-w     <- delete word
 
   // Ctrl-u/-   Undo
   // Ctrl-l     clear screen
 
-  // Ctrl-k     -> cut all
+  // Ctrl-k     -> cut up to newline (or only newline)
   // Alt-d      -> cut word
   // Alt-Backspace  <- cut word
   // Ctrl-y     paste last cut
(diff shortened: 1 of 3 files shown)


lihaoyi:
Rather than leaving `Ctrl-Shift-k` in place, let's just replace it entirely with a `Ctrl-k` that deletes up till the end of the line, and similarly modify `Ctrl-u` to delete from the start of the current line to the cursor.

Other shells do:

- Bash: Ctrl-k delete until end of input-line, Ctrl-u delete until start of input-line
- zsh: Ctrl-k delete until end of line, Ctrl-u delete entire line
- ipython: Ctrl-k delete until end of line, Ctrl-u delete until start of line

The delete-until-start/end-of-input functionality was a naive translation from the readline behavior of the single-line bash shell. Having them both work until start/end of current-line instead of current-input will better match what people probably want in multi-line inputs, while being mostly-consistent with what other shells do and being consistent with Bash

terminal/src/main/scala/ammonite/terminal/filters/ReadlineFilters.scala
```diff
@@ -120,13 +120,53 @@ object ReadlineFilters {
...
+        append(allAfterCursor)
+        (allBeforeCursor, c)
+      } else {
+        val (restOfLine, allAfterLine) = allAfterCursor.splitAt(
+          nextNewlineIndex
+        ) match {
+          case (Vector(), Vector('\n', allAfterNewline @ _*)) ⇒
+            // if there's only a newline after cursor on line, cut it
+            (Vector.empty[Char], allAfterNewline)
+          case nonEmptyCut ⇒
```
lihaoyi: Let's use non-unicode `=>`s for consistency with the rest of the codebase

terminal/src/main/scala/ammonite/terminal/filters/ReadlineFilters.scala
```diff
@@ -120,13 +120,53 @@ object ReadlineFilters {
...
+        )
+        prepend(lineBeforeCursor)
+        val charsBeforeCursorOnLine = lineBeforeCursor.length
+        if (charsBeforeCursorOnLine == 1) {
+          // if only a newline before cursor on line, cut it
+          (allBeforeLine ++ allAfterCursor, c - 1)
+        } else {
+          // if there's more on the line before cursor, cut to beginning of line
+          (
+            allBeforeLine ++ "\n" ++ allAfterCursor,
```
lihaoyi: Let's assign these to temporary `val`s and construct the tuple from those, rather than having these somewhat-awkward multi-line tuples

emanresusername:
looks like the same things that flaked last time (various unresolved dependency errors)

lihaoyi:
This looks great. Thanks for the fix @emanresusername!

PR #537, opened 2016-12-29
Add sudo to chmod installation step

rminehan:
I've tried to install ammonite on a few different machines using the instructions in the docs and often the `chmod` step gets rejected.


lihaoyi:
Looks good, thanks!!!

PR #539, opened 2017-01-01
Add options to disable Ammonite's defalt argument parser (for scripts)

### LINKED ISSUE #479: Script arguments can't start with a dash
I have a script that has these contents:

```
#!/usr/bin/env amm

@main
def ini(args: String*) = {
     ...
}
```

My script can't receive arguments that start with a dash, because they are taken as Ammonite arguments, instead of arguments for my script.

If I try to use e.g.

```
myScript.sc -?
```

then this is the ouput:

```
Error: Unknown option -?
ammonite 0.7.6
Usage: ammonite [options] [<file-args>...] [<args>...]

  -p <value> | --predef <value>
        Any commands you want to execute at the start of the REPL session
  --no-default-predef
        Disable the default predef and run Ammonite with the minimal predef possible
  <file-args>...
        The Ammonite script file you want to execute
  -c <value> | --code <value>
        Pass in code to be run immediately in the REPL
  -x | --execute
        Shim for backwards compatibility - will be removed
  <args>...
        Any arguments you want to pass to the Ammonite script file
  -h <file> | --home <file>
        The home directory of the REPL; where it looks for config and caches
  -f <value> | --predef-file <value>
        Lets you load your predef from a custom location
  -y | --continually
        Lets you run a file over and over, useful for benchmarking purposes
since it lets you hook up a profiler to the long-lived process and
see where all the time is being spent.

  -s | --silent
        Make ivy logs go silent instead of printing though failures will still throw exception
  --repl-api
        Lets you run a script with the `repl` object present; this is
normally not available in scripts and only provided in the
interactive REpl
```

jtcwang:
This fixes #479

This PR adds a `-n / --no-arg-parse` flag to ammonite executable which stops Ammonite from parsing command line arguments intended for the script (after `--`).

Previously this will break because Ammonite is expecting something after the `--world` flag (keyword arguments):

`amm hello.sc -- --world`

But now with `-n` flag, Ammonite will not try to parse anything after `--`. The script with a main function that takes a `String*` vararg will now have `varargs == Array('--world')` and we can plumb that to our custom command line argument parser.

`amm -n hello.sc -- --world`

Happy to rename the flag if you got a better name.

I'll need some pointers as to how I can write tests for this. I couldn't find an example to test command line arguments + script.

### DIFF
--- a/readme/Scripts.scalatex
+++ b/readme/Scripts.scalatex
@@ -228,20 +228,23 @@
         The @hl.scala{main} method does not get automatically called when you
         @hl.scala{load.module} or @hl.scala{load.exec} a script from @i{within}
         the Ammonite REPL. It gets imported into scope like any other method or
-        value defined in the script, and you can just call it normally.
+        value defined in tmhe script, and you can just call it normally.
 
       @p
-        @hl.scala{vararg*} arguments work as you would expect as well, allowing one or
-        more arguments to be passed from the command-line and aggregated into
-        a @code{Seq} for your function to use. This also allows you to use a
-        custom argument-parser (e.g. @lnk("Eugene Yokota",
+        By making your main method receive @hl.scala{vararg*} arguments and running
+        Ammonite with the @code{-n / --no-arg-parse} flag, Ammonite will not attempt
+        to parse the arguments for the script but instead pass them through to the main
+        method. This allows you to use a custom argument-parser (e.g. @lnk("Eugene Yokota",
         "https://github.com/eed3si9n")'s excellent @lnk("Scopt",
-        "https://github.com/scopt/scopt")) library by defining your function
-        as taking @hl.scala{String*}:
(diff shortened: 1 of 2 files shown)


readme/Scripts.scalatex
```diff
@@ -228,20 +228,23 @@
         The @hl.scala{main} method does not get automatically called when you
         @hl.scala{load.module} or @hl.scala{load.exec} a script from @i{within}
         the Ammonite REPL. It gets imported into scope like any other method or
-        value defined in the script, and you can just call it normally.
+        value defined in tmhe script, and you can just call it normally.
```
jhnsmth: typo

jtcwang: done

lihaoyi:
Superseded by https://github.com/lihaoyi/Ammonite/commit/c605d8600c9749b13ade1f85aa6250f959a0a7ff

PR #543, opened 2017-01-04
work around scala reflection bug

tdyas:
Modify the type check for Unit due to a bug in Scala's reflection when applied to certain class annotations in Hadoop jars. The type check would consistently throw an exception when applied to certain Hadoop classes annotated with `org.apache.hadoop.classification.InterfaceAudience`.

See https://issues.scala-lang.org/browse/SI-10129 for additional details.

Workaround: Check the value to be printed against Unit's known instances instead of using the typeOf check.

### DIFF
--- a/amm/repl/src/main/scala/ammonite/repl/ReplAPI.scala
+++ b/amm/repl/src/main/scala/ammonite/repl/ReplAPI.scala
@@ -230,7 +230,26 @@ trait DefaultReplAPI extends FullReplAPI {
                                                  custom: Option[String])
                                                 (implicit cfg: pprint.Config,
                                                  tcolors: pprint.TPrintColors) = {
-      if (typeOf[T] =:= typeOf[Unit]) Iterator()
+      // This type check was originally written as just typeOf[T] =:= typeOf[Unit].
+      // However, due to a bug in Scala's reflection when applied to certain
+      // class annotations in Hadoop jars, the type check would consistently
+      // throw an exception.
+      //
+      // The solution is to catch exceptions thrown by the typeOf check and fallback
+      // to checking the value against Unit's boxed form.
+      //
+      // Why not just check the value? Because that would force evaluzation of `lazy val`'s
+      // which breaks the ammonite.session.EvaluatorTests(lazyvals) test.
+      //
+      // See https://issues.scala-lang.org/browse/SI-10129 for additional details.
+      val isUnit = try {
+        typeOf[T] =:= typeOf[Unit]
+      } catch {
+        case _: Throwable =>
+          value2 == scala.runtime.BoxedUnit.UNIT
+      }
+
+      if (isUnit) Iterator()
       else {
         val implicitPPrint = implicitly[PPrint[V]]
         val rhs = custom match {


lihaoyi:
Do we really need to check against both `()` and `BoxedUnit`? It seems they are equal to each other anyway

```scala
@ scala.runtime.BoxedUnit.UNIT == ()
res1: Boolean = true
```

Passing in `ident` repeatedly as `value` and `value2` was actually intentional in order to work around... something. I don't actually remember what the problem was lol. Anyway if it's still a problem you'll see it in the tests and if it's not a problem anymore then that's great

tdyas:
Yeah, we should just be able to check against `scala.runtime.BoxedUnit.UNIT` then. Also, I bet `()` gets boxed to `scala.runtime.BoxedUnit.UNIT`.

The comments earlier in the file said:
```
    /**
     * Kind of an odd signature, splitting out [[T]] and [[V]]. This is
     * seemingly useless but necessary because when you add both [[pprint.TPrint]]
     * and [[PPrint]] context bounds to the same type, Scala's type inference
     * gets confused and does the wrong thing
     */
```

At least on Scala 2.11.8, turning the context bounds into their equivalent implicit evidence arguments seemed to remove the need for having two copies of `value`.

In any event, there are some test failures, will dig into those.

lihaoyi:
looks like we can't use `value == ()` because it forces `value` and breaks the lazy val tests

tdyas:
Yeah I noticed that too when I dug into the test failures. I modified the check so it uses typeOf first and falls back to checking against BoxedUnit.UNIT if typeOf throws an exception.

tdyas:
classOf doesn't work with generic types unfortunately. I then tried to use ClassTag instead. But then `ammonite.session.ProjectTests.shapeless` fails with:

```[info] ammonite.session.ProjectTests.shapeless
[info]          		utest.AssertionError: assert{
[info]          		error: String = Main4.sc:8: No ClassTag available for Int(2)
[info]          		            .print(res4, res4, "res4", _root_.scala.None)
[info]          		                  ^
[info]          		Compilation Failed
[info]          		warning: String =
[info]          		out: String =
[info]          		info: String =
[info]          		expected: String = res4: 2 = 2
```

The check was: ```      val unitClass: Class[_] = classTag[Unit].runtimeClass
      val clazz: Class[_] = classTag[T].runtimeClass
      val isUnit = unitClass.isAssignableFrom(clazz)
```

**Comment by alexarchambault** [comment]
Why not simply put the test in a try catch, like
```scala
val isUnit =
  try (typeOf[T] =:= typeOf[Unit])
  catch {
    case _: Throwable => false
  }
```
?

tdyas:
The last commit I pushed to this PR already does exactly that. :)

tdyas:
So even though my commit now passes tests, it doesn't fix the issue with the Hadoop class because the catch is only catching `Exception`'s and not `TypeError` which inherits from `Throwable` directly. I'm going to figure out a way to narrow what is actually caught since `Throwable` is too broad.

tdyas:
So I pushed a commit where I catch `Throwable` to fallback. It works for me. One test failed due to a rate limit issue, should be intermittent.

lihaoyi:
Looks good to me, thanks @tdyas!

Issue #538, opened 2016-12-30
Enable `args` in script


ctongfei:
Is it possible to enable access to `args` variable in code without a function annotated with `@main`?
It'd be good if `args` can be accessed.

jtcwang:
This feature is part of #423 which I'm currently working on.

lihaoyi:
Fixed in https://github.com/lihaoyi/Ammonite/commit/c605d8600c9749b13ade1f85aa6250f959a0a7ff#diff-d624941e18d4c528c725d4c3c0bd0517R1089

This lets `@main` methods with varargs properly slurp in everything, allowing you to then use your own custom argument parser. The only caveat is you cannot use `--` in your own argument list, since that is reserved to separate Ammonite's arguments from your script's arguments. 

Please take a look at the changelog entry which describes this and let me know if this satisfies your requirements

lihaoyi:
http://www.lihaoyi.com/Ammonite/#0.9.0 is out

jtcwang:
Excellent will give it a go. Thanks!

PR #549, opened 2017-01-15
Pass unmodified arguments to the interpreter

suiryc:
As requested in #538 and attempted in #440, here is a commit to pass non-ammonite arguments to the interpreter, available as `interp.args` once amm is launched.

Comes with a unit test to have more chances to land ;)

### DIFF
--- a/amm/interp/src/main/scala/ammonite/interp/Interpreter.scala
+++ b/amm/interp/src/main/scala/ammonite/interp/Interpreter.scala
@@ -35,6 +35,7 @@ class Interpreter(val printer: Printer,
                   // the REPL before it starts
                   extraBridges: Interpreter => Seq[(String, String, AnyRef)],
                   val wd: Path,
+                  args: Seq[String] = Nil,
                   verboseOutput: Boolean = true)
   extends ImportHook.InterpreterInterface{ interp =>
 
@@ -684,6 +685,8 @@ class Interpreter(val printer: Printer,
       }
 
     }
+
+    lazy val args = Interpreter.this.args
   }
 
 }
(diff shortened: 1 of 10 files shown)


suiryc:
Also added an integration test.
Now both the interpreter and the script execution are tested for those arguments.

lihaoyi:
I see this, but won't have time to look at it for a few days I think. I'll need to properly page in all my Ammonite-internals related data into my head so I can properly compare it with #539 and decide what to do. Might need to wait until the weekend

suiryc:
Fair enough.

My intention was more like keeping separate the usage of `@main` entry points (which may be multiple and for which ammonite use its own argument parsing) and the possible usage of `interp.args` when a CLI parser like scopt is used in the script (and thus where no entry point would be needed/used).

Having the script file being the separator between ammonite and the script args as proposed in #479 would have been nice. But since scopt does not seem to expose the options (at least names) that were used to build its parser it would be less elegant when looking for where is actually the script file in the CLI arguments.
Anyway, looking forward being able to run `amm` and use scopt inside the script, with e.g. a shebang like
```
#!/bin/bash
exec amm "$0" -- "$@"
!#
```
(and possibly the `--no-arg-parse` to make sure script arguments are not parsed at all)

jtcwang:
Thanks for this. For scripts to accept arbitrary arguments we do need the `--no-arg-parse` PR merged first. Currently I'm stuck on setting up tests for that PR (since ATM there are no integration tests involving making sure that Ammonite arguments e.g. `-s` are parsed correctly)

lihaoyi:
Superseded by https://github.com/lihaoyi/Ammonite/commit/c605d8600c9749b13ade1f85aa6250f959a0a7ff

PR #550, opened 2017-01-16
Fix #454

### LINKED ISSUE #454: Bytecode not available for classes defined in ammonite repl
Ammonite REPL:

``` scala
Ammonite> case class Child(name: String)
defined class Child

Ammonite> val cls = classOf[Child]
cls: Class[Child] = class $sess.cmd0$Child

Ammonite> val resName = cls.getName.replace('.', '/') + ".class"
resName: String = "$sess/cmd0$Child.class"

Ammonite> cls.getClassLoader.getResource(resName)
res3: java.net.URL = null // Some URL representing the compiled class is expected here 

Ammonite> cls.getClassLoader.getResourceAsStream(resName)
res4: java.io.InputStream = null // Some InputStream of the compiled class is expected here
```

Scala REPL:

``` scala
scala> case class Child(name: String)
defined class Child

scala> val cls = classOf[Child]
cls: Class[Child] = class Child

scala> val resName = cls.getName.replace('.', '/') + ".class"
resName: String = Child.class

scala> cls.getClassLoader.getResource(resName)
res0: java.net.URL = repldir:(memory)/Child.class

scala> cls.getClassLoader.getResourceAsStream(resName)
res1: java.io.InputStream = java.io.ByteArrayInputStream@436e852b
```

jhnsmth:
Fix #454. Ported from [here](https://github.com/scala/scala/blob/2.12.x/src/reflect/scala/reflect/internal/util/AbstractFileClassLoader.scala#L68-L80). Not sure if put the test into the right place (or whether it needed at all)

### DIFF
--- a/amm/runtime/src/main/scala/ammonite/runtime/ClassLoaders.scala
+++ b/amm/runtime/src/main/scala/ammonite/runtime/ClassLoaders.scala
@@ -1,7 +1,10 @@
 package ammonite.runtime
 
-import java.net.{URL, URLClassLoader}
+import java.io.ByteArrayInputStream
+import java.net.{URL, URLClassLoader, URLConnection, URLStreamHandler}
 import java.nio.ByteBuffer
+import java.util.{Collections, Enumeration}
+
 import acyclic.file
 import ammonite.ops._
 import ammonite.util.{Imports, Util}
@@ -154,4 +157,25 @@ class SpecialClassLoader(parent: ClassLoader, parentSignature: Seq[(Path, Long)]
       case _ => Nil
     })
   }
+
+  override def findResource(name: String) =
+    getURLFromFileDict(name).getOrElse(super.findResource(name))
+
+  override def findResources(name: String) = getURLFromFileDict(name) match {
+    case Some(u) => Collections.enumeration(Collections.singleton(u))
+    case None    => super.findResources(name)
+  }
+
+  private def getURLFromFileDict(name: String) = {
+    val className = name.stripSuffix(".class").replace('/', '.')
+    newFileDict.get(className) map { x =>
+      new URL(null, s"memory:${name}", new URLStreamHandler {
+        override def openConnection(url: URL): URLConnection = new URLConnection(url) {
+          override def connect() = ()
+          override def getInputStream = new ByteArrayInputStream(x)
+        }
+      })
+    }
+  }
+
 }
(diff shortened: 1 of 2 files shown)


lihaoyi:
Looks like this makes CI mad as it can't find Akka configs. I've tried re-starting those jobs and they fail with the same error, so probably not just flakes. Akka loads the default config through resources IIRC

jhnsmth:
@lihaoyi Thanks for the quick reply. Didn't  see them , when run the tests locally, because akka project test aren't  run for 2.12. I think it the second overridden function is guilty for that. I'll let you know , when I fix it.

jhnsmth:
Travis doesn't like me, two build failed complaining about github API rate limit. Looks like this failures aren't related to my changes and [failed](https://travis-ci.org/lihaoyi/Ammonite/jobs/192502277#L2443) akka test is [green now](https://travis-ci.org/lihaoyi/Ammonite/jobs/192852773#L2465), so please review.

BTW Since those tests randomly fail, maybe it makes sense to replace github api with something else or even to remove them.

lihaoyi:
Yeah I just ignore the rate limited tests. They suck but they'll do until someone gets around to fixing/removing them. Feel free to send a PR to replace that API endpoint with some other

amm/runtime/src/main/scala/ammonite/runtime/ClassLoaders.scala
```diff
@@ -154,4 +157,25 @@ class SpecialClassLoader(parent: ClassLoader, parentSignature: Seq[(Path, Long)]
...
+    getURLFromFileDict(name).getOrElse(super.findResource(name))
+
+  override def findResources(name: String) = getURLFromFileDict(name) match {
+    case Some(u) => Collections.enumeration(Collections.singleton(u))
+    case None    => super.findResources(name)
+  }
+
+  private def getURLFromFileDict(name: String) = {
+    val className = name.stripSuffix(".class").replace('/', '.')
+    newFileDict.find(_._1 == className) map { c =>
```
lihaoyi: Could this be replaced by `newFileDict.get(className)`? That will let it be O(1) rather than O(n) as the size of `newFileDict` grows

jhnsmth: done

lihaoyi:
It's not easy being geen

lihaoyi:
This should be published in the next unstable version if CI doesn't flake

PR #555, opened 2017-01-19
Update sbt and plugins

jhnsmth:
Updated readme to scala 2.12.  Also renamed project/build.sbt to project/plugins.sbt to avoid confusion with build.sbt. and removed sbt-dependency-graph plugins, it's just an informative one so having it in repo seems useless.


lihaoyi:
Looks good, thanks!

PR #557, opened 2017-01-20
Avoid github exceeded API rate limit error in tests

jhnsmth:
An attempt to remove this annoying error. I replaced github api  with [jsonplaceholder](http://jsonplaceholder.typicode.com/) ,which has no rate limits.

### DIFF
--- a/integration/src/test/resources/ammonite/integration/basic/HttpApi.sc
+++ b/integration/src/test/resources/ammonite/integration/basic/HttpApi.sc
@@ -1,33 +1,31 @@
 #!/usr/bin/env amm
 // HttpApi.sc
 import scalaj.http._
-/**
-  * Github URL shortener script using Scalaj-HTTP and the git.io API.
-  */
+
 @main
-def shorten(longUrl: String) = {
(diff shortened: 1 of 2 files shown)


jhnsmth:
To be honest I'm sure whether it's the right way to fix it or if this service is reliable. So I'm glad to hear any suggestions. 

I personally think this test should be removed, as It tests noting: upickle and scalaj.Http have their own test suits and passing args to a script is covered by other tests. The only reason to keep it  is to ensure that the [provided example](http://www.lihaoyi.com/Ammonite/#BundledLibraries) is correct,  but the changes makes it useless for this case. Anyway it's shouldn't be kept as is.

lihaoyi:
Let's do this. At the very least hopefully it'll stop the rate-limit flaking

We can figure out whether we want to remove this test, or convert the doc-example to follow, later

AronT-Emelai:
It seems that when this change was merged into the code, the function *shorten* was replaced by *addPost* & *listReleases* by *comment* which I don't think is what you wanted to do!

PR #558, opened 2017-01-23
(WIP) Publishing for Scala 2.10 and 2.11

jhnsmth:
it's not finished yet.  but let's start a discussion.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,10 +33,10 @@ matrix:
 
   - env: CI_SCRIPT="ci/build.sc test published/test"
     jdk: openjdk7
-    scala: 2.10.5
+    scala: 2.10.6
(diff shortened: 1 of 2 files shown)


lihaoyi:
Let's give it a shot

lihaoyi:
Looks good with a tweak or two that I applied after; let me get this working since this is the most sketchy/untested portion of the CI code, and then sending another PR to put a link in the readme should be easy and locally-testable

jhnsmth:
thanks for the quick reply! I'll send another PR to update docs in a few days. My idea is to  re[place 'curlUrl' and 'unstablecurlUrl' with sequences of tuples with scala version and a link  to the corresponding binary.

PR #559, opened 2017-01-24
Minor docs fixes

jhnsmth:
Just fixed remaining .scala extensions and removed trailing white-spaces.

Looks like you already did all necessary docs changes for #558.  But if there is still something I could help with, please let me know.


lihaoyi:
Looks good to me, thanks!

PR #562, opened 2017-01-30
Add an overloaded version of LoadJar.cp that takes a sequence of paths

mallman:
Add an overloaded version of `LoadJar.cp` that takes a sequence of paths for efficiency. In testing performance adding 226 jars, the overloaded version is about two orders of magnitude faster than the original method.

Also, trim some EOL whitespace and add an EOF newline (BSR)


mallman:
This looks like a spurious test failure. Is it possible to rerun the CI build?

lihaoyi:
No need to re-run the CI, it looks good to me, thanks!

mallman:
Thank you!

Issue #460, opened 2016-09-05
How to load ivy dependencies through a HTTP proxy


DavidPerezIngeniero:
In an Ammonite script I have this line:

```
import $ivy.`org.postgresql:postgresql:9.4.1209.jre7`
```

The result when running this script is:

```
Server access error at url http://repo1.maven.org/maven2/org/postgresql/postgresql/9.4.1209.jre7/postgresql-9.4.1209.jre7.pom (java.net.SocketException: Connection reset)
Exception in thread "main" ammonite.runtime.tools.IvyThing$IvyResolutionException: failed to resolve ivy dependencies unresolved dependency: org.postgresql#postgresql;9.4.1209.jre7: not found
```

It seems that a proxy isn't used even though the environment variables `http_proxy` and `https_proxy` are properly set.

lihaoyi:
Ammonite doesn't provide any way to allow proxies right now; would be happy to make it take them as something you could initialize in your predef, like how we do resolvers today

DavidPerezIngeniero:
Or like SBT that autodetects proxies by using the http_proxy environment variable. (At least in Linux).

PerArneng:
This would be a very nice feature to have. It is very hard to make use of the excellent Ammonite when behind a corporate firewall with a mandatory http proxy.

PerArneng:
It seems to use the proxy when these are set.
```
$ export JAVA_OPTS="-Dhttp.proxyHost=my.little.proxy.com -Dhttp.proxyPort=808"
```
However it adds the scala version to all the dependencies and can not find them.

```
$ ./test.sc
Compiling test.sc
:: loading settings :: url = jar:file:/usr/local/bin/amm!/org/apache/ivy/core/settings/ivysettings.xml
:: resolving dependencies :: org.fusesource.jansi#jansi_2.12-caller;working
        confs: [default]

:: problems summary ::
:::: WARNINGS
                module not found: org.fusesource.jansi#jansi_2.12;1.14

        ==== local: tried

          /home/kod425/.ivy2/local/org.fusesource.jansi/jansi_2.12/1.14/ivys/ivy.xml

          -- artifact org.fusesource.jansi#jansi_2.12;1.14!jansi_2.12.jar:

          /home/kod425/.ivy2/local/org.fusesource.jansi/jansi_2.12/1.14/jars/jansi_2.12.jar

        ==== m2: tried

[...]

PR #563, opened 2017-01-31
#460 - Read env proxy settings.

cuzfrog:
#460 

Ivy reads jvm system properties for proxy settings. This change looks for proxy info in environment variables, and set in system properties.

1. Added a utility object `ProxyFromEnv` that fetch current env proxy info.
2. Change only invades `IvyThing.scala` - above `ivy.resolve(md, options)`.
3. Unit test for `ProxyFromEnv` covers most common cases.
4. `System.setProperty()` in `predef.sc` will override envs. 

How did the integration test go(I had tough time compiling):

1. Use `ufw` to only allow proxy network.
2. Make `amm` run in its own jvm by  `set fork := true`.
3. put `interp.ivy.load...` in `/tmp/loadivy.sc`
4. run `test:run` with arg `"/tmp/loadivy.sc"` 

I enabled and disabled proxy resolution, it did what I expected.

Please check is there anything wrong or needed modification. Thanks.


lihaoyi:
This will probably need to be reworked due to https://github.com/lihaoyi/Ammonite/commit/662859f84f5a77efca3e7e4d3074626570a992cf ?
I'm not familiar with how proxying works in @alexarchambault's Coursier, but presumably it will work differently, and as of https://github.com/lihaoyi/Ammonite/commit/b5c44b0d8496b07adc2d323c497fdf622a22fd09 you can properly configure stuff in a script's earlier blocks rather than needing to build the functionality into Ammonite proper

cuzfrog:
@lihaoyi 

>you can properly configure stuff in a script's earlier blocks

Yes. As we discussed in #460:  https://github.com/lihaoyi/Ammonite/issues/460#issuecomment-275983057

We can set `System.setProperty` in `predef.sc`. This is what exactly `ivy` does. I did not inspect that in `Coursier`.

Being able to read proxy is no generallly considered as a feature, I think it only deserves an intuitive enhancement. And it's pretty minor. **This PR is to give Ammonite the ability to read system env, the configuration resides in OS, not in Ammonite proper**. And it depends on the user whether setting up proxy in Linux system env.

lihaoyi:
I don't really understand. Given we're not using sbt/ivy anymore, why should we `System.setProperty`? If Coursier depends on it, then we should make Coursier work, but if it's user-code which depends on it, that code can do the setting

cuzfrog:
@lihaoyi 

### The explanation:

I checked Ammonite 0.8.5, proxy works the same way. I reckon Coursier did not change any proxy setting over ivy. 

    System.setProperty("http.proxyHost","some_address")
    System.setProperty("http.proxyPort","some_port")

You see `"http.proxyHost"` and `"http.proxyPort"` are the java parameter names. see:[Oracal Doc](https://docs.oracle.com/javase/7/docs/api/java/net/doc-files/net-properties.html). This is a java property.

### PR updated:

1.Added a utility object ProxyFromEnv into package `ammonite.runtime.tools` alongside with `IvyThing`.
2.Call `ProxyFromEnv.setPropProxyFromEnv()` at the beginning of `IvyThing.resolveArtifact`
3.Others and test strategy remain the same. See top of this thread.

**Negtive test:** If no system proxy has been setup (in linux):
```
@ import $ivy.`com.google.guava:guava:18.0` 
https://repo1.maven.org/maven2/com/google/guava/guava/18.0/guava-18.0.pom
         [          ] 0 B (0 B / s)

```
**Positive test:** Setup system proxy:
```
@ import $ivy.`com.google.guava:guava:18.0` 
https://repo1.maven.org/maven2/com/google/guava/guava/18.0/guava-18.0.pom
  100.0% [##########] 5.5 KiB (4.4 KiB / s)
```

lihaoyi:
@cuzfrog If these are JVM-level properties, and affect everything rather than just Coursier/Ivy, I think we should initialize them in the main method since that's where all the rest of the "process-global" stuff lives.

cuzfrog:
@lihaoyi Negative. If, in the main method, once the system envs have changed, we have to restart amm.

lihaoyi:
@cuzfrog it seems reasonable to say that if you want to pick up modified system env variables, you have to restart Ammonite. Most programs I'm aware of aren't responsive to changing env vars halfway through a process. 

I'm also not comfortable with making the `Repl`, which is otherwise a Plain-Old-Scala-Object, start mucking with process-wide env through `IvyThing`. If you are in the `main` method, we can be confident that Ammonite is the only thing using this process, but otherwise if someone is embedding Ammonite within a larger application, this sort of global side-effect is icky. And if someone wants to force reload it, while they have a REPL running, they can do so in code manually.

I'll merge this as is with the proxy setting moved to `main`, and we can continue discussion if you really-truly think it should be otherwise

lihaoyi:
Huh for some reason this is making the travis go beserk and print out tons of ivy-related output... not sure why

lihaoyi:
I had to back this out due to the build failures I do not have time to investigate.

@cuzfrog it's up to you to make the build green and then I'll merge this again

alexarchambault:
@cuzfrog Why do the env var have to be updated? Setting system properties from them should be enough.

cuzfrog:
Update:

1. move `ProxyFromEnv` to package `ammonite.main`.
2. Call ProxyFromEnv.setPropProxyFromEnv() at the beginning of `Main.run()`. (I saw object `Main.main` is primarily doing args parsing, and not necessarily called when starting REPL.)
3. Manual test OK.

Please review the code, and give me some advice. Thanks very much.

lihaoyi:
Code looks fine I guess.

My main concern is that this diff still seems to be causing Travis to print out huge quantities of spam during the test suite, which is the primary problem I couldn't figure out that caused me to back out my merge (https://travis-ci.org/lihaoyi/Ammonite/jobs/235463840#L2199). 

This slows down the build (it normally doesn't take 27 minutes!), causes timeouts, and doesn't happen on master.

Could you investigate that and see what's happening? That's my main constraint; the rest of my nitpicks I can tweak/fix myself on-merge

cuzfrog:
I was finally able to reproduce the berserk logs locally and find out why.

There is a "fall back mode" detection inside method: `TermDisplay.defaultFallbackMode`, where Coursier checks if the context is `CI` [Portal](https://github.com/coursier/coursier/blob/master/cache/src/main/scala/coursier/TermDisplay.scala#L69).

So I changed this line with `IvyThing`: 
```scala
val logger = new coursier.TermDisplay(new PrintWriter(Console.err),fallbackMode = true)
```
And the test: https://github.com/cuzfrog/Ammonite/blob/proxy-test/proxy-test/src/test/scala/anywhere/ProxyTest.scala

You can clone my fork and test in branch `proxy-test`.

And the truncations happen not within `truncatedPrintln`, but in `reflowed`: [portal](https://github.com/coursier/coursier/blob/master/cache/src/main/scala/coursier/TermDisplay.scala#L260)

-------------------------------------

So the problem should be clear: It happens because of Coursier's output strategy falls back to adapt those that Coursier thinks don't or should not support progress update. This should not happen when a common user uses amm, who, by chance, has set up a bad proxy. However, when this does happen, we should give the user some indication that the proxy is not working(in fact generally the network is not working), despite it's the user's responsibility to ensure their network's availability.

lihaoyi:
@cuzfrog sorry for neglecting this

I don't fully understand your explanation. What does setting `fallbackMode = true` do? Does it prevent the spam from happening? Does it force it to always happen? Or something else?

cuzfrog:
@lihaoyi 
`fallbackMode = true` is to manually set the "console mode" to `fallback`, bypassing Coursier's checking logic. This is a parameter Coursier reserves so that it can be set manually(for testing, etc.)

By setting `fallbackMode = true`, Coursier regards the console as a "fallback" console, so it will print as it does in Travis CI console(reproduce the berserk logs, but in my local console).

Then we are able to explain the scenario: When a bad proxy blocks the network, Coursier retries and spawns the urls.

PR #570, opened 2017-02-14
Replaced while loops with functions that return a Vector, avoiding

bullocp913:
Replaced while loops with functions that return a Vector, avoiding having to call .toVector on the mutable Buffer.

There is a modest startup gain.

Tests in amm subproject all pass.
The long lines test fails.....

ammonite-ops_2.12-0.8.1.jar:ammonite/Constants.class complains about different file contents when assembly is tried on main ammonite project, does it check for "current" and "files"?

### DIFF
--- a/amm/runtime/src/main/scala/ammonite/runtime/Classpath.scala
+++ b/amm/runtime/src/main/scala/ammonite/runtime/Classpath.scala
@@ -33,9 +33,12 @@ object Classpath {
     loop(System.getProperty(property).split(java.io.File.pathSeparator),Vector())
   }
 
-  private def getCPFromClassLoader(loader:ClassLoader = Thread.currentThread.getContextClassLoader):Vector[File] = {
+  private def getCPFromClassLoader(
+    loader:ClassLoader = Thread.currentThread.getContextClassLoader): Vector[File] = {
     @annotation.tailrec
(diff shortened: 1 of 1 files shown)


lihaoyi:
I don't really understand this; it would have expected constructing it as a `Vector` to be just as slow as building a `mutable.Buffer` and calling `toVector`.

Furthermore, if the goal is just to avoid the `toVector` call while still keeping `classpath` immutable, can we just convert

```diff
-  val classpath = files.toVector.filter(_.exists)
+  val classpath: IndexedSeq[...] = files.filter(_.exists)
```

and call it a day?

bullocp913:
Hi,

I was thinking it was better to stay with a collection that stays in so called 'constant time' as stated in the scala docs(as close as Vector could get). When I first wrote this, it was actually a List....., but then I found out that it needed be a Vector somewhere else.

 Here is the doc I'm referring to: http://docs.scala-lang.org/overviews/collections/performance-characteristics.html

lihaoyi:
Ok, I'm not convinced there is a performance gain in this diff, and even if there is one I don't think it's necessary to rewrite everything as you have done so, and I don't think this change has been sufficiently thought through given what you've said so far. 

If you can write up your reasoning, motivation and performance results more thoroughly I'll take a look again

PR #571, opened 2017-02-19
Add `read` to docs when comparing with `cat`

stephennancekivell:
Add `read` to docs when comparing with `cat`


lihaoyi:
looks good thanks!

PR #572, opened 2017-02-20
Update example to scalaz 7.2.7

stephennancekivell:
Keeps it consistent with the earlier example and 7.1.1 isnt publish for scala 2.12.


lihaoyi:
Looks good i guess

PR #578, opened 2017-03-18
Add a sleep in Shellout.executeStream to reduce spinning.

rorygraves:
Whilst waiting for IO Shellout.executeStream was running a tight loop and consuming significant
amounts of CPU.  Adding a small sleep when no input was seen reduces cpu usage significant.

### DIFF
--- a/ops/src/main/scala/ammonite/ops/Shellout.scala
+++ b/ops/src/main/scala/ammonite/ops/Shellout.scala
@@ -8,7 +8,7 @@ import scala.io.Codec
 import scala.language.dynamics
 
 /**
-  * Iternal utilities to support spawning subprocesses
+  * Internal utilities to support spawning subprocesses
   */
 object Shellout{
   val % = new Command(Vector.empty, Map.empty, Shellout.executeInteractive)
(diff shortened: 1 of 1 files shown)


ops/src/main/scala/ammonite/ops/Shellout.scala
```diff
@@ -65,14 +65,17 @@ object Shellout{
     val stdout = process.getInputStream
     val stderr = process.getErrorStream
     val chunks = collection.mutable.Buffer.empty[Either[Bytes, Bytes]]
+    val sources = Seq(stdout -> (Left(_: Bytes)), stderr -> (Right(_: Bytes)))
```
rorygraves: This prevents allocations during each cycle.

ops/src/main/scala/ammonite/ops/Shellout.scala
```diff
@@ -81,6 +84,9 @@ object Shellout{
           ))
         }
       }
+      // if we did not read anything sleep briefly to avoid spinning
```
rorygraves: Only sleep if we have not seen input during this iteration.  This means that for fast processes we do not add unnecessary pauses.

rorygraves:
@lihaoyi CI build failure looks unrelated

lihaoyi:
Thanks!

Issue #121, opened 2015-07-25
Repl.run: Add `postdef` argument


sirthias:
I'm using ammonite-repl in a project, which talks to a serial adapter over a native library.
Everything runs smoothly, however there is one thing I'd like to get rid of:
Having to manually clean up at the end of my session, which is tedious and easy to forget.

It'd be great if the `Repl.run(...)` had a `postdef` argument in addition to `predef`, which allowed me to automatically run a few commands upon having pressed Ctrl-D to exit the session.

lihaoyi:
Seems reasonable. To be clear, this is to clean up when you exit the ammonite-repl, *not* when you exit the JVM right?

lihaoyi:
Here's another question: are you calling this programmatically? If you're calling `Repl.run` programmatically then it seems you could just... add additional code after the `Repl.run` call =P 

The other case, which is to run stuff upon exiting the JVM, you can already accomplish via shutdown hooks e.g. http://stackoverflow.com/questions/2921945/useful-example-of-a-shutdown-hook-in-java

sirthias:
I want to run commands when I exit the REPL, not the JVM.
I know I can run stuff after ammonite-repl has exited, yes. I already do this and call `sys.exit` to immediately exit the standard SBT console after the ammonite-repl session has finished.
However, in my cleanup code I need access to the definitions I create in `predef` for the ammonite-repl,
so the cleanup code needs to run in the same scope rather than in an enclosing one.

lihaoyi:
> However, in my cleanup code I need access to the definitions I create in predef for the ammonite-repl,
so the cleanup code needs to run in the same scope rather than in an enclosing one.

Ok makes sense

sirthias:
Great!
By the way: Thanks for some great tools!
Ammonite is not the only awesome thing that you've built!

lihaoyi:
Closing as wishlist; if anyone wants to implement this that'd be great. Otherwise I probably won't get to it

joprice:
I just found myself wishing I had this after starting an http client, then hitting ctrl+d, and waiting a bit until I realized it was hanging. It would be nice to have something like sbt's `cleanupCommands`.

emanresusername:
@lihaoyi i'm willing to take a whack at this (assuming the barrier to entry isn't too high), any pointers on where to start?

lihaoyi:
Yeah this should be one of the most straightforward issues ever to implement 😛 

Starting points:

[...]

PR #581, opened 2017-03-21
add beforeExitHooks for cleanup before repl exits

emanresusername:
inspired by #121 but not exactly what we were discussing there
not sure if this is the best (or even a good 😅 ) way to do this
but really just wanted something that worked [like this](https://github.com/emanresusername/.ammonite/blob/akkators/predef.sc#L112-L125) and this handled that usecase well enough for me :)

glad to hear other thoughts/suggestions/improvements/etc

### DIFF
--- a/amm/interp/src/main/scala/ammonite/interp/Interpreter.scala
+++ b/amm/interp/src/main/scala/ammonite/interp/Interpreter.scala
@@ -55,6 +55,7 @@ class Interpreter(val printer: Printer,
   val dynamicClasspath = new VirtualDirectory("(memory)", None)
   var compiler: Compiler = null
   var pressy: Pressy = _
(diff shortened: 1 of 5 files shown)


deusaquilus:
I'm also seeing issues in the Travis Scala 2.10.6 build and I don't understand the error at all. Maybe this build is flaky?

emanresusername:
yeah, i was getting something about a `class.bak` file somewhere not being found, pretty sure it's not related to the code changes here @lihaoyi any ideas?

lihaoyi:
Yeah that's just a flake; we can ignore it, unless you want to dig into why it's happening (I haven't had the time to)

lihaoyi:
Looks good to me

PR #587, opened 2017-04-02
Fix warnings

andrelfpinto:


lihaoyi:
I doubt this is going to work as is; as you can see, the build is broken, because the newer non-warning APIs we could use do not exist in older versions of Scala. We could split those callsites into version-specific source files, but I think it's easier just to live with a few deprecation warnings or swap over to https://github.com/milessabin/macro-compat

PR #594, opened 2017-05-03
Various improvements to ammonite-sshd:

espenhw:
* Modern OpenSSH clients will refuse to accept an ssh-dss server key; generate using RSA instead
* Enable keyboard-interactive authentication as well as password-based
* Allow password-based and/or publickey-based authentication


lihaoyi:
Merged this into https://github.com/lihaoyi/Ammonite/commit/b80b7638454cafdfff68368793614b5e5f81a454

PR #604, opened 2017-05-19
Shade coursier / scalaz / scala-xml

alexarchambault:
This PR shows how coursier / scalaz / scala-xml (all brought by coursier) can be shaded into ammonite.

For this to work fine, the shaded coursier / scalaz shouldn't be used from the REPL, as shading doesn't adapt some scala-specific annotations in the JARs (discussed in https://github.com/coursier/coursier/issues/454 - should happen with any shading lib for now).

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -149,6 +149,9 @@ lazy val amm = project
         val oldStrategy = (assemblyMergeStrategy in assembly).value
         oldStrategy(x)
     },
+    assemblyShadeRules in assembly := shadedNs.toSeq.map { ns =>
+      ShadeRule.rename(s"$ns.**" -> s"$shadingNs.@0").inAll
+    },
     parallelExecution in Test := false
   )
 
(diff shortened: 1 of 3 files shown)


alexarchambault:
Maybe went too fast on that one, that only works when launching ammonite via coursier
```
$ sbt published/publishLocal
$ coursier launch com.lihaoyi:ammonite_2.12.2:version-SNAPSHOT
```

If you want to do the same in the assembly, you'll have to setup https://github.com/sbt/sbt-assembly#shading.

lihaoyi:
Thanks for the PR! I'm probably not going to merge this for now, since I've started directly exposing Coursier APIs to the user via the `interp.repositories` reference. Now Coursier is a user-facing part of Ammonite, and shading bits and pieces of it may cause it to behave in confusing ways...

scala-xml is already brought in by `scala-compiler`, so no real benefit in shading that

I'm still hoping for Scalaz to be removed as a dependency of Coursier - It's literally contributing over 10 megabytes of the 36mb Ammonite download - but I don't think this is worth prioritizing landing right now. This will definitely be a useful reference if the need re-arises

PR #605, opened 2017-05-21
Cookbook HTTP Requests update scalaj-http to work with scala 2.12.x

piotrkwiecinski:
Updated docs to make sure that HTTP Requests example compiles with Scala 2.12


lihaoyi:
lgtmthx

PR #606, opened 2017-05-21
Fix import file docs to reflect changes to parent dir introduced in 0.7.2

piotrkwiecinski:
Update docs to match changes introduced in 0.7.2:

Changed the syntax for import $file segments outside your current directory from import $file.`..`.foo to import $file.^.foo. This makes it a lot shorter and mitigates a problems caused by the file-name being too long.

Instead of introducing second version of parent separator as in https://github.com/lihaoyi/Ammonite/pull/504

Aligned with current implementation:
https://github.com/lihaoyi/Ammonite/blob/master/amm/util/src/main/scala/ammonite/util/Util.scala#L13

### DIFF
--- a/readme/Repl.scalatex
+++ b/readme/Repl.scalatex
@@ -275,7 +275,7 @@
           Or if the script is in an @i{outer} folder,
 
         @hl.scala
-          @@ import $file.`..`.MyScript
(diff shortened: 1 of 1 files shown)


lihaoyi:
Thanks @piotrkwiecinski!

PR #608, opened 2017-05-22
Update jline to 2.14.3

jhnsmth:
2.14.3 is the latest bug fix release for 2.x.x branch. Update to 3.x.x branch isn't possible as it requires Java 8.

Assembly workaround is not needed any more, cause now jline versions in Scala  2.11.11, 2.12.x and Ammonite are the same.


lihaoyi:
Looks good, thanks!

Someday perhaps we can upgrade to jline3 and drop our own custom multi-line terminal implementation...

Issue #611, opened 2017-05-23
Resolution of local Maven artefacts does not work


mduerig:
Since the move to coursier in 0.9 resolving artefacts only published to the local Maven repository does not work anymore unless I add

`interp.repositories() ++= Seq(coursier.MavenRepository("file://" + java.lang.System.getProperties.get("user.home") +  "/.m2/repository/"))`

And even then I have to explicitly specify the version (e.g. `1.3-SNAPSHOT`) where before I could simply us `latest.integration`.

lihaoyi:
@mduerig Feel free to add those artifacts to this listing https://github.com/lihaoyi/Ammonite/blob/7b18fcc2cb935c30a07c563e8d6dfd98dd6ec94e/amm/runtime/src/main/scala/ammonite/runtime/tools/IvyThing.scala#L73-L76

mduerig:
Not sure whether this is the right approach. It is rather a work around relying on a specific location of the local Maven cache.

lihaoyi:
I don't know much about maven, but having a hard-coded list of "places jars are most likely to resolve from" seems like a reasonable thing to do. We already hardcode sonatype and ivyLocal, what's a few more? 😛 if the list starts getting large/unwieldy then can we can fix it, but if it's just 1 or 2 then it's probably fine

mduerig:
Pull request for the suggested change: https://github.com/lihaoyi/Ammonite/pull/612

lihaoyi:
Closing this, as discussed in https://github.com/lihaoyi/Ammonite/pull/612#issuecomment-303522406

PR #612, opened 2017-05-23
 Resolution of local Maven artefacts does not work #611

mduerig:
Add resolver for local Maven cache

### DIFF
--- a/amm/runtime/src/main/scala/ammonite/runtime/tools/IvyThing.scala
+++ b/amm/runtime/src/main/scala/ammonite/runtime/tools/IvyThing.scala
@@ -72,7 +72,8 @@ object IvyThing{
 
   val defaultRepositories = List(
     coursier.Cache.ivy2Local,
-    coursier.MavenRepository("https://repo1.maven.org/maven2")
+    coursier.MavenRepository("https://repo1.maven.org/maven2"),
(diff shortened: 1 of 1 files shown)


alexarchambault:
I wouldn't recommend adding `~/.m2/repository` by default. It sometimes misses artifacts (like it has the POMs but not the JARs), which confuses coursier - if a repo has a POM, coursier expects it to have the JAR too... I could add a helper in coursier (like `coursier.MavenRepository.m2Local`) to make it easier to add it from ammonite though.

mduerig:
`coursier.MavenRepository.m2Local` would be a better solution and nicely inline with `coursier.Cache.ivy2Local`.

alexarchambault:
Possibly even `coursier.Cache.m2Local` :-) Or I wanted to move `ivy2Local` under `coursier.IvyRepository`...

lihaoyi:
@mduerig Regardless of what changes you end up doing to coursier, do you still want to add it by default to Ammonite's resolvers? If not you can always add it manually in your own predef/earlier-block, and I'll close this issue/PR

mduerig:
@lihaoyi , I agree with @alexarchambault that this is better tackled in one way or another in coursier. Once this is done we can get back to this issue and see how to fix it. In the meanwhile let's close the PR.

lihaoyi:
sgtm

Issue #423, opened 2016-08-01
Use custom argument parser with ammonite


benwbooth:
I would like to be able to use a custom argument parser with ammonite instead of the built in one. How can I achieve this? Is there some way to access the command line arguments directly from the script without needing to use the `@main` annotation? It would be nice if there were an annotation like this:

`@args val args: Seq[String] = null`

or maybe:

`val args: Seq[String] = ammonite.main.Scripts.args`

which automatically filled in the command line arguments to a variable. This would also enable users to not have to provide a `@main` function in their scripts. Is this possible to achieve?
Thanks!

lihaoyi:
Ideally 

```scala
@main def main(args: String*) = {
  ...
}
```

Should work. It currently doesn't, but that can be fixed.

We could expose the arguments as a global, which is what Python does, but I'd like to avoid that if possible. Exposing the command-line arguments as, well, arguments should in theory push people in the direction of dealing with them at the entry-point rather than grabbing them randomly throughout their scripts, which I think is the right thing to do

benwbooth:
Fair enough. The one reason I like having the arguments as a "global" or being able to get them from a global function, is that it allows me to write the entire script at the top-level instead of within the main() function, which to me feels more script-y.

But if the syntax you posted worked, that would be a big improvement.

benwbooth:
What about setting `this.args` to be the full command-line arguments in the script top-level? That way, the arguments are visible only to the script, and are not global.

lihaoyi:
Discussed with @coderabhishek in gitter, we will try first to make String* work fine with the `generateRoutes` macro and scopt. If that doesn't work, we'll fall back to having it available as a global.

lihaoyi:
Fixed by https://github.com/lihaoyi/Ammonite/commit/7608d6f726fd05256f03e237a49ea20b69887502; this allows you to take `String*` arguments in your `@main` methods, which will then take in all kinds of sets of arguments un-changed that you can then feed into whatever argument parser you want

benwbooth:
I noticed a problem, with the following script I'm not getting the right command line arguments:

```
#!/usr/bin/env amm

@main def main(args: String*) = {
  println(args)
}
```

If I run that script using a `--`:

```
$ ./test.sc a b c -- d e
Vector(d, e, a, b, c)
```

The `--` is removed, and everything after the `--` is moved before the first set of arguments. This bug will prevent users from being able to correctly use custom argument parsers.

[...]

PR #619, opened 2017-06-01
[WIP] Bring Ammonite in closer alignment with argument conventions of other interpretters

### LINKED ISSUE #618: Ammonite parameter conventions out of alignment with common interpreters such as bash/Python/Ruby
In version 0.9.0, the argument passing behavior of Ammonite was changed in order to support script arguments directly following the script name. This particular functionality is compatible with other interpreters.

What is not expected, however, is:

- inability to provide ammonite parameters prior to the script in question
- the use of `--` is not consistent with other interpreters (script is the first argument following the `--`)

timcharper:
As documented in #618, it was not possible to specify ammonite arguments
if also specifying a scala script. With the change, the following
invocations work:

    amm --silent my-script.sc scriptarg1 scriptarg2
    amm --silent my-script.sc -- scriptarg1 scriptarg2
    amm my-script.sc --silent -- scriptarg1 scriptarg2
    amm my-script.sc --silent scriptarg1 scriptarg2

Additionally, we add assertions to prevent scriptargs from being
specified both before and after the delimiter. The following is
rejected:

    amm my-script.sc --silent scriptarg1 -- scriptarg2

Fixes #618

(we also add proper `--help`)

### DIFF
--- a/amm/src/main/scala/ammonite/Main.scala
+++ b/amm/src/main/scala/ammonite/Main.scala
@@ -179,6 +179,8 @@ object Main{
     var replApi = false
     var remoteLogging = true
     var watchScripts = false
+    var fileToExecute: Option[Path] = None
+    val indent = "                           "
     val replParser = new scopt.OptionParser[Main]("ammonite") {
       // Primary arguments that correspond to the arguments of
       // the `Main` configuration object
@@ -210,11 +212,6 @@ object Main{
         .foreach(x => codeToExecute = Some(x))
         .text("Pass in code to be run immediately in the REPL")
 
-      arg[String]("<args>...")
-        .optional()
-        .unbounded()
-        .foreach{ x => passThroughArgs = passThroughArgs :+ x }
-        .text("Any arguments you want to pass to the Ammonite script file")
       opt[File]('h', "home")
         .valueName("<file>")
         .foreach( x => ammoniteHome = Some(Path(x, pwd)))
@@ -231,33 +228,51 @@ object Main{
       opt[Unit]("repl-api")
         .foreach(x => replApi= true)
         .text(
-          """Lets you run a script with the `repl` object present; this is
-            |normally not available in scripts and only provided in the
-            |interactive REPL
-          """.stripMargin
+          s"""Lets you run a script with the `repl` object present; this is
+             |${indent}normally not available in scripts and only provided in the
+             |${indent}interactive REPL
+           """.stripMargin.trim
         )
+      arg[String]("script-file")
+        .optional()
+        .text("Path to a scala script for Ammonite to run")
+        .foreach { x => fileToExecute = Some(Path(x, pwd)) }
+      arg[String]("<script-args>...")
+        .optional()
+        .unbounded()
+        .foreach{ x => passThroughArgs = passThroughArgs :+ x }
+        .text(s"""Any arguments you want to pass to the Ammonite script file.
+                 |${indent}To pass arguments that begin with a -, please use -- to
+                 |${indent}separate Ammonite args from script args
+               """.stripMargin.trim)
+
+      help("help").text("Show this help and exit")
     }
 
     // amm foo.sc
     // amm -h bar
     // amm foo.sc hello world
     // amm foo.sc -h bar -- hello world
-    val (fileToExecute, ammoniteArgs, flatScriptArgs) = args0.lift(0) match{
-      case Some(x) if x.head != '-' =>
-        val fileToRun = Some(Path(args0(0), pwd))
-        // running a file
-        args0.indexOf("--") match {
-          // all args to to file main
-          case -1 => (fileToRun, Array.empty[String], args0.drop(1))
-          // args before -- go to ammonite, args after -- go to file main
-          case n => (fileToRun, args0.slice(1, n), args0.drop(n+1))
-        }
-      case _ => (None, args0, Array.empty[String]) // running the REPL, all args to to ammonite
+    val (ammoniteArgs, flatScriptArgs) = args0.indexOf("--") match {
+      case -1 =>
+        (args0, Array.empty[String])
+      case idx =>
+        (args0.slice(0, idx), args0.drop(idx + 1))
     }
 
-    val scriptArgs = ammonite.main.Scripts.groupArgs(flatScriptArgs)
 
     for(c <- replParser.parse(ammoniteArgs, Main())) {
+      val scriptArgs: Seq[(String, Option[String])] =
+        if (flatScriptArgs.nonEmpty && passThroughArgs.nonEmpty) {
+          Console.err.println(
+            "Error: You may not provide script args before and after the -- delimiter")
+          System.exit(1)
+          ???
+        } else if (flatScriptArgs.nonEmpty) {
+          ammonite.main.Scripts.groupArgs(flatScriptArgs)
+        } else {
+          ammonite.main.Scripts.groupArgs(passThroughArgs)
+        }
       def main(isRepl: Boolean) = Main(
         c.predef,
         c.defaultPredef,
@@ -297,7 +312,7 @@ object Main{
                 if (value != ()){
                   pprint.PPrinter.BlackWhite.pprintln(value)
                 }
-              case Res.Skip   => // do nothing on success, everything's already happened
+              case Res.Exit(_) | Res.Skip => // do nothing on success, everything's already happened
             }
             if (!watchScripts) false
             else{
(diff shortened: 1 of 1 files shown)


timcharper:
Behavior is now as I would expect with this change:

```
16:35 $ java -jar amm/target/scala-2.12/ammonite-version-SNAPSHOT-2.12.2  ~/test.sc  --silent -- arg1 arg2
List(arg1, arg2)
 ~/src/oss/Ammonite [master ↑·1|✚ 1…61]
16:35 $ java -jar amm/target/scala-2.12/ammonite-version-SNAPSHOT-2.12.2  ~/test.sc  --silent arg1 arg2
List(arg1, arg2)
 ~/src/oss/Ammonite [master ↑·1|✚ 1…61]
16:35 $ java -jar amm/target/scala-2.12/ammonite-version-SNAPSHOT-2.12.2  ~/test.sc  --silent arg1 -- arg2
Error: You may not provide passThrough args before and after the -- delimiter
 ✘-1 ~/src/oss/Ammonite [master ↑·1|✚ 1…61]
16:36 $
```

timcharper:
Not sure if I caused these failures. `ammonite.session.ProjectTests` passes for me, locally:

```
[info] ------------------Starting Suite ammonite.session.ProjectTests------------------
[info] ammonite.session.ProjectTests.load.ivy.standalone		Success
[info] ammonite.session.ProjectTests.load.ivy.akkahttp		Success
[info] ammonite.session.ProjectTests.load.ivy.resolvers		Success
[info] ammonite.session.ProjectTests.load.ivy		Success
[info] ammonite.session.ProjectTests.load.code		Success
[info] ammonite.session.ProjectTests.load		Success
[info] ammonite.session.ProjectTests.shapeless		Success
[info] ammonite.session.ProjectTests.scalaz		Success
[info] ammonite.session.ProjectTests.cats		Success
[info] ammonite.session.ProjectTests.guava		Success
[info] ammonite.session.ProjectTests.resources		Success
[info] ammonite.session.ProjectTests.scalaparse		Success
[info] ammonite.session.ProjectTests.finagle		Success
[info] ammonite.session.ProjectTests.spire		Success
[info] ammonite.session.ProjectTests.pegdown		Success
[info] ammonite.session.ProjectTests		Success
[info] -----------------------------------Results-----------------------------------
[info] ammonite.session.ProjectTests		Success
[info]     load		Success
[info]         ivy		Success
[info]             standalone		Success
[info]             akkahttp		Success
[info]             resolvers		Success
[info]         code		Success
[info]     shapeless		Success
[info]     scalaz		Success
[info]     cats		Success
[info]     guava		Success
[info]     resources		Success
[info]     scalaparse		Success
[info]     finagle		Success
[info]     spire		Success
[info]     pegdown		Success
[info]
[info] Tests: 16
[info] Passed: 16
[info] Failed: 0
```

amm/src/main/scala/ammonite/Main.scala
```diff
@@ -231,33 +228,51 @@ object Main{
...
+        (args0.slice(0, idx), args0.drop(idx + 1))
     }
 
-    val scriptArgs = ammonite.main.Scripts.groupArgs(flatScriptArgs)
 
     for(c <- replParser.parse(ammoniteArgs, Main())) {
+      val scriptArgs: Seq[(String, Option[String])] =
+        if (flatScriptArgs.nonEmpty && passThroughArgs.nonEmpty) {
+          Console.err.println(
+            "Error: You may not provide passThrough args before and after the -- delimiter")
```
mgedigian: "passThrough args" is terminology from the implementation, not user-facing (in usage help or docs)

timcharper: Yeah, I thought about this after I'd pushed. Updated.

lihaoyi:
This was changed intentionally in http://www.lihaoyi.com/Ammonite/#0.9.0. The behaviour is documented [here](http://www.lihaoyi.com/Ammonite/#AmmoniteArgumentsinScripts).

A solution is to use the 

```
#!/bin/bash
exec amm "$0" --predef 'println("welcome!")' -- "$@"
!#
```

shebang, as is now (since I made the change) mentioned in the doc-site.

From what I understand, a shebang such as `#!/usr/bin/env amm -s` was never really portable and [doesn't work on Linux](https://stackoverflow.com/questions/4303128/how-to-use-multiple-arguments-with-a-shebang-i-e). You'd have to do the multi-line shebang above anyway if you want to run your scripts on e.g. a server

I'm a bit surprised the tests pass, I guess we don't have much test coverage of the `main` code, but this breaks the following use case:

```scala
lihaoyi Ammonite$ cat foo.sc
@main def main(args: String*) = args

lihaoyi Ammonite$ amm foo.sc --bar # 0.9.x
ArrayBuffer("--bar")

lihaoyi Ammonite$ amm/target/amm foo.sc --bar # PR #619
Error: Unknown option --bar
Try --help for more information.

lihaoyi Ammonite$ amm/target/amm foo.sc -bar # PR #619
Error: Unknown option -bar
Try --help for more information.
```

Which was important for people who wanted to properly "pass through" all arguments for their own custom processing, in cases such as #423, #479, #538.

I'd be curious if you could make this work while maintaining the behaviour shown above, and while maintaining understandability ("always put script first" is relatively straightforward...). Otherwise, you could just swap over to the `exec`-style shebangs and call it a day

lihaoyi:
With some thought, I'm pretty sure we could come up with an implementation that satisfies those requirements, but I don't think we'd be able to use Scopt to do it. We'd need to loop over and deal with the arguments ourselves (as we are already doing in other parts of Ammonite, e.g. the `@main` method router)

timcharper:
@lihaoyi thank you for clarifying! I am ashamed I didn't dig deeper to read the documentation; I simply learned of the behavior change by it breaking when I updated to 0.9.x, finding many of my scripts had broken, and had only read the `--help` screen from Ammonite.

The behavior you describe matches my experience. What was the guiding principle when arriving on that implementation? It seems inconsistent with most interpreter implementations that I've checked.

I have a ruby script, python script, and bash script which simply output their arguments on launch. The following is the output:

```
$ python -d output-args.py -- a b c
['--', 'a', 'b', 'c']

$ python output-args.py -d -- a b c
['-d', '--', 'a', 'b', 'c']

$ python -d -- output-args.py a b c
['a', 'b', 'c']

$ python -d output-args.py a b c
['a', 'b', 'c']

--------

$ ruby -v output-args.rb -- a b c
ruby 2.0.0p648 (2015-12-16 revision 53162) [universal.x86_64-darwin16]
["--", "a", "b", "c"]

$ ruby output-args.rb -v -- a b c
["-v", "--", "a", "b", "c"]

$ ruby -v -- output-args.rb a b c
ruby 2.0.0p648 (2015-12-16 revision 53162) [universal.x86_64-darwin16]
["a", "b", "c"]

$ ruby -v output-args.rb a b c
ruby 2.0.0p648 (2015-12-16 revision 53162) [universal.x86_64-darwin16]
["a", "b", "c"]

--------

$ bash -x output-args.sh -- a b c
+ echo -- a b c
-- a b c

$ bash output-args.sh -x -- a b c
-x -- a b c

$ bash -x -- output-args.sh a b c
+ echo a b c
a b c

$ bash -x output-args.sh a b c
+ echo a b c
a b c
```

It would be possible for me to modify this pull request to support the following cases:

```
amm --predef 'println("welcome!")' script.sc scriptarg1 scriptarg2
amm --predef 'println("welcome!")' -- script.sc scriptarg1 scriptarg2
amm --predef 'println("welcome!")' -- script.sc scriptarg1 scriptarg2

# these do not match convention of bash/ruby/python/etc. but can be supported for backwards compat at the cost of complexity

amm script.sc --predef 'println("welcome!")' -- scriptarg1 scriptarg2
amm --predef 'println("welcome!")' script.sc -- scriptarg1 scriptarg2
```

Would such a change be welcome by you? What is your preference?

lihaoyi:
The guiding principle is "I want to be able to pass in unchanged script args to the script, but I'm too lazy to re-implement Scopt to allow it to properly forward trailing arguments after the script name" 😛

lihaoyi:
I think I would have very much preferred "args coming before script go to amm, args after to go script", but couldn't figure out how to implement that easily; if you want to contribute an implementation I'd take it

lihaoyi:
We have existing code which does the "walk arguments one by one and figure out what to do with them" in here:

- https://github.com/lihaoyi/Ammonite/blob/master/amm/src/main/scala/ammonite/main/Scripts.scala#L19
- https://github.com/lihaoyi/Ammonite/blob/master/amm/src/main/scala/ammonite/main/Router.scala#L54

Would be great if we could re-use the code rather than duplicate it

lihaoyi:
Fixed by https://github.com/lihaoyi/Ammonite/commit/24c424495d9ed012a065dd6adbc17538432a63d5

timcharper:
Awesome, this is much better <3 thanks @lihaoyi !!

PR #621, opened 2017-06-02
Make Spark2 example more idiomatic

donderom:
The idea is to use SparkSession for Spark integration example as it's more idiomatic way in Spark 2.

### DIFF
--- a/integration/src/test/resources/ammonite/integration/basic/Spark2.json
+++ b/integration/src/test/resources/ammonite/integration/basic/Spark2.json
@@ -1,6 +1,4 @@
-[
-  {"hello": 1},
-  {"hello": 2},
+{"hello": 1},
+{"hello": 2},
 
-  {"hello": 3}
-]
+{"hello": 3}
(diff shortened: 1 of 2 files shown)


integration/src/test/resources/ammonite/integration/basic/Spark2.json
```diff
@@ -1,6 +1,4 @@
-[
-  {"hello": 1},
-  {"hello": 2},
+{"hello": 1},
+{"hello": 2},
 
-  {"hello": 3}
-]
+{"hello": 3}
```
lihaoyi: Uhhh... this isn't a JSON file anymore? Is spark fine with that?

lihaoyi:
Looks good i guess, except for the weird non-JSON file... not saying it's wrong, but if you could tell me why it isn't wrong I'll merge this

integration/src/test/resources/ammonite/integration/basic/Spark2.json
donderom: That's how Spark JSON reader works by default. With the current Spark2.json the result of `df` would be:
```
+---------------+-----+
|_corrupt_record|hello|
+---------------+-----+
|              [| null|
|           null|    1|
|           null|    2|
|           null|    3|
|              ]| null|
+---------------+-----+
```

Now it produces the following:
```
+-----+
|hello|
+-----+
|    1|
|    2|
|    3|
+-----+
```
The official Spark documentation uses the same format for their examples.

lihaoyi: Huh ok, well I guess this is great then

PR #622, opened 2017-06-03
Fix object Properties is not a member of package ammonite.util in example-predef.sc

piotrkwiecinski:
After downloading new predef introduced in 0.9.5 I get:

```
.ammonite/predef.sc:3: object Properties is not a member of package ammonite.util
  s"ammonite-shell_${util.Properties.versionNumberString}" %
```
There is difference between example-predef.sc and example-predef-bare.sc


lihaoyi:
Looks good to me, thanks!

Issue #546, opened 2017-01-09
Loading jar file in script via interp.load.cp does not work properly


danbills:
after running interp.load.cp([valid Path]) I am unable to use classes from Jar file.

I tried the same code in Repl and it works as expected.

I also printed out repl.sess.frames.head.classpath and it shows the jar file in the classpath, as expected.

Unfortunately when I try to import or use classes it simply says "value not found" on root package name.

code I'm running:
```
import $ivy.`com.google.protobuf:protobuf-java:3.1.0`

val path = java.nio.file.FileSystems.getDefault().getPath("/home/dan/speech/jsoundProto/target/scala-2.10/jsoundproto_2.10-0.1-SNAPSHOT.jar")

val x = ammonite.ops.Path(path)

interp.load.cp(x)

println(repl.sess.frames.head.classpath)

//FAILS HERE
val x = billco.sound.Javasound.MixerInfo.newBuilder

println("done")
```

foxlog:
No problem in my Mac OS X.

val path = java.nio.file.FileSystems.getDefault().getPath("/tmp/ojdbc14.jar")
val x = ammonite.ops.Path(path)
interp.load.cp(x)

oracle.jdbc.driver.Const.BUILD_DATE

//res7: String = "Sat_Feb__2_11:39:35_PST_2008"

danbills:
That's strange, I'm on Mac OS X.

I verified w/ another jar (scala parser combinators):
```
val path = java.nio.file.FileSystems.getDefault().getPath("/tmp/scala-parser-combinators_2.12-1.0.5.jar")
val x = ammonite.ops.Path(path)
interp.load.cp(x)

import scala.util.parsing.combinators.RegexParsers //No package parsing found in scala.util
```

danbills:
Also confirmed on LInux java 8

lihaoyi:
You need to put a `@` between your `interp.load.cp` and your `import`, as documented http://www.lihaoyi.com/Ammonite/#Multi-stageScripts

PR #628, opened 2017-06-11
make the required use of ampersand a bit clearer in docs

danbills:
I originally confused the use of `@` in docs as part of the formatting for the README.

This snippet should make it clear that 

  * it's code
  * it's required

Inspired by my errant bug here: https://github.com/lihaoyi/Ammonite/issues/546


lihaoyi:
Looks good, thanks!

PR #634, opened 2017-06-15
Include @doc annotations for main methods in usage

mgedigian:
@doc annotations for main methods are described in documentation.  Everything is already in place to process them, but they're not displayed in the usage.
With this change, those docs are included on the line following the method name.


lihaoyi:
Looks good to me, thanks!

Issue #635, opened 2017-06-15
Allow full-cross-versioning with :::


lihaoyi:
Right now you can import java-style ivy dependencies:

```scala
import $plugin.$ivy.`org.scalamacros:paradise:2.1.0`
```

And Scala-style dependencies:

```scala
import $plugin.$ivy.`org.scalamacros::paradise:2.1.0`
```

Where the `::` automatically appends the scala binary version, e.g. `_2.12`, to the name of the artifact. However, in some cases we really want to append the scala *full* version, e.g. `_2.12.2`, to the name of the artifact. This can be done now manually:

```scala
import $plugin.$ivy.`org.scalamacros:paradise_2.12.2:2.1.0`
```

But it would be nice to have a `:::` syntax to handle this common use case for us:

```scala
import $plugin.$ivy.`org.scalamacros:::paradise:2.1.0`
```

lihaoyi:
Fixed in #637

PR #637, opened 2017-06-16
#635 Allow full cross-versioning

aeffrig:
I've never worked with Ammonite source before, but here is an attempt to allow cross-versioning.  Please comment if you believe more test cases are needed or if you believe there are better ways to handle the issue.


lihaoyi:
@aeffrig no, this looks great. Thanks!

Issue #152, opened 2015-08-10
Incorrect behavior on Home, End, etc. under screen


senia-psm:
Reproduce steps: start screen, start Ammonite using `sbt projectName/test:console` (I can't download https://git.io/vOdoc), tap Home, tap Enter:
```Scala
@
SyntaxError: found "\u001b[1~", expected Import | Prelude ~ BlockDef | Expr | End in
```
End, Enter:
```Scala
@
SyntaxError: found "\u001b[4~", expected Import | Prelude ~ BlockDef | Expr | End in
```
Same for Alt+arrow.
Home, End, etc works fine for default scala repl under screen and for Ammonite without screen.
Difference between `export` under screen and without screen is in `TERM`, `TERMCAP`, `WINDOW`, `SHLVL`, `STY`, `OLDPWD` variables. `TERM="xterm" TERMCAP="" WINDOW="" SHLVL="1" STY="" OLDPWD="" sbt projectName/test:console` command doesn't fix Ammonite under screen and doesn't break Ammonite without screen.

Blaisorblade:
I can also reproduce this under `tmux`, a `screen` replacement that also sets `TERM=screen`.

Workaround: use C-a/C-e (that is, Ctrl-a for Home, or Ctrl-e for End).

> TERM="xterm" TERMCAP="" WINDOW="" SHLVL="1" STY="" OLDPWD="" sbt projectName/test:console command doesn't fix Ammonite under screen and doesn't break Ammonite without screen.

Makes total sense to me.

Background: Different terminal emulators use different escape sequences for the same commands, so native *x programs need to use libraries to interpret those escape sequences (that is, termcap/terminfo, or higher-level libraries wrapping those, such as readline or ncurses). Those libraries are hard to access on the JVM, hence I guess Ammonite doesn't use them. Indeed, escape sequences are hardcoded in `terminal/src/main/scala/ammonite/terminal/SpecialKeys.scala`.

Working hypothesis: screen sends for Home/End a different escape sequence than xterm, hence Ammonite-REPL does not understand it.

lihaoyi:
Is termcap/terminfo accessible via shelling out? That would be the easiest, since we already shell out to get TTY height/width, set terminal state, etc. like we do [here](https://github.com/lihaoyi/Ammonite/blob/master/terminal/src/main/scala/ammonite/terminal/Utils.scala#L51-L72)

[...]

PR #638, opened 2017-06-17
Fix home, end keys in urxvt

aolshevskiy:
Same issue as #152 but for urxvt. It emits `^[[7~` for Home and `^[[8~` for End.

By the way you can use your favorite Linux distro `/etc/inputrc` as a reference. Here is one from my Arch Linux: https://gist.github.com/aolshevskiy/cd4113179e22e1a831b65276614611de

Here is an excerpt with relevant screen and urxvt keycodes:
```
# for linux console and RH/Debian xterm
"\e[1~": beginning-of-line
"\e[4~": end-of-line
...
"\e[7~": beginning-of-line
...
# for rxvt
"\e[8~": end-of-line
```

And one more thing: you're doing great work here! Cheers :clap:


lihaoyi:
Looks good to me, thanks!

PR #650, opened 2017-06-27
Don't require WeakTypeTag for printing (fix #649)

### LINKED ISSUE #649: Wrapper method generated by Ammonite REPL is too large to commit to JVM
The log is too large, see attachment.

[amm-log.txt](https://github.com/lihaoyi/Ammonite/files/1104382/amm-log.txt)

Atry:


amm/repl/src/main/scala/ammonite/repl/FullReplAPI.scala
```diff
@@ -41,27 +58,11 @@ trait FullReplAPI extends ReplAPI{
...
-      // See https://issues.scala-lang.org/browse/SI-10129 for additional details.
-      val isUnit = try {
-        typeOf[T] =:= typeOf[Unit]
-      } catch {
-        case _: Throwable => value == scala.runtime.BoxedUnit.UNIT
-      }
+    def print[T: pprint.TPrint: FullReplAPI.IsUnit](
+        value: => T,
+        ident: String,
+        custom: Option[String])(implicit tcolors: pprint.TPrintColors) = {
```
lihaoyi: Let's maintain the old formatting style here

lihaoyi:
The idea looks great. Clearly the tests don't pass now, but If you could make them pass I'll happily merge this

Atry:
I don't know how to fix it. The implicit resolution for `Nothing` seems very strange

Atry:
I simply use `ClassTag` instead of `WeakTypeTag`, as other solutions are all buggy due to https://github.com/scala/bug/issues/10393

Atry:
@lihaoyi How do you think about the new `ClassTag` approach?

lihaoyi:
Looks reasonable. Can you put in a comment that enumerates the obvious/alternative solutions (typetag, special IsUnit typeclass, classtag without the default value) and explains why they dont work in this case. Then I'll merge it since everything is green

Atry:
I didn't mention `IsUnit` type class because this solution is similar to `T =:= Unit`.

Atry:
@lihaoyi Is it mergeable now?

lihaoyi:
Looks good to me

PR #654, opened 2017-07-05
Magic import for $url 

Atry:


lihaoyi:
CI is complaining about long lines.

Today is pretty packed, will take a look at this when I'm a bit more free

Atry:
Long line fixed.

lihaoyi:
A few notes:

- This does not currently allow "relative" `import $file`s within a `import $url`ed. That is probably fine, but is worth taking note of in the commit message, and possibly as a comment in the code.

- What's the story around caching? As far as I know this downloads the script every time and runs it every time. 
    - This seems reasonable as an "obvious" behavior, but is slightly surprising given how most other Ammonite scripts cache compiled things locally for performance. 
    - This means that if the number of `import $url`s grows, your script would spend significant amounts of time fetching stuff over HTTP, even after being compiled, which I would find surprising. 
    - On the other hand, I do not see any way to avoid making a HTTP request if we wish to invalidate any caches when the script at the URL changes.
    - On the other *other* hand, perhaps we could cache things, and leave it up to the user to only link to stable URLs? For example, versioned links to Gists would be fine `https://gist.githubusercontent.com/lihaoyi/7e6c429bf0c18465479450c0a351e8f5/raw/e6f33304a0fbe47cfa25176e8371487a91a1f688/debug.scala`, since if the file content change the URL changes

    - Regardless of whether-or-not we end up landing this, and if we do land it which-solution-we-decide-to-go-with, this definitely deserves some discussion and serious consideration as there is (to me) no "obvious" best solution

- This would need docs added to http://ammonite.io/#UnstableChangelog if we decide to land it, which would be moved into http://ammonite.io/#MagicImports on the next stable release

Atry:
1. Relative URL makes sense, though I guess we can leave it to next version after people get used to import absolute URLs.
1. Be aware that HTTP has lots of mechanisms and headers around caching. If I understand correctly, JRE respects these headers and will cache public resource automatically.

Atry:
Unstable changelog added

lihaoyi:
FWIW I've been traveling and haven't had the solid blocks of time & concentration necessary to properly take a look at this. I haven't forgotten, but may be another week or two before my schedule frees up.

Atry:
Any update?

lihaoyi:
Nope, still traveling

Atry:
Any progress?

lihaoyi:
Sorry, been away for a while.

I'm not totally happy with this, but I'll merge it anyway and then we can discuss further changes from there. Perhaps we may want to start caching HTTP scripts locally and forcing people to include version/git-SHAs to force updates, or maybe we'll decide the whole thing is unworkable, or maybe it'll morph into some kind of ad-hoc package management system.

Anyway, let's get this in so you can start using it and we can cross those bridges when we come to them

Atry:
> Perhaps we may want to start caching HTTP scripts locally and forcing people to include version/git-SHAs to force updates
It's called ETag. By default JRE will use cache according to ETag:

https://docs.oracle.com/javase/7/docs/api/java/net/URLConnection.html#setUseCaches(boolean)

> The default value comes from DefaultUseCaches, which defaults to true.

PR #656, opened 2017-07-05
Clipboard operations: xclip or pbcopy/pbpaste style operations.

pfcoperez:
One of production booster when working with terminals is the ability to read and write from and to the system clipboard. 

This will be extremely useful to run Scala programs on data brought from selected text. An example is selecting part of Ammonite README file on GitHub:

Selected text

And counting words in the selected text:

```
val words = clipboard.read.split("\n").flatMap(_.split(" ").map(_.toLowerCase))

val wordCount = (Map.empty[String, Int] /: words) { (count, word) =>
    count + (word -> (count.getOrElse(word, 0)+1)) 
}

```

Result

I am open to discuss & change implementation details. I have some doubts:

- I  didn't consider this should be under `FileOps` so I created `ClipboardOps`
- If the idea is welcomed further functionality could be added.
- The code could be more abstract and generic.
- ...

### DIFF
--- a/amm/repl/src/main/scala/ammonite/repl/ApiImpls.scala
+++ b/amm/repl/src/main/scala/ammonite/repl/ApiImpls.scala
@@ -1,9 +1,12 @@
 package ammonite.repl
 
-
+import ammonite.ops.Internals
 import ammonite.runtime._
 import ammonite.util.Util._
 import ammonite.util._
+
+import java.awt.Toolkit
+import java.awt.datatransfer.{DataFlavor, StringSelection}
 
 import scala.collection.mutable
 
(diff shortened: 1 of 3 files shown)


pfcoperez:
@lihaoyi Are you opposed to this kind of functionality or is it being provided in other PR? If the answer is no to both of this questions, I am open to following a different approach.

lihaoyi:
Haha i just closed it because i assumed it was stale. I think it looks good to me, but how about putting it in the `ReplApi` rather than in `ammonite.ops`? Then that i'll be happy to merge it

lihaoyi:
lgtm

PR #667, opened 2017-07-19
Use Main instead of repl.Repl in Sshd. This is a temporary fix for #659

### LINKED ISSUE #659: NullPointerException in sshd remote repl
I'm trying to embed Ammonite, but got the following error (note that it "works" correctly with bad inputs).

```
Password authentication
Password:
Welcome to the Ammonite Repl 1.0.0
(Scala 2.12.1 Java 1.8.0_131)
If you like Ammonite, please support our development at www.patreon.com/lihaoyi
@ 1
java.lang.NullPointerException


@ x
cmd1.sc:1: not found: value x
val res1 = x
           ^
Compilation Failed
@
```
There is nothing in the log except for a line about ssh client connecting.

Any pointer on how to debug this? I would be happy to dive in and diagnose it.

ubolonton:
Fixes #659.

`Main` contains a lot of initialization code, without which the wrapped `repl.Repl` is not usable. In this case, initializing `repl.ReplBridge.value`.

### DIFF
--- a/sshd/src/main/scala/ammonite/sshd/SshdRepl.scala
+++ b/sshd/src/main/scala/ammonite/sshd/SshdRepl.scala
@@ -3,11 +3,11 @@ package ammonite.sshd
 
 import java.io.{InputStream, OutputStream, PrintStream}
 
+import ammonite.Main
 import ammonite.ops.Path
-import ammonite.sshd.util.Environment
-import ammonite.util.{Bind, Name, PredefInfo}
 import ammonite.runtime.Storage
-import ammonite.repl.Repl
+import ammonite.sshd.util.Environment
+import ammonite.util.{Bind, Colors}
 
 /**
  * An ssh server which serves ammonite repl as it's shell channel.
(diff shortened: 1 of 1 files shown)


lihaoyi:
What's temporary about this fix?

ngthanhtrung:
He may want to say using `Main` here is not clean. Code looks good to me as a temporary fix.

ubolonton:
Yeah, I wanted to reorganize the code a bit to make it clearer that `Repl` is an implementation detail, and `Main` should be used instead. That can perhaps be left for later though.

Another thing is that there's currently no test suite for the fix.

lihaoyi:
Bah, I guess that's inevitable since we disabled those tests due to flakiness =/ https://github.com/lihaoyi/Ammonite/blob/master/sshd/src/test/scala/ammonite/sshd/SshdReplTest.scala#L14

Otherwise looks good I guess!

lihaoyi:
w.r.t. using `Main` vs using Repl, I have no strong preference; `Main` originally did much less than it does now, but since it's grown in scope migrating the SshdRepl over to `Main` seems OK. If this causes problems elsewhere we can fix those as they turn up

lihaoyi:
Also, thanks!

lihaoyi:
You should be able to use the automatically published unstable release once this goes through CI

Issue #663, opened 2017-07-16
Ammonite does not properly backtick a value with `/*` in its name


fstoqnov:
After importing Media Range `*/*` any further commands do not get processed and the entire console needs to be restarted for ammonite to work again - I am running ammonite on Windows 10 in the default bash console
```scala
$ amm 
Loading... 
Welcome to the Ammonite Repl 1.0.0 (Scala 2.12.2 Java 1.8.0_131) 
If you like Ammonite, please support our development at www.patreon.com/lihaoyi 
@ import $ivy.`com.typesafe.akka::akka-http:10.0.9` 
import $ivy.$ 
@ import akka.http.scaladsl.model.MediaRanges.`image/*` 
import akka.http.scaladsl.model.MediaRanges.`image/*` 
@ 2 + 2 
res2: Int = 4 
@import akka.http.scaladsl.model.MediaRanges.`*/*` 
import akka.http.scaladsl.model.MediaRanges.`*/*`
@ 2 + 2 
cmd4.sc:164: unclosed comment  
 */*   
^ 
cmd4.sc:180:  '}' expected but eof found. 
}  
^ 
Compilation Failed
```

bblfish:
I can confirm that the same problem occurs on OSX terminal.  

It looks like it cannot even be imported as `_`. 

```scala
$ amm
Loading...
Welcome to the Ammonite Repl 1.0.0
(Scala 2.12.2 Java 1.8.0_131)
If you like Ammonite, please support our development at www.patreon.com/lihaoyi
hjs-ammonite@ import $ivy.`com.typesafe.akka::akka-http:10.0.9`
import $ivy.$
hjs-ammonite@ import akka.http.scaladsl.model.MediaRanges._
import akka.http.scaladsl.model.MediaRanges._
hjs-ammonite@ 2+2
cmd2.sc:164: unclosed comment
  */*,
  ^
cmd2.sc:183: '}' expected but eof found.
}
 ^
Compilation Failed
```

bblfish:
There is a workaround

```scala
import akka.http.scaladsl.model.MediaRanges
val star_star = akka.http.javadsl.model.MediaRanges.ALL.asInstanceOf[MediaRanges.PredefinedMediaRange]
akka.http.scaladsl.model.headers.Accept(star_star)
```

lihaoyi:
Minimal-er repro
```scala
@ def `*/*` = 1

defined function */*

@ 1
cmd1.sc:61: unclosed comment
[...]

PR #668, opened 2017-07-19
fix for Ammonite does not properly backtick a value #663

justinhj:
works

### DIFF
--- a/amm/src/test/scala/ammonite/session/EvaluatorTests.scala
+++ b/amm/src/test/scala/ammonite/session/EvaluatorTests.scala
@@ -287,6 +287,37 @@ object EvaluatorTests extends TestSuite{
 
         @ `l o l`
         res10: Int = 123
+
+        @ import ` `.{` ` => `//`}
+        
+        @ `//`
+        res12: Int = 123
+
+        @ import ` `.{` ` => `Something then //`}
+
+        @ `Something then //`
+        res14: Int = 123
+
+        @ import ` `.{` ` => `// and then`}
+
+        @ `// and then`
+        res16: Int = 123
+
+        @ import ` `.{` ` => `/*`}
+
+        @ `/*`
+        res18: Int = 123
+
+        @ import ` `.{` ` => `/*/`}
+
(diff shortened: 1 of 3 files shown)


lihaoyi:
Looks good. Three  things to do before landing:

- Add something to the changelog here https://github.com/lihaoyi/Ammonite/blob/master/readme/Footer.scalatex#L1400

- Add this case to the unit test that's meant to test these case (https://github.com/lihaoyi/Ammonite/blob/master/amm/src/test/scala/ammonite/session/EvaluatorTests.scala#L289), make sure the old code fails your updated test, and the new code passes

- Also add a fix for operators containing `//`, which currently don't work for presumably the same reason:

```scala
@ val `//` = 1
cmd2.sc:77: ')' expected but '}' found.
      ) }
        ^
Compilation Failed
```

justinhj:
Cool will address these issues

amm/src/test/scala/ammonite/session/EvaluatorTests.scala
```diff
@@ -287,6 +287,37 @@ object EvaluatorTests extends TestSuite{
...
+
+        @ `Something then //`
+        res14: Int = 123
+
+        @ import ` `.{` ` => `// and then`}
+
+        @ `// and then`
+        res16: Int = 123
+
+        @ import ` `.{` ` => `/*`}
```
lihaoyi: Could you convert one of these to 

```scala
val `/*` = 123
```

Just to make sure that works too

justinhj:
I have tests for `/*` and `*/`

lihaoyi:
Yeah I know you have tests for `/*`, I want you to also explicitly test the 

```scala
val `/*` = ...
```

syntax because it fails with a different error than the `import` command, and we should make sure that the `val` assignment works

justinhj:
Ah i see, will do

lihaoyi:
Looks good, thanks!

lihaoyi:
Fixed in https://github.com/lihaoyi/Ammonite/issues/663

PR #674, opened 2017-07-30
add missing dollar sign in exec import

joprice:
As far as I can tell, a dollar was intended here.


lihaoyi:
No idea why appveyor is unhappy but it looks fine

PR #677, opened 2017-08-08
Dependency updates

golem131:
sbt 0.13.15 -> 0.13.16
sbt-assembly 0.14.3 -> 0.14.5
sbt-coursier 1.0.0-M15 -> 1.0.0-RC9
scopt 3.5.0 -> 3.6.0
coursier and coursier-cache 1.0.0-RC8 -> 1.0.0-RC9
jline 2.14.3 -> 2.14.5
javaparser-core 3.2.5 -> 3.3.0
bcprov-jdk15on 1.56 -> 1.57
scalacheck 1.12.6 -> 1.13.5

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -260,8 +260,8 @@ lazy val ammRepl = project
     crossVersion := CrossVersion.full,
     name := "ammonite-repl",
     libraryDependencies ++= Seq(
-      "jline" % "jline" % "2.14.5",
-      "com.github.javaparser" % "javaparser-core" % "3.3.0"
+      "jline" % "jline" % "2.14.3",
+      "com.github.javaparser" % "javaparser-core" % "3.2.6"
     ),
     unmanagedSourceDirectories in Compile ++= {
       if (Set("2.12", "2.11").contains(scalaBinaryVersion.value))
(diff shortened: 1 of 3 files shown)


build.sbt
```diff
@@ -326,13 +326,13 @@ lazy val sshd = project
       name := "ammonite-sshd",
       libraryDependencies ++= Seq(
         // sshd-core 1.3.0 requires java8
-        "org.apache.sshd" % "sshd-core" % "1.2.0",
-        "org.bouncycastle" % "bcprov-jdk15on" % "1.56",
+        "org.apache.sshd" % "sshd-core" % "1.6.0",
```
piotrkwiecinski: I reckon this was left as 1.2 on purpose to support java < 8

golem131: Java 6 EOL more than 4 years (https://developers.slashdot.org/story/13/06/20/1819245/java-6-eold-by-oracle).
May be it's time to drop support of Java 6?

lihaoyi: I'd say we should keep it, unless there's some really-truly pressing reason we need to upgrade `sshd-core`. I don't have numbers but wouldn't be surprised if there were double-digit % of Ammonite users still on Java6

piotrkwiecinski:
@golem131 could you downgrade the ```sshd-core``` back to ```1.2.0``` and give it another go? We can upgrade rest of the dependencies.

lihaoyi:
Closing due to inactivity

Issue #679, opened 2017-08-10
ls.rec fails if a single file is inaccesible


DavidPerezIngeniero:
When doing `ls.rec` over `/home/david`, I get this exception:

```
java.nio.file.AccessDeniedException: /home/david/.cache/dconf
        at sun.nio.fs.UnixException.translateToIOException(UnixException.java:84)
        at sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:102)
        at sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:107)
        at sun.nio.fs.UnixFileSystemProvider.newDirectoryStream(UnixFileSystemProvider.java:427)
        at java.nio.file.Files.newDirectoryStream(Files.java:457)
        at ammonite.ops.ls$iter$.$anonfun$apply$6(FileOps.scala:210)
        at geny.Generator$SelfClosing.generate(Generator.scala:219)
        at geny.Generator$Filtered.generate(Generator.scala:269)
        at geny.Generator$FlatMapped.generate(Generator.scala:284)
        at geny.Generator$Concat.generate(Generator.scala:236)
        at geny.Generator$Mapped.generate(Generator.scala:276)
        at geny.Generator$FlatMapped.$anonfun$generate$5(Generator.scala:285)
        at geny.Generator$Filtered.$anonfun$generate$3(Generator.scala:269)
        at geny.Generator$SelfClosing.generate(Generator.scala:222)
        at geny.Generator$Filtered.generate(Generator.scala:269)
        at geny.Generator$FlatMapped.generate(Generator.scala:284)
        at geny.Generator$Concat.generate(Generator.scala:236)
        at geny.Generator$Mapped.generate(Generator.scala:276)
        at geny.Generator$FlatMapped.$anonfun$generate$5(Generator.scala:285)
        at geny.Generator$Filtered.$anonfun$generate$3(Generator.scala:269)
        at geny.Generator$SelfClosing.generate(Generator.scala:222)
        at geny.Generator$Filtered.generate(Generator.scala:269)
        at geny.Generator$FlatMapped.generate(Generator.scala:284)
        at geny.Generator$Mapped.generate(Generator.scala:276)
        at geny.Generator.foreach(Generator.scala:48)
        at geny.Generator.foreach$(Generator.scala:48)
        at geny.Generator$Mapped.foreach(Generator.scala:274)
        at geny.Generator.toBuffer(Generator.scala:120)
[...]

PR #682, opened 2017-08-12
ls.rec handle inaccessible files

piotrkwiecinski:
Attempt to fix https://github.com/lihaoyi/Ammonite/issues/679

For now without tests, but I'll try to figure our reliable way to verify it on travis.

1) Bare minimum would be to check if directory with no access doesn't throw Exception and we get empty LsSeq
2) Better would be to have directory with at least file inaccessible file and check if we get non empty LsSeq

Maybe we can use /tmp

### DIFF
--- a/ops/src/main/scala/ammonite/ops/FileOps.scala
+++ b/ops/src/main/scala/ammonite/ops/FileOps.scala
@@ -207,9 +207,13 @@ object ls extends StreamableOp1[Path, Path, LsSeq] with ImplicitOp[LsSeq]{
 
   object iter extends (Path => geny.Generator[Path]){
     def apply(arg: Path) = geny.Generator.selfClosing{
-      val dirStream = Files.newDirectoryStream(arg.toNIO)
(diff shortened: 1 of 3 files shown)


lihaoyi:
Looks like the right direction; if you can figure out how to get a test for this that would be very valuable. If not, let me know and I'll be happy to merge it as is (after rebasing)

piotrkwiecinski:
@lihaoyi for now I don't see straightforward way to test it on both appveyor and travis. Even with travis to do it properly we would need to use sudo and it would increase build time in general.

lihaoyi:
Could we test it on travis only? I don't mind leaving it untested on appveyor

lihaoyi:
If not say the word and I'll merge it as is

piotrkwiecinski:
@lihaoyi I've added test based on /var/run. Hopefully it won't be flaky.

lihaoyi:
Thanks!

PR #684, opened 2017-08-12
Small clean up in ammonite.ops

piotrkwiecinski:
- Replace collection.JavaConversions with collection.JavaConverters
- Remove unused import statements
- Replace .length > 0 by .nonEmpty
- Replace getOrElse(null) with .orNull on option
- Remove `new` on case classes


lihaoyi:
Thanks!

PR #691, opened 2017-08-23
Fix #688 Clarify % operator documentation for non-zero return-code

### LINKED ISSUE #688: % operator throws exception if result code != 0
There is a bug in either the documentation or the code.

```
 %grep("kk", "/etc/hosts") 
ammonite.ops.InteractiveShelloutException
  ammonite.ops.Shellout$.executeInteractive(Shellout.scala:53)
  ammonite.ops.Shellout$.$anonfun$$percent$1(Shellout.scala:14)
  ammonite.ops.Shellout$.$anonfun$$percent$1$adapted(Shellout.scala:14)
  ammonite.ops.Command.applyDynamic(Shellout.scala:111)
  ammonite.$sess.cmd4$.<init>(cmd4.sc:1)
  ammonite.$sess.cmd4$.<clinit>(cmd4.sc)

@ %grep("local", "/etc/hosts") 
127.0.0.1        oyente02.fccma.com oyente02 localhost localhost.localdomain localhost4 localhost4.localdomain4
```

The documentation tells:

In short, % lets you run a command as you would in bash, and dumps the output to standard-out in a similar way, **returning the return-code**. 

I expect not to throw an exception when the return code isn't 0.

Instead, the %% command should (and does) throw an exception.

Tested with Ammonite 1.0.0

piotrkwiecinski:
Fix for https://github.com/lihaoyi/Ammonite/issues/688

### DIFF
--- a/readme/Ops.scalatex
+++ b/readme/Ops.scalatex
@@ -326,7 +326,9 @@
         ...
   
       @p
-        In short, @hl.scala{%} lets you run a command as you would in bash, and dumps the output to standard-out in a similar way, returning the return-code. This lets you run @code{git} commands, edit files via @code{vim}, open @code{ssh} sessions or even start @code{SBT} or @code{Python} shells right from your Scala REPL!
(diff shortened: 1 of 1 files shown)


lihaoyi:
Thanks!

PR #697, opened 2017-09-10
Add output stream for results

alexarchambault:
This PR adds an extra field `resultStream` to `Printer`, and uses it solely to print the "result" of the user code (like `n: Int = 2` if the user entered `val n = 1 + 1`).

If one uses Ammonite via its API, that allows to get specifically these results, rather than whatever else can be printed via `outStream`. It's useful when these results are then treated in a specific way.


lihaoyi:
Looks good to me, thanks!

PR #699, opened 2017-09-17
Ensure repl module can be used in a standalone way

alexarchambault:
This PR makes sure that one can use Ammonite via its API, solely by depending on its repl module.

To illustrate that, it moves [`TestRepl`](https://github.com/lihaoyi/Ammonite/blob/79f3f8b32e70245842ffa8b12d9e5939d0cd286e/amm/src/test/scala/ammonite/TestRepl.scala), along with most tests, to the repl module. To make it work, it moves a few things from the main amm module to the repl module too (mainly [stubs](https://github.com/lihaoyi/Ammonite/blob/79f3f8b32e70245842ffa8b12d9e5939d0cd286e/amm/src/main/scala/ammonite/Stubs.scala) and a [bunch of helpers](https://github.com/lihaoyi/Ammonite/blob/79f3f8b32e70245842ffa8b12d9e5939d0cd286e/amm/src/main/scala/ammonite/main/Defaults.scala)).

Lastly, to emphasize that the repl module can be used purely programmatically (without terminal interaction), it removes its dependencies towards jline and the terminal module. For that, it mainly moves the `FrontEnd` implementations to the main amm module, and passes a terminal width as argument here or there (rather than having it be fetched via the terminal module from repl).

Don't hesitate to tell me if ever the last point (which corresponds to the last commit) hinders reviewing. It logically follows what this PR does, but it could be sent in a separate PR after this one.


amm/repl/src/test/scala/ammonite/interp/AutocompleteTests.scala
```diff
@@ -48,7 +48,7 @@ object AutocompleteTests extends TestSuite{
 
       'import - checking{ complete =>
         complete("""import <caret>""", Set("java", "javax", "scala", "javassist") -- _)
-        complete("""import j<caret>""", Set("java", "javax", "jline", "jawn", "javassist") -- _)
+        complete("""import j<caret>""", Set("java", "javax", "jawn", "javassist") -- _)
```
alexarchambault: (jline is no more accessible from this module.)

alexarchambault:
The failing test is fine locally :/

lihaoyi:
@alexarchambault could you drop by the gitter channel? This is probably easier to discuss real-time rather than going back and forth over the issue tracker

alexarchambault:
@lihaoyi I saw your new messages on Gitter... I'm really busy this week, but I should be able to catch up on Friday I think.

alexarchambault:
@lihaoyi I updated the PR, it's now down to its very substance ;)

The only thing that bugs me is that nothing prevents regressions for the use this PR allows (creating an `Interpreter` with a `ReplAPI` bridge, via the `repl` module). That's why I initially moved `TestRepl` along with most tests in the `repl` module too.

lihaoyi:
Sorry, been without 3G last few days and comcast apartment wifi was borked. Will look at this when things are more stable...

lihaoyi:
Sorry for the delay. I still don't have wifi, but my LTE is back up. Looks good to me!

lihaoyi:
Feel free to send the test movements in another PR; I agree that it is a good thing to do, just that removing them from this PR made it so much easier for me to understand what you were doing.

If you do send a PR to shift the tests and ensure we do not regress in this behavior, don't forget to update the readme (https://github.com/lihaoyi/Ammonite#manual-testing) to mention that we now have tests split over multiple submodules in the `amm/` folder

alexarchambault:
Will do, thanks a lot!

PR #709, opened 2017-10-16
Move most tests to repl module

alexarchambault:
Follow-up of https://github.com/lihaoyi/Ammonite/pull/699.

Only the tests that can be moved with no changes are moved here.

`ScriptTests` could be moved in large part, but not in its entirety (because a test calls `Scripts.runScript` from the `amm` module). It can be split then moved in another PR.


build.sbt
```diff
@@ -261,7 +250,8 @@ lazy val ammRepl = project
     name := "ammonite-repl",
     libraryDependencies ++= Seq(
       "jline" % "jline" % "2.14.3",
-      "com.github.javaparser" % "javaparser-core" % "3.2.5"
+      "com.github.javaparser" % "javaparser-core" % "3.2.5",
+      "com.github.scopt" %% "scopt" % "3.5.0" % Test
```
lihaoyi: What's the reason for this addition?

build.sbt
```diff
@@ -132,17 +132,6 @@ lazy val amm = project
     // classpaths, so that the `source` macro can find their sources and
     // highlight/display them.
```
lihaoyi: This comment used to apply to the code you moved, which is now a bit far away. Perhaps you could split the comment in two and have the relevant text near the relevant piece of code?

alexarchambault: Yep, fixed that in the third commit.

build.sbt
alexarchambault: scopt is used in [`SourceTests`](https://github.com/lihaoyi/Ammonite/blob/73a874173cd337f953a3edc9fb8cb96556638fdd/amm/src/test/scala/ammonite/unit/SourceTests.scala#L52) (moved to the `repl` module), and is not around in the `repl` module without that.

lihaoyi:
Thank @alexarchambault! Let me know if you want me to kick off a numbered release; it's as simple as pushing a tag and CI does the rest

alexarchambault:
@lihaoyi I'm in no hurry for a release for now, I'd like to send 2 or 3 more PRs before that I think (to finally close the gap with my [ammonium](https://github.com/jupyter-scala/ammonium)).

alexarchambault:
Hello @Ammonite-Bot !

lihaoyi:
Now that 2.12.4 is out I'm gonna go tag off a release to cross-build for it

Issue #665, opened 2017-07-18
Watch shellout stream


NeQuissimus:
I would like to "watch" continuous output from a command, essentially something like the following:

```scala
import ammonite.ops._
import ammonite.ops.ImplicitWd._ 
%%("docker", "logs", "-f", "kafka").chunks.foreach(println)
```

However, `Shellout.executeStream` only passes back a result once the command has finished.
Is there currently an easy way for me to achieve what I am trying?

Thank you

lihaoyi:
Basically subsumed by https://github.com/lihaoyi/Ammonite/pull/717

Issue #367, opened 2016-03-27
Ammonite should be able to pipe into programs


dmacias:
Ammonite should be able to pipe stdout to another program.  I would think the intent would be something to the effect of:

`home/"file1.xml" |% less`

Though the same can be achieved with:

`home/"file1.xml" |> (%less _)`

it doesn't account for programs like pbcopy (mac) that only works via stdin.

dmacias:
@lihaoyi suggested:

> currently the %% shellouts all return a CommandResult which just encapsulates the output. we could make it return something richer that allows two-way communication with the subprocess

lihaoyi:
Ideally, we'd have the returned `CommandResult`, instead of being a frozen representation of the total output of the process, be a live `Process` object we can interact with: reading output, sending input, reading more output, killing, signaling, whatever. We could even leave the existing methods to "read all output" or "read all error" to make the common case convenient, while still leaving open the ability to feed in input and otherwise interact with the running process.

This will require a significant amount of rewriting of `Shellout.scala` but I don't think it's hard in-principle

lihaoyi:
Moot since the EOL of ammonite-ops. OS-Lib has its own way of doing this, where you can specify a `os.proc().call(stdin = ...)` to pipe data between subprocesses or other streaming data sources

PR #717, opened 2017-11-10
[WIP] Interactive processes

Krever:
This is a draft of a more interactive way of spawning processes. The goal is to provide a more convenient API to spawn a process and communicate with it. 

This implementation is just a first step and I would like to hear some voices on how it should look like, what to change and what to add. It would probably be good enough for my use case but not for general availability :)

I wanted to use `Internals.Writable` for `write` method but got into compile cycles and didn't want to waste time on solving it. I will do that if you agree that this is a correct thing to do.

Initially, I was thinking about something like `await` that would convert `CommandProcess` into `CommandResult` but I'm no longer sure it will give much benefit over current implementation.

This potentially solves #367  and #665 

I left following TODOs until we have some stable API

- [ ] inline comments
- [ ] documentation
- [ ] more tests


lihaoyi:
subsumed by https://github.com/lihaoyi/os-lib

PR #713, opened 2017-10-28
Jitpack over coursier rc10

emanresusername:
is this better as a pr or [an issue](https://github.com/lihaoyi/Ammonite/issues/712) 🤔

PR #725, opened 2017-12-14
Update coursier to 1.0.0

piotrkwiecinski:
Latest stable release or coursier

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -205,8 +205,8 @@ lazy val ammRuntime = project
 
     name := "ammonite-runtime",
     libraryDependencies ++= Seq(
-      "io.get-coursier" %% "coursier" % "1.0.0-RC10",
-      "io.get-coursier" %% "coursier-cache" % "1.0.0-RC10",
+      "io.get-coursier" %% "coursier" % "1.0.0",
+      "io.get-coursier" %% "coursier-cache" % "1.0.0",
       "org.scalaj" %% "scalaj-http" % "2.3.0"
     )
   )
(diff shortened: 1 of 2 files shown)


alexarchambault:
@piotrkwiecinski Note that [this line](https://github.com/lihaoyi/Ammonite/blob/master/project/plugins.sbt#L14) can now be removed (thanks to https://github.com/coursier/coursier/pull/712).

lihaoyi:
Thanks!

emanresusername:
@lihaoyi @alexarchambault any idea why [this test fails](https://github.com/lihaoyi/Ammonite/pull/713/files#diff-ba4b71b2d310a9c84efade59dff114d4R331) on `RC9` and up? or can anybody reproduce or is something just wrong with my computers? (tried on two)
is there some cached config somewhere (aside from `~/{.ivy2,.sbt,.coursier}` that might be interfering (blew all those away and it still breaks for me) 😭

alexarchambault:
@emanresusername I think that originates from the "optional" artifacts, that were added in some RCs of coursier. Namely, one should get the artifacts via [the method accepting an extra arg](https://github.com/coursier/coursier/blob/b21042a42a6b2718ec1116f28f76dd769112c6aa/core/shared/src/main/scala/coursier/core/Resolution.scala#L1048), passing it `true`, then ignore not found errors for artifacts that are [optional](https://github.com/coursier/coursier/blob/b21042a42a6b2718ec1116f28f76dd769112c6aa/core/shared/src/main/scala/coursier/core/Definitions.scala#L206) (but not ignoring the other kinds of errors).

That way, optional artifacts would be loaded in ammonite, and that test case should be fine.

PR #726, opened 2017-12-15
Fix Appveyor build

piotrkwiecinski:
* Update sbt version to 0.13.15
* Use chocolatey to install sbt  - chocolatey is package manager and should allow for straightforward updates to newer versions.
* Make sure all tests pass - fullInfo.group can be null on windows
* Updated cache configuration including new Coursier location

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -24,7 +24,9 @@ test_script:
 
 cache:
   - C:\sbt\
-  - C:\Users\appveyor\.m2
-  - C:\Users\appveyor\.ivy2\cache
+  - '%USERPROFILE%\.ivy2\cache'
+  - '%USERPROFILE%\.m2'
+  - '%USERPROFILE%\.sbt'    
+  - '%LOCALAPPDATA%\Coursier\cache'
   - C:\ProgramData\chocolatey\lib -> appveyor.yml
   - C:\ProgramData\chocolatey\bin -> appveyor.yml
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -8,16 +8,8 @@ environment:
   SBT_OPTS: -Dfile.encoding=UTF8
   JAVA_OPTS: -Dfile.encoding=UTF8
 install:
-  - ps: |
-      Add-Type -AssemblyName System.IO.Compression.FileSystem
-      if (!(Test-Path -Path "C:\sbt" )) {
-        (new-object System.Net.WebClient).DownloadFile(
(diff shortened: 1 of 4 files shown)


appveyor.yml
```diff
@@ -30,3 +22,9 @@ test_script:
   - sbt ++2.11.11 amm/test
   - sbt ++2.11.11 integration/test
 
+cache:
+  - C:\sbt\
+  - C:\Users\appveyor\.m2
```
piotrkwiecinski: @alexarchambault any recommendations for coursier cache on windows ? As there is a chance I'm using incorrect dir here.

piotrkwiecinski:
@lihaoyi it should be ready to review.

lihaoyi:
@piotrkwiecinski I'm having a hard time distinguishing random formatting tweaks and the actual important bits, and why they are important. Could you revert any unnecessary changes and describe any important changes in the PR description?

piotrkwiecinski:
@lihaoyi ok I've removed all distractions.

ops/src/test/scala/test/ammonite/ops/ExampleTests.scala
```diff
@@ -114,8 +114,9 @@ object ExampleTests extends TestSuite{
       val fullInfo = stat.full(wd/"file1.txt")
       fullInfo.ctime: FileTime
       fullInfo.atime: FileTime
-      fullInfo.group: GroupPrincipal
-
+      if (Unix()) {
```
lihaoyi: @piotrkwiecinski is this really necessary? We are just doing a type-ascription, and not a `asInstanceOf`, so even if `group` is null it shouldn't fail. The only way this can fail is by failing to compile; the runtime value of `fullInfo.group` isn't ever used

piotrkwiecinski: @lihaoyi I would say it's more like a temporary workaround. Issue comes from https://github.com/lihaoyi/utest/blob/master/utest/shared/src/main/scala/utest/framework/Formatter.scala#L19

It fails on my windows machine, so I'll play around with it more.

lihaoyi:
Hmm ok, I guess a better workaround for that could be to put a `()` at the end of the block so the `null` doesn't get returned

piotrkwiecinski:
@lihaoyi it's green now. Should the ```null``` somehow be handled properly in utest in formatter?

lihaoyi:
@piotrkwiecinski yeah it should, but this looks good for now

lihaoyi:
Feel free to send a PR upstream to uTest

PR #727, opened 2017-12-17
[WIP]Remove tput from terminal dependencies

piotrkwiecinski:
This should allow to create thinner docker boxes. I've checked it works for me on mac and linux (debian).

I'm not sure it these all dependencies to remove ncurses as dependency, but we'll be closer to creating alpine docker image just by adding java binary.


piotrkwiecinski:
I'll have to do more debugging as running docs command on local env works without any issues.

lihaoyi:
Closing due to inactivity

PR #732, opened 2017-12-25
Remove bash terminal dependencies

piotrkwiecinski:
@lihaoyi this one should be ready to review. Basic sh is enough for the job.

I'll try to get rid of tput in separate PR as it needs more debugging.


piotrkwiecinski:
@lihaoyi I've minimised amount of changes here too.

lihaoyi:
Looks good to me

PR #734, opened 2018-01-02
Report predef initialization error during tests

alexarchambault:
This is useful while debugging or changing things in the internals of Ammonite. Previously, errors during predefs where trapped. Now they are loudly reported.


lihaoyi:
Looks good to me

PR #735, opened 2018-01-02
Discard load command result

alexarchambault:
Rather than failing at runtime with ClassNotFoundException.

This makes the last command in the following fail at compile time, rather than at runtime with a ClassNotFoundException:
```scala
@ repl.sess.save("foo")

@ val a = repl.sess.load("foo")

@ a
```

Although this looks like a detail here, this is required for classwrapping (upcoming PR).

It works by "freezing" the current frame after loading a session, preventing it to accept extra imports. That way, no imports from the command that loaded the session are added, and we can't run into ClassNotFoundException about it.

### DIFF
--- a/amm/runtime/src/main/scala/ammonite/runtime/ClassLoaders.scala
+++ b/amm/runtime/src/main/scala/ammonite/runtime/ClassLoaders.scala
@@ -25,22 +25,38 @@ class Frame(val classloader: SpecialClassLoader,
             val pluginClassloader: SpecialClassLoader,
             private[this] var imports0: Imports,
             private[this] var classpath0: Seq[java.io.File]){
+  private var frozen0 = false
+  def frozen = frozen0
+  def freeze(): Unit = {
+    /*
+     * Once frozen, a frame won't accept new imports or classpath elements.
+     * This is useful in commands that load / save sessions, whose result
+     * doesn't need to be kept in the frame.
+     */
+    frozen0 = true
+  }
   private[this] var version0: Int = 0
   def version = version0
   def imports = imports0
   def classpath = classpath0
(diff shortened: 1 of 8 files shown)


amm/runtime/src/main/scala/ammonite/runtime/ClassLoaders.scala
```diff
@@ -25,22 +25,33 @@ class Frame(val classloader: SpecialClassLoader,
             val pluginClassloader: SpecialClassLoader,
             private[this] var imports0: Imports,
             private[this] var classpath0: Seq[java.io.File]){
+  private var frozen0 = false
+  def frozen = frozen0
+  def freeze(): Unit = {
```
lihaoyi: Can you add a comment here explaining what the whole point of this freezing thing is? That way the knowledge will remain in the codebase even after further file edits/moves mangles the history and makes the commit message hard to find

alexarchambault: yep

lihaoyi:
@alexarchambault the line length checker is mad

alexarchambault:
Added a somehow clearer test (https://github.com/lihaoyi/Ammonite/pull/735/commits/31aa3f1933f508defcdaaccd83eb7be064b5dc77#diff-9a0e0675ceb797d13226a7b70eebe6beR245), and an extra commit (https://github.com/lihaoyi/Ammonite/pull/735/commits/0d80e77c4ee224b2a03181ae1630f8e60470e902).

alexarchambault:
(Looks like a CI transient error again.)

lihaoyi:
Looks good to me, thanks alex!

PR #750, opened 2018-02-14
Empty welcome banner should not be printed

ocadaruma:
Current behavior:

```
$ amm -b ""
Loading...

@
```

I think following is desired behavior.

```
$ amm/target/amm -b "" 
Loading...
@
```


lihaoyi:
Looks good

PR #755, opened 2018-02-22
autocomplete: blacklisted anything that comes from Any

scalway:
We decide to filter out all methods that comes with Any.

There was two options. One to edit Pressy.scala:57:
```scala
      blacklist(s.fullNameAsName('.').decoded) ||
      s.fullNameAsName('.').decode.startsWith("scala.Any.") || //this line
      s.isImplicit ||
```
but then when people'll press tab then it'll never complete into full name (eg ".toString") even if it is the only option.

And changes mede here.
They'll be completed when it is the only option and show also definitions after second `Tab`.


scalway:
demn... :100:

lihaoyi:
looks good

Issue #675, opened 2017-08-03
amm throws NPE on OpenJDK 9


PaulAtBanno:
The amm runtime Classpath system attempts to get the "sun.boot.class.path" property, which doesn't exist in OpenJDK 9, throwing a NullPointerException. I'm not sure what the appropriate workaround is, sorry!

lihaoyi:
lol I don't know what the solution is either. Hopefully someone who knows something about Java 9 can chime in and enlighten us...

PaulAtBanno:
OK, so as you might expect, especially if you've ever worked with OSGi, this has to do with the module system. In JRE/JDK 9, there is a new file system provider, "jrtfs," that apparently provides access to the classes used by the JRE/JDK itself in a principled way. [This post](http://www.jroller.com/andyl/entry/say_hello_to_jep_220) has some further information and [this code](http://www195.pair.com/mik3hall/JRTLister.java) is an example of using the file system.

There is also a [backport of jrtfs to Java 7](https://github.com/forax/jigsaw-jrtfs), but I don't know how useful that would be to Ammonite.

Hope this helps!

lihaoyi:
Still doesn't work on Java 9. On the other hand, I couldn't get SBT to work on Java 9 either, so I'll probably wait until at least that is supported so Ammonite can be built

pshirshov:
sbt 1.0.3 / scala 2.12.4 are working fine on j9

gabrieljones:
`System.getProperty("sun.boot.class.path")` @ [Classpath.scala#L29](https://github.com/lihaoyi/Ammonite/blob/1.0.3/amm/runtime/src/main/scala/ammonite/runtime/Classpath.scala#L29) returns null

francisdb:
This can be widened to OracleJDK as shown in #739

francisdb:
I guess this can be closed?

PR #761, opened 2018-03-07
Support for Java 9 (and 10)

robby-phd:
This pull request addresses #675.

The approach uses https://github.com/retronym/java9-rt-export to extract Java lib class files to `~/.ammonite/rt.jar` if `sun.boot.class.path` is null and use that (and `amm`) for compilation.
This necessitates `Classpath.classpath` to know Ammonite's storage backend.

`sun.misc.Signal.handlers` is no longer available in Java 9 so `Signaller` is adjusted by maintaining a map of sig handler by itself (please double check).

I copied the code from https://github.com/retronym/java9-rt-export (with its license embedded as comments) directly into the change set of` amm/runtime`. If undesirable, a different mechanism needs to be decided (the tool is not published in maven).

All automated tests (but one) passed (in my machine); `BasicTests.source` fails because the rt tool does not export Java lib source files. The tests were run using Zulu JDK 9.0.4.1.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,7 @@ matrix:
   # published tests tend to run slower than integration tests, so start
   # them first
   - stage: build
-    env: CI_SCRIPT="sbt singleCrossBuilt/test shell/test sshd/test ammInterp/test"
+    env: CI_SCRIPT="sbt published/test"
     jdk: oraclejdk9
     scala: 2.12.4
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,7 @@ matrix:
   # published tests tend to run slower than integration tests, so start
   # them first
   - stage: build
-    env: CI_SCRIPT="sbt published/test"
+    env: CI_SCRIPT="sbt singleCrossBuilt/test shell/test sshd/test ammInterp/test"
     jdk: oraclejdk9
     scala: 2.12.4
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,40 +15,50 @@ matrix:
   # published tests tend to run slower than integration tests, so start
   # them first
   - stage: build
-    env: CI_SCRIPT="ci/build.sc test published/test"
+    env: CI_SCRIPT="sbt published/test"
+    jdk: oraclejdk9
+    scala: 2.12.4
+
+  - stage: build
+    env: CI_SCRIPT="amm ci/build.sc test published/test"
     jdk: oraclejdk8
     scala: 2.12.4
 
   - stage: build
-    env: CI_SCRIPT="ci/build.sc test published/test"
+    env: CI_SCRIPT="amm ci/build.sc test published/test"
     jdk: oraclejdk7
     scala: 2.11.12
 
   - stage: build
-    env: CI_SCRIPT="ci/build.sc test integration/test"
+    env: CI_SCRIPT="sbt integration/test"
+    jdk: oraclejdk9
+    scala: 2.12.4
+
+  - stage: build
+    env: CI_SCRIPT="amm ci/build.sc test integration/test"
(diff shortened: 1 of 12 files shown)


lihaoyi:
This looks great; is there any way you can turn on some Java 9 tests in the Travis config, so we can ensure this works going forward?

robby-phd:
Hmm... amm/test is somehow breaking now... I'm investigating.

robby-phd:
@lihaoyi I made some more fixes and added stages to run integration and published tests on oraclejdk9. 

I had to use sbt to run the tests because published amm does not currently support Java 9.

Status:
* `BasicTests.source` still fails on integration/test due to missing Java lib source files as I previously mentioned
* Some `AutocompleteTests.selection.{import, scope, scopePrefix}` fail. (I missed running published/test before because it wasn't listed in the Ammonite readme.md.) 
  Anyway, can you please take a look at these since you have more insights on how amm  autocompletion should work? (I'm actually not familiar with Ammonite; I'm just trying to get mill to run under Java 9 and hoping the changes will work for Java 10 that's coming soon :smiley:.)

lihaoyi:
@robby-phd sounds good. Even if autocompletion doesn't work, if you can get most of the tests working that's already a huge step forward.

Feel free to conditionally disable any tests that don't work and we can merge what you have once it's green

robby-phd:
@lihaoyi tests are green now.

robby-phd:
@lihaoyi I have changed classpath and rt to be managed by Storage instead of the global variable. 

Unfortunately, there seems to be massive memory leaks when running `sbt amm/test` and `sbt ammRepl/test`, which I think are caused by all the rt jars created by many of the tests (using different temp storage backends) that are then dynamically classload-ed but I think somehow keep lingering around.

`sbt publish/test` actually works on my machine with Java max heap set to 16G; it unfortunately can't be run on travis with the (documented) max 7.5G memory limit.

What do you suggest?

lihaoyi:
@robby-phd the only thing I can think of is that we're forgetting to close some classloader that's keeping all those classes in memory. Could you hook it up to jprofiler and see who's keeping the old classloaders alive in between the tests?

robby-phd:
@lihaoyi That sounds like another pull request.

For now, I simplified my pull request by not caching rt.jar in Ammonite's storage backend; instead, I modified java9-rt-export's `Export` to create a temp rt.jar that's reused in the same JVM instance.  `Classpath.classpath` is back now as a `val` (and I rolled back my changes to {`CompilerLifecycleManager`, `Interpreter`, `Storage`}). `published/test` now passes using default heap memory configuration.

Does this work for you?

Btw, I've tested `published/test` under openjdk 10 early-access build 10+46 (also with default heap memory configuration) and it works.

lihaoyi:
@robby-phd yeah that sounds like a reasonable compromise, will take a look

robby-phd:
@lihaoyi I'm not sure how users typically use Ammonite, but if we're talking about compromise, although not ideal, [2d4d1f](https://github.com/lihaoyi/Ammonite/pull/761/commits/2d4d1f381dadbf19206c125e5ecf15cc3aecb7aa) seems a better choice performance-wise because it does not create rt.jar every time Ammonite is fired up in a fresh JVM. The global mutable variable was used to simulate `Classpath.classpath`, which is a global `val` anyway. 

In that version, it just means that the first storage backend that's used to get the classpath is always used to cache rt.jar in the same JVM instance. For most common use cases (i.e., not Ammonite tests), the storage backend is in `~/.ammonite` (or `~/.mill/ammonite`) which only requires rt.jar extraction once (if the Java version is the same).

Anyway, just a thought.

lihaoyi:
looks good to me

PR #764, opened 2018-03-08
scala 2.10 compat in repl readme

emanresusername:
idk if anybody else has encountered this with projects still on sbt 0.13?
could just be my local config and/or that project setup/plugins is/are borked
either way, feel free to add this to evidence in the "sbt is hard, try mill" case :)
[potentially related](https://github.com/sbt/sbt/issues/3006#issuecomment-355199121)

### DIFF
--- a/readme/Repl.scalatex
+++ b/readme/Repl.scalatex
@@ -82,10 +82,8 @@
   @hl.scala
     libraryDependencies += {
       val version = scalaBinaryVersion.value match {
-        case "2.10" ⇒
-          "1.0.3"
-        case _ ⇒
-          "@ammonite.Constants.version"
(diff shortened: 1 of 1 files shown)


lihaoyi:
oh i stopped publishing 2.10 artifacts for Ammonite; could that be related?

emanresusername:
Yeah I think that's part of it
Idk too much about how SBT resolves SBT plugins vs libraryDependencies
But it looked like it was trying to load the global ammonite during the plugin loading phase (and all plugins are 2.10 im SBT 0.13?? I'm totally guessing at this point)

The exact same project except with SBT 1.x in build.properties didn't have any issues 🤔

PR #766, opened 2018-03-10
Adds preConfigureCompiler to the interpreter

Baccata:
Adds a preConfigureCompiler method to the InterpAPI, allowing to mutate settings in a way that is applied before the interpreter instantiates its next compiler.


amm/interp/src/main/scala/ammonite/interp/CompilerLifecycleManager.scala
```diff
@@ -145,7 +149,13 @@ class CompilerLifecycleManager(headFrame: => Frame){
     }
   }
 
-  def addToClasspath(classFiles: ClassFiles) = synchronized{
+  def preConfigureCompiler(callback: scala.tools.nsc.Settings => Unit) =
+    synchronized {
```
Baccata: wonder if this should be applied straight away ... looks like it doesn't get taken into account in the pre-processing of the next unit

lihaoyi:
lgtm

Issue #762, opened 2018-03-07
scala code blocks are broken in docs


antonkulaga:
What I see on ammonite website is:
```
  scala.collection.MapLike.default$(MapLike.scala:231)
  scala.collection.AbstractMap.default(Map.scala:59)
  scala.collection.MapLike.apply(MapLike.scala:141)
  scala.collection.MapLike.apply$(MapLike.scala:140)
  scala.collection.AbstractMap.apply(Map.scala:59)
```
Instead of scala code blocks

typingArtist:
Looks like a dup of #745

typingArtist:
@lihaoyi , please close.

PR #767, opened 2018-03-10
Add AMMONITE_TEST_SHELL env var for proper doc building

### LINKED ISSUE #745: Shell documentation shows bash not ammonite, includes partial stack trace
Browsing ammonite.io docs right now, clicking on Shell -> Shell Basics -> Working Directory, the examples show bash, not ammonite, and include a partial stack trace, as shown in the image I just captured from the website below (uses cd not cd!). Seems to be true for all the examples in this section - rather than showing ammonite commands, they show bash commands and include this stack trace. 

screen shot 2018-01-29 at 8 53 05 am

typingArtist:
fixes #745 and #762.

I could be wrong, but I think that L354 might need the same path ``(packageBin in (shell, Compile))`` as well.


lihaoyi:
lgtm

PR #771, opened 2018-03-15
Added logic to avoid losing range information

Baccata:
* Adds logic to shift the point in range positions (it does not shift start / end because of some internal validation that crashes the compilation)
* Shares the source file trimming between nodes for optimisation

Rationale : 

When compiling ammonite script with `YrangePos` enabled, the transformation performed in the AmmonitePlugin was discarding the "range" nature of the position. This allows to conserve range information, which improves interoperability with some compiler plugins.

### DIFF
--- a/amm/interp/src/main/scala/ammonite/interp/AmmonitePlugin.scala
+++ b/amm/interp/src/main/scala/ammonite/interp/AmmonitePlugin.scala
@@ -291,6 +291,14 @@ object LineNumberModifier {
 
       override def transform(tree: g.Tree) = {
         val transformedTree = super.transform(tree)
+        // The `start` and `end` values in transparent/range positions are left
+        // untouched, because of some aggressive validation in scalac that checks
+        // that trees are not overlapping, and shifting these values here
+        // violates the invariant (which breaks Ammonite, potentially because
+        // of multi-stage).
+        // Moreover, we rely only on the "point" value (for error reporting).
+        // The ticket https://github.com/scala/scala-dev/issues/390 tracks down
+        // relaxing the aggressive validation.
         val newPos = tree.pos match {
           case s : TransparentPosition if s.start > topWrapperLen =>
               new TransparentPosition(trimmedSource, s.start, s.point - topWrapperLen, s.end)
--- a/amm/interp/src/main/scala/ammonite/interp/AmmonitePlugin.scala
+++ b/amm/interp/src/main/scala/ammonite/interp/AmmonitePlugin.scala
@@ -5,7 +5,7 @@ import ammonite.util.{ImportData, Name, Util}
 import scala.reflect.NameTransformer
 import scala.tools.nsc._
 import scala.tools.nsc.plugins.{Plugin, PluginComponent}
-import scala.reflect.internal.util.{BatchSourceFile, OffsetPosition}
+import scala.reflect.internal.util._
 
 /**
  * Used to capture the names in scope after every execution, reporting them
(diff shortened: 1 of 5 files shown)


amm/interp/src/main/scala/ammonite/interp/AmmonitePlugin.scala
```diff
@@ -284,25 +284,30 @@ object LineNumberModifier {
                        topWrapperLen: => Int) = {
 
     object LineNumberCorrector extends g.Transformer {
+      import scala.reflect.internal.util._
+
+      private val trimmedSource = new BatchSourceFile(g.currentSource.file,
+        g.currentSource.content.drop(topWrapperLen))
```
Baccata: this is an optimisation to avoid re-trimming the source for every node on the tree

amm/interp/src/main/scala/ammonite/interp/AmmonitePlugin.scala
```diff
@@ -284,25 +284,30 @@ object LineNumberModifier {
...
-              val p = new OffsetPosition(con, s.point - topWrapperLen)
-              transformedTree.pos = p
-            }
-          case _ =>    //for position  = NoPosition
+        val newPos = tree.pos match {
+          case s : TransparentPosition if s.start > topWrapperLen =>
+              new TransparentPosition(trimmedSource, s.start, s.point - topWrapperLen, s.end)
+          case s: RangePosition if s.start > topWrapperLen =>
+              new RangePosition(trimmedSource, s.start, s.point - topWrapperLen, s.end)
+          case s: OffsetPosition if s.start > topWrapperLen =>
```
Baccata: s.start == s.point in this case

Baccata:
Not sure I'll ever finish shaving the yak, but my hope is to get ammonite and semanticdb-scalac compatible with each other, eventually allowing some tools written for standard Scala to work with ammonite scripts (maybe even LSP support while we're at it ^^)

amm/interp/src/main/scala/ammonite/interp/AmmonitePlugin.scala
lihaoyi: Shouldn't the `startIn` of the `RangePosition` also be `s.start - topWrapperLen`? and the `endIn` be `s.end - topWrapperLen`?

Baccata: Ideally it would be the case, but the compiler performs some later validation on the trees to make sure the ranges aren't overlapping, which throws when I try what your suggest. I'm not quite sure why, it could have to do with multi-phase scripts. 

With the current change, the behaviour of ammonite remains the same whilst allowing external tools (semanticdb) to work around this issue by adding post-processing logic that would look up the position of the symbol matching the filename (thanks for that btw 😄) and applying the appropriate shift. 

I can open a ticket for further investigation and reference it in a TODO in this piece of code, if that is acceptable

olafurpg: There is a ticket to tone down position validation errors in the compiler https://github.com/scala/scala-dev/issues/390 The validation is required for Scala IDE to efficiently find enclosing trees given a position. semanticdb does not require the validated properties to hold so the errors are unnecessary.

lihaoyi: Sure that sounds good to me

Baccata: Done !  fb8ed68

lihaoyi:
Looks good to me, thanks!

PR #773, opened 2018-03-16
Optimize away Java 9 rt.jar export if one is already in the classpath

robby-phd:
Avoid exporting Java 9 rt.jar if one is already in the classpath in order to mainly save time (exporting rt.jar can take seconds).

One use case is mill's client/server where [the client exports rt.jar once and caches it which it then provides the cached rt.jar to the mill's server](https://github.com/lihaoyi/mill/blob/712f331d24d0934cd9c653a12be7e47e1fff1f5d/clientserver/src/mill/clientserver/Client.java#L33-L37) (that uses Ammonite).

The detection of rt.jar is similar to [mill's](https://github.com/lihaoyi/mill/blob/712f331d24d0934cd9c653a12be7e47e1fff1f5d/core/src/mill/util/ClassLoader.scala#L11-L17).

### DIFF
--- a/amm/runtime/src/main/scala/ammonite/runtime/Classpath.scala
+++ b/amm/runtime/src/main/scala/ammonite/runtime/Classpath.scala
@@ -24,7 +24,16 @@ object Classpath {
    * want to do something.
    */
   var current = Thread.currentThread().getContextClassLoader
-  val files = collection.mutable.Buffer.empty[java.io.File];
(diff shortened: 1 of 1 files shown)


lihaoyi:
Looks good to me

robby-phd:
It looks like the sonatype release failed, anyway to re-trigger it?

PR #776, opened 2018-03-20
Java 9 rt.jar caching redux

### LINKED ISSUE #675: amm throws NPE on OpenJDK 9

robby-phd:
This PR re-introduces Java 9 rt.jar caching from #761 that I removed due to memory leaks in Ammonite tests.

The main difference is that it now avoids caching rt.jar if the storage backend is in a temp directory, which is detected based on `System.getProperty("java.io.tmpdir")`. This temp directory caching avoidance sidesteps the memory leaks in Ammonite tests, which mostly use temp dir storage backend (and some `Storage.InMemory`). The heuristics for avoiding caching in `rtCacheDir ` can be changed in the future to cover more (or different) cases.

This finally fixes #675.

### DIFF
--- a/amm/interp/src/main/scala/ammonite/interp/CompilerLifecycleManager.scala
+++ b/amm/interp/src/main/scala/ammonite/interp/CompilerLifecycleManager.scala
@@ -20,7 +20,7 @@ import scala.tools.nsc.Settings
   * compiler objects are initialized, or worry about initializing them more
   * than necessary
   */
-class CompilerLifecycleManager(headFrame: => Frame){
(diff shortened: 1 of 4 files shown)


robby-phd:
@lihaoyi Any thoughts on this?

lihaoyi:
@robby-phd lgtm

PR #779, opened 2018-03-26
Always create Java 9 rt cache dir

robby-phd:
This PR adds creation of Java 9 rt cache dir for robustness.
While Ammonite creates its home directory used for caching rt.jar before `Classpath.classpath` is called, mill, however, does not. It is safer to fix the issue here by removing any assumption on the existence of rt cache directory.


robby-phd:
@lihaoyi Ready for review (ignore those failing refs `49726a2` and `c2c3c02`).

lihaoyi:
@robby-phd no idea but looks good to me :)

PR #781, opened 2018-03-27
Run amm as shell and cmd script

LolHens:
As of v1.1.0 ammonite can run on windows.
This pr adds the ability to run the ammonite shell directly as a cmd script while it remains being a proper shell script.
With the .cmd or .bat extension it can be run by double clicking making it very easy for newcomers to launch the ammonite shell.
Feedback is very welcome :)

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -152,10 +152,33 @@ lazy val amm = project
 
 
     assemblyOption in assembly := (assemblyOption in assembly).value.copy(
-      prependShellScript = Some(
+      prependShellScript = {
+        def universalScript(shellCommands: Seq[String],
+                            cmdCommands: Seq[String],
+                            shebang: Boolean = true): Seq[String] = {
+          Seq(
+            Seq("#!/usr/bin/env sh")
+              .filter(_ => shebang),
+            Seq(":; alias ::=''"),
+            (shellCommands :+ "exit")
+              .map(line => s":: $line"),
+            "@echo off" +: cmdCommands :+ "exit /B",
+            Seq("\r\n")
+          ).flatten
+        }
+
+        def defaultUniversalScript(javaOpts: Seq[String] = Seq.empty, shebang: Boolean = true): Seq[String] = {
+          val javaOptsString = javaOpts.map(_ + " ").mkString
+          universalScript(
+            shellCommands = Seq(s"exec java -jar $javaOptsString" + """$JAVA_OPTS "$0" "$@""""),
+            cmdCommands = Seq(s"java -jar $javaOptsString" + """%JAVA_OPTS% "%~dpnx0" %*"""),
+            shebang = shebang
+          )
+        }
+
         // G1 Garbage Collector is awesome https://github.com/lihaoyi/Ammonite/issues/216
-        Seq("#!/usr/bin/env sh", """exec java -jar -Xmx500m -XX:+UseG1GC $JAVA_OPTS "$0" "$@"""")
-      )
+        Some(defaultUniversalScript(Seq("-Xmx500m", "-XX:+UseG1GC")))
+      }
     ),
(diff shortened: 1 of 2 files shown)


lihaoyi:
@robby-phd as the author of https://github.com/lihaoyi/mill/pull/243, could you help review this? =)

build.sbt
```diff
@@ -152,10 +152,34 @@ lazy val amm = project
...
+            Seq("shopt -s expand_aliases", "alias ::=''")
+              .map(line => s":; $line"),
+            (shellCommands :+ "exit")
+              .map(line => s":: $line"),
+            "@echo off" +: cmdCommands :+ "exit /B",
+            Seq("\r\n")
+          ).flatten
+        }
+
+        def defaultUniversalScript(javaOpts: Seq[String] = Seq.empty, shebang: Boolean = true): Seq[String] = {
```
robby-phd: `shebang` defaults to `true` gives me:

```
'#!' is not recognized as an internal or external command,
operable program or batch file.
```

Defaulting to `false` works for both sh and batch modes (Windows 10).
Any reason why it defaults to `true`?

build.sbt
```diff
@@ -152,10 +152,34 @@ lazy val amm = project
...
+          val javaOptsString = javaOpts.map(_ + " ").mkString
+          universalScript(
+            shellCommands = Seq(s"exec java -jar $javaOptsString" + """$JAVA_OPTS "$0" "$@""""),
+            cmdCommands = Seq(s"java -jar $javaOptsString" + """%JAVA_OPTS% "%~dpnx0" %*"""),
```
robby-phd: Though it's not necessary at this point as the commands are single-line now, it would be nice to use multiline `String`s here instead of `Seq[String]` and have `shellCommands` and `cmdCommands` split by `universalScript` just in case the scripts need to be adjusted in the future to include more lines.

lihaoyi:
@robby-phd I also gave you commit access to the Ammonite repo, feel free to merge when you're happy!

robby-phd:
This is an interesting coincidence, I was looking at https://stackoverflow.com/questions/17510688/single-script-to-run-in-both-windows-batch-and-linux-bash to see if mill's sh and batch versions can be combined after I did lihaoyi/mill#243 (but I had to fix other things ;p).

@lihaoyi Thanks; that would make it easier to do round trip Ammonite<->mill (trivial) changes. :)

@LolHens Would you be willing to bring this feature to mill as well (adapting lihaoyi/mill#243)? =)

build.sbt
LolHens: Yes, this is an unfortunate tradeoff. If the shebang line is disabled you cannot simply run "./amm" on linux.

build.sbt
LolHens: I'm not quite sure if I understand this correctly but multiple lines would be possible like this:
```Seq("line1", "line2")```
or like that:
```Seq("line1\nline2")```

**Comment by LolHens** [comment]
I'll have to look into it but it should be possible

File: build.sbt (hunk 1 above)
Comment (robby-phd) [reply build.sbt]: Hmm, perhaps a different shell might not like it; I tried without it on macOS's terminal (`GNU bash, version 3.2.57(1)-release (x86_64-apple-darwin17)`) and KDE Neon's Konsole (`GNU bash, version 4.3.48(1)-release (x86_64-pc-linux-gnu)`). 
At any rate, I think it's a better trade-off to enhance the curl+chmod command in Ammonite installation guide to prepend the shebang line (rather than all batch users getting the above error message).

File: build.sbt (hunk 2 above)
Comment (robby-phd) [reply build.sbt]: Yes, but one would rather write:

```scala
"""line1
  |line2""".stripMargin
```

Though, I think using what you have and passing:

```scala
"""line1
  |line2""".stripMargin.split('\n')
```

would work too. So, nevermind.

robby-phd:
That would be great!

PR #789, opened 2018-03-30
Updated universal script and removed shebang line by default

LolHens:
This pr changes updates the generated universal script to be on par with lihaoyi/mill#264.
Also the shebang line isn't generated anymore by default so that it doesn't show an error on windows anymore.
The linux download instructions should be updated to add this line afterwards.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -159,7 +159,8 @@ lazy val amm = project
           Seq(
             Seq("#!/usr/bin/env sh")
               .filter(_ => shebang),
-            Seq(":; alias ::=''"),
+            Seq("shopt -s expand_aliases", "alias ::=''")
+              .map(line => s":; $line"),
             (shellCommands :+ "exit")
               .map(line => s":: $line"),
             "@echo off" +: cmdCommands :+ "exit /B",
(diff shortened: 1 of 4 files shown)


robby-phd:
Nice, thanks. Do you want to propose the change to the curl+chmod command [here](https://github.com/LolHens/Ammonite/blob/c1e5cef116c7062a946b0466ce38977bde0ab0f4/readme/Sample.scala#L16-L21) and [here](https://github.com/LolHens/Ammonite/blob/c1e5cef116c7062a946b0466ce38977bde0ab0f4/readme/Footer.scalatex#L1451) (need a separate `unstableCurlCommand` for now); perhaps throw in some Windows instructions for good measure (e.g., a simple url direct link with a brief "save as amm.bat" instruction would suffice).

LolHens:
What do you think about this?
```sudo sh -c '(echo "#!/usr/bin/env sh" && curl -L $curlUrl) > /usr/local/bin/amm && chmod +x /usr/local/bin/amm' && amm```

robby-phd:
That works.

LolHens:
Okay I added this to the docs as well as the info for windows. Sadly I was not able to figure out how to generate the page to test out if it looks right since I have never worked with scalatex before.
I already tried running ```sbt "readme/run"``` but it only seems to generate the api docs.

robby-phd:
You can run `amm ci/build.sc docs` to compile the readme (it does not replace the curl url, however).

Anyway, this looks good to me. Thanks.

PR #792, opened 2018-04-06
Adding CodeSource to CodeWrapper API

Baccata:
Motivation : with this change, an extremely minor change to mill [there](https://github.com/lihaoyi/mill/blob/e7ac7323693edbdc350ec36b578ba3e101657360/main/src/mill/main/MainRunner.scala#L127) would allow to wire together different projects built with mill using `import $file.external.directory.build`, effectively providing what https://github.com/storm-enroute/mecha tried to provide in a much simpler way. 

(Tried it on a very simple example, it worked)


lihaoyi:
Not sure I understand how this helps you but looks good i guess

PR #807, opened 2018-05-10
Switch to latest sbt-coursier

alexarchambault:


lihaoyi:
@alexarchambault sorry but SBT is gone 😄

PR #808, opened 2018-05-10
Try not to load duplicated JARs if possible

alexarchambault:
This PR adds a few tricks to try not to load several versions of a given dependency.

It doesn't cover all cases (one can always load newer versions of already loaded stuff), and doesn't try to warn users in case something wrong was prevented (if users add a newer version of something already in Ammonite itself, this PR prevents it to be loaded - silently).

In more details, the first commit tries to prevent loading JARs that are already in Ammonite itself. It generates a resource file (`ammonite-deps.txt`), listing all the dependencies of Ammonite as known from sbt. It then excludes these dependencies from resolutions, so that no newer or older versions of those can be loaded again.

Second commit is just some refactoring for the next one.

Third commit tries not to load JARs for previous versions of already loaded dependencies. It keeps the already loaded dependencies in `Frame`, and adds those to the resolution. That way no old version of an already loaded dependency can end up being loaded when adding new dependencies.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -72,7 +72,7 @@ script:
 - export PATH=~/bin/amm:~/bin/mill:$PATH
 - mkdir -p ~/bin
 # We use 2.11 since we need to run tests on Java 7
-- if [ ! -f ~/bin/mill ]; then sudo sh -c '(echo "#!/usr/bin/env sh" && curl -L https://github.com/lihaoyi/mill/releases/download/0.2.3/0.2.3) > ~/bin/mill && chmod +x ~/bin/mill'; fi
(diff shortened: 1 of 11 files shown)


alexarchambault:
Looks like github messed with the commits ordering… Right now, it's `dee5fe1`, then `90bb6bb`, and `e42e39f`.

alexarchambault:
Overall, the motivation for that is to use things like spark et al. from Ammonite. Spark needs to be passed all the JARs loaded from the session (so that it can send them to a bunch of remote machines, and then send them code to run - there are similar stuff in flink, or scio…) Some issues arise if several versions of some dependencies are passed to it (like Java deserialization issues when several versions of scala-library are loaded). So that PR tries to prevent that situation from happening…

alexarchambault:
Ok, got to replace the sbt code by some mill stuff!

alexarchambault:
@lihaoyi This is ready for review I guess.

lihaoyi:
@alexarchambault I don't have time to give this a proper review right now, but if you're confident it does what you think and everything is green then feel free to merge

alexarchambault:
Ok, merging!

PR #831, opened 2018-06-30
Fix compilation of build for mill 0.2.3

kubukoz:
- Fix `SonatypePublisher` constructor arguments in build.sc


kubukoz:
Looks like I'd need to update the Mill version in the build, too. Currently the version used is:
`github.com/lihaoyi/mill/releases/download/0.2.0/0.2.0-32-14d002`, which one should I use?

And most importantly: Do we want to upgrade Mill now, or is this too early and you'd like it updated only when the minor version is bumped?

lihaoyi:
bumping the mill version seems fine

alexarchambault:
Merged via https://github.com/lihaoyi/Ammonite/pull/808.

PR #841, opened 2018-08-01
Fix custom bridges handling

alexarchambault:
This PR adds a bridge in `TestRepl`, and ensures it can be used from the user code. For the test to be green, this required to pass the extra bridge class names to `SpecialClassLoader`, so that it can special case them too (like it does for the default Ammonite bridges).


alexarchambault:
I guess I'll merge that one in the coming days, unless you oppose, @lihaoyi!

PR #842, opened 2018-08-02
Allow to adjust coursier resolutions just before they are run

alexarchambault:


alexarchambault:
@lihaoyi Like https://github.com/lihaoyi/Ammonite/pull/841, I guess I'll merge in the coming days, unless you disagree (or want to wait a bit…)

PR #843, opened 2018-08-02
Allow to use custom base class loader for the session

alexarchambault:


alexarchambault:
Only thing I'm not fully happy with in this PR is that what these changes bring isn't illustrated in some tests… I use that to add some class loader isolation in the upcoming version of [jupyter-scala](https://github.com/jupyter-scala/jupyter-scala), which requires starting from a different class loader than `Thread.currentThread().getContextClassLoader`. The right class loader is set up via the `-i` / `-I` options of the `launch` and `bootstrap` commands of `coursier` (that lack a bit of doc…)

PR #845, opened 2018-08-07
Tweak and test the position returned by Pressy.complete

alexarchambault:


alexarchambault:
I'd like to add some scaladoc to `Pressy.complete` to clarify the motivations behind this PR (and ensure I don't break anything).

alexarchambault:
@lihaoyi This PR changes the index returned by the `Pressy.complete` method in *some* cases.

IIUC, it should correspond to the index from which a completion should be inserted (that is, a completion could be substituted between the returned index and the "caret" position). [Some test cases](https://github.com/lihaoyi/Ammonite/blob/5ef5ee1799de3bc82ff4dd6e4c6189b0b614084e/amm/repl/src/test/scala/ammonite/interp/AutocompleteTests.scala#L106-L108) fall short of that explanation though…

alexarchambault:
I just hope this doesn't break stuff that would not be in the tests…

alexarchambault:
I'm gonna go ahead and merge that one… If anyone thinks anything broke wrt to completion in the console, come tell it here!

PR #846, opened 2018-08-09
Fix jline autocomplete

### LINKED ISSUE #817: Completion causes cursor misalignment
image
If you press TAB here, the cursor will move to the wrong position:
image
image

sake92:
Fixes #817, #823.  
Cursor now behaves correctly. Please report if there are still issues with it.  
I still couldn't figure out why it doesn't accept a candidate when there are braces at the end (assume `|` is cursor): `x.toSt|()` (but it stays there now.. xD).

When you hit `Enter` now, it will only accept the input when cursor is at the end.

There's also a fix in `FrontEndUtils` for `width` and `height`, I forgot when the exception was happening...


lihaoyi:
looks good to me, gave you commit access so merge whenever you're happy

sake92:
Thank you @lihaoyi! It means a lot to me. 😃

PR #850, opened 2018-08-21
Allow non-files in classpath

alexarchambault:
~*Not sure I'd like that to be merged yet*, but weighting in / out welcome~

Motivation for that PR is to have Ammonite be fine with unusual setups, where the initial classpath isn't necessarily made entirely of files, but could be made of more general `java.net.URL`s.

For now, Ammonite assumes its initial classpath [is made of files](https://github.com/lihaoyi/Ammonite/blob/c4ab4986cd950ae2efd6e97b56d212db7cfe050c/amm/runtime/src/main/scala/ammonite/runtime/Classpath.scala#L27). This PR keeps that classpath as URLs a bit longer. If a URL doesn't correspond to a file, a `scala.reflect.io.URLZipArchive` rather than a [`scala.reflect.io.FileZipArchive`](https://github.com/lihaoyi/Ammonite/blob/c4ab4986cd950ae2efd6e97b56d212db7cfe050c/amm/interp/src/main/scala-2.11/ammonite/interp/GlobalInitCompat.scala#L64) is passed to scalac.

`scala.reflect.io.URLZipArchive` seems buggy, so this PR also adds a substitute for it (second commit).

The main case of unusual setup where this is useful is when one uses "standalone bootstraps" of coursier (`--standalone` option of `coursier bootstrap`). The classpath of the apps it launches corresponds to resources in the bootstrap, so aren't files per se.

### DIFF
--- a/amm/interp/src/main/scala-2.11/ammonite/interp/GlobalInitCompat.scala
+++ b/amm/interp/src/main/scala-2.11/ammonite/interp/GlobalInitCompat.scala
@@ -62,7 +62,11 @@ object GlobalInitCompat {
     val jarCP =
       jarDeps.filter(x => x.getPath.endsWith(".jar") || Classpath.canBeOpenedAsJar(x))
         .map { x =>
-          val arc = new FileZipArchive(java.nio.file.Paths.get(x.toURI).toFile)
+          val arc =
+            if (x.getProtocol == "file")
+              new FileZipArchive(java.nio.file.Paths.get(x.toURI).toFile)
+            else
+              new internal.CustomURLZipArchive(x)
           new DirectoryClassPath(arc, jCtx)
         }
         .toVector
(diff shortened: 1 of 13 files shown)


alexarchambault:
Not sure yet how I could add non regression tests for that here…

lihaoyi:
Maybe spin up a file server on localhost using jetty and add an integration test with some URL items on the spawned ammonite classpath?

alexarchambault:
I may go ahead and merge that one… It'll be quite useful to generate [assembly-like launchers](https://github.com/almond-sh/almond/blob/d9f838f74dbc95965032e8b51568f7c1c7f2e71b/docs/pages/install-other.md#all-included-launcher) for almond. (Standard assemblies aren't fine for it, as they mix everything which breaks class loader isolation. coursier-bootstrap-generated stuff, OTOH, is fine, but relies on [custom URL protocols](https://github.com/coursier/coursier/blob/d848b574e11483a0ec76c3b09ee946aa0e9819ee/modules/bootstrap/src/main/java/coursier/bootstrap/Bootstrap.java#L360), hence the need of this PR.)

Also, I'm planning to propose to add similar class loader isolation stuff in Ammonite, and this could help for that too.

Atry:
I guess this feature can help integration with Spark class server.

alexarchambault:
@Atry What do you mean?

Atry:
Never mind. I just thought into a wrong path.

PR #851, opened 2018-08-21
Fix JLine dependencies

smarter:
jline-$VERSION.jar is a bundle that contains the content of
jline-terminal-$VERSION.jar and other modules. jline-terminal-jna
depends on jline-terminal. This means that the Ammonite build depends on
both jline and jline-terminal and classfiles may be read from either
jar (dependencies on the classpath are usually not in a stable order).
They should be identical so it's not a big problem, but it breaks Zinc
incremental compilation. This commit fixes this by not using the jline
bundle jar anymore.

### DIFF
--- a/build.sc
+++ b/build.sc
@@ -152,8 +152,9 @@ object amm extends Cross[MainModule](fullCrossScalaVersions:_*){
       terminal()
     )
     def ivyDeps = Agg(
-      ivy"org.jline:jline:3.6.2",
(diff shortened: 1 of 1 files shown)


smarter:
@lihaoyi Do you think you could publish a release of ammonite with this PR so that Mill can use it? It would make it easier to track other incremental compilation issues I'm working on.

lihaoyi:
try `1.1.2-28-e8b4996`

PR #853, opened 2018-08-23
Optims

alexarchambault:
In a few words, this PR:
- tries to have Ammonite not unnecessarily run stty commands before / after each line of code, but only when Ammonite starts and exits instead (first commit),
- tries to have Ammonite not discard its compiler instance after almost each line, but only when new JARs are added to the classpath (actual JARs, not just compilation class files, second and third commits).

(of course, github tangled with the commits ordering…)

### DIFF
--- a/amm/interp/src/main/scala/ammonite/interp/CompilerLifecycleManager.scala
+++ b/amm/interp/src/main/scala/ammonite/interp/CompilerLifecycleManager.scala
@@ -27,7 +27,6 @@ class CompilerLifecycleManager(storage: Storage, headFrame: => Frame){
   private[this] object Internal{
     val dynamicClasspath = new VirtualDirectory("(memory)", None)
     var compiler: Compiler = null
-    var dynamicClasspathChanged: Boolean = true
     var pressyStale: Boolean = true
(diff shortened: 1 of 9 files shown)


alexarchambault:
Results of a quick benchmark:
amm benchmark

Once the CI is green here, I can post the detailed methodogy if that's of interest to anyone…

alexarchambault:
I have only a ~surface understanding of the stuff I changed… (compiler discarding, and stty tangling) Have a careful look at them, before possibly merging that…

alexarchambault:
Preliminary benchmark instructions [here](https://github.com/alexarchambault/ammonite-benchmark)

alexarchambault:
I'll probably merge that in the coming days, unless anyone opposes.

About the commented out failing test case, it seems unrelated to the core of this PR… More likely an unearthed corner case in TPrint (it only happens in 2.12, when class wrapping is enabled, the latter not being enabled by default).

The only point I'm not too sure about is whether this can mess with the terminal when one runs commands with `%` and `%%`. I guess the terminal state could be applied again right after a command is run. It can always be done in other PR in any case…

alexarchambault:
So, merging!

Issue #751, opened 2018-02-14
Scrolling on Website is Unusably Slow


jamesward:
On the latest Chrome, scrolling up/down on http://ammonite.io/ takes forever.  Here is the profiling output:
![screen shot 2018-02-14 at 12 53 41 pm](https://user-images.githubusercontent.com/65043/36224885-1ed7796e-1186-11e8-8bde-4b08983fdd1a.png)

AndyMoreland:
workaround is to block javascript on the page -- makes it nice and speedy.

sake92:
Overall site load says: 
- 12.3 MB, most of it are images, gifs...  
- 14 seconds to load

Some stuff from audit tab:
- first meaningful paint: after 5.7 seconds
- does not use HTTPS

All in all, very poor performance. 😟   
We should split it to multiple pages. It would be more user-friendly and easier to read.

My suggestion is to use a static site generator like Jekyll, Docusaurus ([scalafmt docs](https://github.com/scalameta/scalafmt/pull/1199) is migrating to it) or even my [Hepek generator](https://sake92.github.io/hepek/) (IMHO better than to use a custom script like Mill [does](https://github.com/lihaoyi/mill/blob/master/docs/build.sc)).  
I'm willing to spend time on this.

sake92:
After #1170 now it takes 3 seconds to load, and around 2MB on initial load. 🎉

PR #854, opened 2018-08-27
WIP: Docs with hepek generator

sake92:
This is my attempt to help with #751 issue (Scrolling on Website is Unusably Slow).  
General problem is that the whole website is one big page. It should be organized better to smaller units, for easier navigation, links and reading.

[Here](https://sake92.neocities.org/amm/index.html) is live preview of what I managed to migrate for now (only homepage and bit of REPL).  
I need some feedback in order to continue with this.

Some stuff that are different from Scalatex are:
- no templates, everything is embedded in Scala (more cumbersome to read but we get IntelliJ etc.)
- section/page links are typesafe, e.g. `Page.ref` and `Page.section.ref`
- site is more modular and it is easier to swap frameworks (Bootstrap currently), dependencies etc.
- we can use PDF generator integrated in Hepek


lihaoyi:
Honestly I don't really like how the proposed docs look, and not just the CSS. Even split up, each page is very tall, and lacking a navbar impossible to find your way around. 

If you're going to go for a non-auto-expanding navbar, you'll have to make sure every page has an appropriate number of sub-sections so the navbar is of reasonable size.

I'd prefer going to a Mill-style readme over what's proposed here, with it's hardcoded two-level nav bar (inter-page and intra-page), using Markdown with some preprocessing, over transition to another large static site generator.

sake92:
This is just the default theme, everything regarding layout can be overriden or reimplemented like [here](https://github.com/sake92/hepek/blob/master/src/main/scala/ba/sake/hepek/theme/bootstrap3/HepekBootstrap3BlogPage.scala).  
Regarding navbar/navigation, we could add [bootstrap srollspy](https://www.w3schools.com/bootstrap/bootstrap_scrollspy.asp) or something.  

I wanted to get some feedback before continuing.  
If you don't feel this is the way to go, I'm ok with that, of course. 😄

PR #863, opened 2018-09-12
Support auto-complete of back-quoted symbols.

### LINKED ISSUE #861: Implement auto-complete of `Symbols with Backquotes`
The stock Scala REPL gets this wrong as well, which I filed as https://github.com/scala/bug/issues/11140

I'd like to contribute here, though having trouble building locally at the moment, so will reach out on Gitter.

jameskoch2:
Fixes #861.

First-time contributor, apologies in advance 😄 

1. This includes a hack to leverage protected backquoting method in scala-compiler. Arguably better to just re-implement the backquoting rules in Ammonite until they're exposed more-accessibly upstream.
2. I stumbled upon a dubious prior behavior. Prior to my changes:
  a. `val x = new { val foo = 1 }; x.<caret>` completed as `"foo"`
  b. `object x { val foo = 1 }; x.<caret>` completed as `"foo "` (note trailing space)

I tracked this back to the a Scala implementation detail called LOCAL_SUFFIX_STRING which seems to have [caused head-scratching for others](https://github.com/scala/bug/issues/5736).

I had to strip this whitespace to avoid having the trailing space trigger back-quoting unnecessarily.

There's a risk that users' muscle memory is actually relying on the whitespace.

I'm also worried there's nuance to the whitespace that I'm missing. The nsc REPL alters behaviors based on the trailing space in ways that I don't fully grok. See uses of `symNameDropLocal` [here](https://github.com/scala/scala/blob/1e904d75450373d7845eb7452d940e52d955b27d/src/repl/scala/tools/nsc/interpreter/PresentationCompilation.scala#L158).

### DIFF
--- a/amm/interp/src/main/scala/ammonite/interp/Pressy.scala
+++ b/amm/interp/src/main/scala/ammonite/interp/Pressy.scala
@@ -1,6 +1,6 @@
 package ammonite.interp
 
-
+import java.io.{OutputStream, PrintWriter}
 
 import ammonite.interp.Compiler.makeReporter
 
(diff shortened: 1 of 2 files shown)


jameskoch2:
@alexarchambault FYI

alexarchambault:
@jameskoch2 The CI errors seems legit: two are because of too long lines (check locally with `mill 'ops[2.12.6].test.test test.ammonite.ops.ExampleTests'`), and the 2.11 one seems to originate from a dead-lock. Replacing `memberToString` by a `val` of type `pressy.Member => String` seems to fix the latter - not sure what causes it in the first place though…

lihaoyi:
not gonna have time to look closely soon, @alexarchambault want to help review? :P :P :P

alexarchambault:
Looks good to me, merging! Thanks @jameskoch2!

PR #866, opened 2018-09-15
Fix pre scala 2.11.8 builds

alexarchambault:
https://github.com/lihaoyi/Ammonite/pull/863 uses stuff not available in scala-reflect < 2.11.8, which broke [one of the *release* jobs](https://travis-ci.org/lihaoyi/Ammonite/jobs/428887990) that builds Ammonite against pre 2.11.8 scala versions. This PR fixes that.


alexarchambault:
We could manage to have the build jobs check that Ammonite is fine with scala < 2.11.8, but I'm sure it's worth consuming more build resources for that (this shouldn't happen too often…) Don't know what is your opinion about it, @lihaoyi.

PR #878, opened 2018-10-02
Remove ci/build.sc

alexarchambault:
Checking if this breaks anything on the CI...


alexarchambault:
@lihaoyi So I guess this can be removed?

sake92:
Most of the stuff in the script references old sbt build, and the docs part is moved to root `build.sc`.  
So I guess this is redundant now. 🙂

Issue #885, opened 2018-10-10
rm follows symlinks thus recursively deleting linked content


arberg:
if x refers to root (as in /) then rm(x) deletes entire harddrive (according to my tests)

Reproduce steps
bash
	```
mkdir -p /mnt/user/homes/alex/test
cd /mnt/user/homes/alex/test
ln -s /mnt/user/homes/alex/test subtest
echo test > test.file
echo test2 > subtest/test2.file
mkdir aDir
echo test3 > aDir/test3.file

Now run in ammonite
`
cd!(Path("/mnt/user/homes/alex/test"))
rm! 'subtest
`

Problem:
Delete entire test-dir, except aDir (but also deleted its file). For some reason had that setup on my media -> /media folder, and then I played around using ammonite to delete the link, and OMG it took longer than 1 sec. So I broke the command, killed the shell, checked ps to make sure ammonite wasn't running, and then did some checks against the filesystem coverage with md5 that I have to check nothing got deleted. Phew! 

rm definitely should not traverse symlinks dirs to wipe user content within dirs

arberg:
see https://github.com/lihaoyi/Ammonite/pull/886

arberg:
I just noticed the original rm-code only traverses and deletes linked directories when the file given to rm is a symlinked dir, not when subdirs contain symlinked dirs. The problem being the first line of the Walker.      
def these = ls.iter(p)

This is also the reason why my suggested solution calls delete and not recursiveDelete on the root-target.

I don't know if this is actually a problem in the walker, or if it is desired behaviour in other areas. Of cause ls('symlinkedDir) should list the content of the symlinked dir, but maybe ls.rec('symlinkedDir) should not, since it does not recursively list content of other symlinked dirs.

PR #886, opened 2018-10-11
rm does not follow symlinks

arberg:


arberg:
https://github.com/lihaoyi/Ammonite/issues/885

arberg:
I havn't managed to get IntelliJ to properly compile Ammonite source yet, so I have not updated tests. The tests making sure rm don't traverse symlinks wont work on windows, because there administrator rights are required to create symlinks. But maybe we could wrap the tests in a test for OS.

I think its important to weed out this bug quickly as it nearly killed all my media files, and that's kind of what ammonite was created to avoid or so I heard :)

I think its a good idea to not build a huge list of file to traverse for deletion before starting the deletion, because it'll take forever for a large dir to generate the list, and the rm may fail on the very first file due to access errors or similar. Better get started with deletion quickly using simple recursion directly in rm.

The isSymLink test is not necessary, but I added it in case isDir is changed in the future to be true in case its a symlink to a dir. If we get a test written, we can remove the isSymlink test.

lihaoyi:
this has been superseded by the extraction of ammonite-ops logic into os-lib

PR #891, opened 2018-10-25
Make the presentation compiler accessible from the REPL API

alexarchambault:
That also makes it accessible when using Ammonite as a library, which is what actually interests me here…

### DIFF
--- a/amm/repl/src/main/scala/ammonite/repl/Repl.scala
+++ b/amm/repl/src/main/scala/ammonite/repl/Repl.scala
@@ -86,6 +86,7 @@ class Repl(input: InputStream,
         def history = repl.history
         def newCompiler() = interp.compilerManager.init(force = true)
         def compiler = interp.compilerManager.compiler.compiler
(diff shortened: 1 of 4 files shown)


alexarchambault:
Ok, so the JDK9 CI jobs fail… ([Travis](https://travis-ci.org/lihaoyi/Ammonite/jobs/446166498), [Appveyor](https://ci.appveyor.com/project/lihaoyi/ammonite/builds/19787094/job/05wl81i46pqo2d7d)) Complaining that the additional method on `ReplAPI` cannot be found on `FullRepAPI` (which extends the former)

```
ammonite.session.AdvancedTests.accessPressy 1468ms 
  utest.AssertionError: assert{
  error: String = cmd1.sc:4: value interactiveCompiler is not a member of ammonite.repl.FullReplAPI
    val c = repl.interactiveCompiler
                 ^
  Compilation Failed
```

SethTisue:
change 9 to 11 and see if it still fails?

alexarchambault:
Same with JDK 11 😓

PR #893, opened 2018-10-26
Make Pressy and its compiler accessible

alexarchambault:
Like https://github.com/lihaoyi/Ammonite/pull/891, but without an extra method to also access it via `ReplAPI`. See the discussion in https://github.com/lihaoyi/Ammonite/pull/891 for the issues this triggers.


alexarchambault:
Gonna go ahead and merge that one…

Issue #895, opened 2018-11-02
Possible caching bug when manually nuking cache


jackkoenig:
I realize that nuking caches manually is not the greatest idea, but I (and some of my users) got in the habit when SBT sometimes gets in a weird state.

Using Ammonite version `1.3.3`
```
$ amm --help
Ammonite REPL & Script-Runner, 1.3.3
```

Given the following script that has an ivy import
```scala
#!/usr/bin/env amm
import $ivy.`com.lihaoyi::scalatags:0.6.7`
println("Hello World!")
```

Running it, works great
```
$ ./script.sc
https://repo1.maven.org/maven2/com/lihaoyi/scalatags_2.12/0.6.7/scalatags_2.12-0.6.7.pom
  100.0% [##########] 2.6 KiB (7.0 KiB / s)
https://repo1.maven.org/maven2/com/lihaoyi/scalatags_2.12/0.6.7/scalatags_2.12-0.6.7.jar
  100.0% [##########] 474.2 KiB (2.5 MiB / s)
https://repo1.maven.org/maven2/com/lihaoyi/scalatags_2.12/0.6.7/scalatags_2.12-0.6.7-sources.jar
  100.0% [##########] 70.3 KiB (1.4 MiB / s)
Compiling script.sc
Hello World!
```

But if I manually nuke my .coursier cache, it just crashes:
```
$ rm -rf ~/.coursier/cache/*
$ ./script.sc
Compiling /Users/jack/work/mill-bug/script.sc
java.nio.file.NoSuchFileException: /Users/jack/.coursier/cache/v1/https/repo1.maven.org/maven2/com/lihaoyi/scalatags_2.12/0.6.7/scalatags_2.12-0.6.7.jar
  sun.nio.fs.UnixException.translateToIOException(UnixException.java:86)                
  sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:102)
  sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:107)                    
  sun.nio.fs.UnixFileAttributeViews$Basic.readAttributes(UnixFileAttributeViews.java:55)
  sun.nio.fs.UnixFileSystemProvider.readAttributes(UnixFileSystemProvider.java:144)
  java.nio.file.Files.readAttributes(Files.java:1737)                                 
  scala.tools.nsc.classpath.FileBasedCache.$anonfun$getOrCreate$1(ZipAndJarFileLookupFactory.scala:187)
[...]

PR #896, opened 2018-11-03
Check that cached ivy dependencies exist

jackkoenig:
Possible fix for #895

When reading the cached IvyMap, check that the jar files actually exist. Filtering based on their existence allows `Ammonite` to simply re-download them if something has happened to one of the `jars`

### DIFF
--- a/amm/runtime/src/main/scala/ammonite/runtime/Storage.scala
+++ b/amm/runtime/src/main/scala/ammonite/runtime/Storage.scala
@@ -1,6 +1,7 @@
 package ammonite.runtime
 
 import java.nio.file.FileAlreadyExistsException
+import java.nio.file.{Files, Paths}
 
 import ammonite.ops._
 import ammonite.util._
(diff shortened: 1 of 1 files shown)


alexarchambault:
Let's merge this, thanks @jackkoenig!

alexarchambault:
Ideally, I wish the Ammonite's ivy cache didn't have its own logic, that comes on top to the one of coursier. With subtle differences between both (coursier has TTLs for snapshot stuff in particular, so it knows whether elements in its cache can be relied on or not…)

But @lihaoyi makes good points about scrapping Ammonite's ivy cache altogether in https://github.com/lihaoyi/Ammonite/pull/876…

jackkoenig:
Thanks for the merge! I agree that dropping the ivy logic in favor of coursier would be ideal (if it meets functional and performance requirements of course).

In any case I stupidly kept running into #895 so thanks for merging my patch!

PR #898, opened 2018-11-05
Fix shellout on Windows

sake92:
With these changes `mill -i ops[2.12.7].test` passes on Windows also.


lihaoyi:
This looks good I guess. Now that most of Ammonite-Ops is just a wrapper around https://github.com/lihaoyi/os-lib, should we make similar changes there too?

lihaoyi:
closing due to inactivity

PR #908, opened 2018-12-11
Allow to tweak FullReplAPI.Internal…

alexarchambault:
…when using Ammonite as a library.

In particular, this allows to print nothing at all for types other than `Unit`. I'll also probably rely on it to do more fancy stuffs with pretty-printing from [almond / jupyter-scala](https://github.com/almond-sh/almond).


lihaoyi:
lgtm

PR #911, opened 2018-12-21
Fix build for early 2.12 versions

alexarchambault:
https://travis-ci.org/lihaoyi/Ammonite/jobs/471031887#L1969


alexarchambault:
Seems https://github.com/lihaoyi/Ammonite/pull/850 broke stuff only run on the CI after merges…

Issue #920, opened 2019-01-24
Ctrl-C is broken


lihaoyi-databricks:
It no longer seems to interrupt runaway while loops. @alexarchambault could it have been one of your changes that broke it?

alexarchambault:
Possibly https://github.com/lihaoyi/Ammonite/commit/84d48542654285a9268964e4571fe0a91ef335d8 (seems it broke between `1.2.0` and `1.2.1`). Looking at it…

alexarchambault:
And as I don't rely on that in almond, interrupts work fine there…

PR #921, opened 2019-01-24
Allow to interrupt user code via Ctrl-C again

alexarchambault:
Fixes https://github.com/lihaoyi/Ammonite/issues/920.

### DIFF
--- a/terminal/src/main/scala/ammonite/terminal/Terminal.scala
+++ b/terminal/src/main/scala/ammonite/terminal/Terminal.scala
@@ -36,7 +36,12 @@ object Terminal {
                filters: Filter,
                displayTransform: (Vector[Char], Int) => (fansi.Str, Int) = LineReader.noTransform)
                : Option[String] = {
+    TTY.stty("intr undef")
+    try {
       new LineReader(ConsoleDim.width(), prompt, reader, writer, filters, displayTransform)
         .readChar(TermState(LazyList.continually(reader.read()), Vector.empty, 0, ""), 0)
+    } finally {
+      TTY.stty("intr '^C'")
+    }
   }
 }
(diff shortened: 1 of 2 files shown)


lihaoyi:
no idea but ok

lihaoyi:
👍

alexarchambault:
@lihaoyi This is not a perfect fix. Now when one type Ctrl-C after having entered some code, without pressing enter, nothing happens.

alexarchambault:
Previously, the current line was left in the terminal, and a new cursor was printed below.

alexarchambault:
I'd like to have it behave like before too… Do you know what code is handling the Ctrl-C when accepting input? The `Signaller` in the second commit here should call it somehow. I guess it must be around `AmmoniteFrontEnd`, but I can't find it…

lihaoyi:
@alexarchambault not sure, might have been this guy? https://github.com/lihaoyi/Ammonite/blob/1.2.0/terminal/src/main/scala/ammonite/terminal/filters/BasicFilters.scala#L92

alexarchambault:
So the new commit restores a bit the behavior prior to https://github.com/lihaoyi/Ammonite/pull/853. But this runs commands before / after reading input, which slows down things a bit (like before https://github.com/lihaoyi/Ammonite/pull/853, but most of its other gains should still be there).

Ideally, a `Signaller("INT")` could be added around the call to initial call to `loop()` in `ammonite.repl.Repl`. The handler could then manage to tell the `LineReader` to do as if it received a Ctrl-C. When the Ctrl-C arrives, the `LineReader` is just reading input, so it should be interrupted somehow, or the Ctrl-C could be manually injected in its input stream…

alexarchambault:
Superseded by https://github.com/lihaoyi/Ammonite/pull/923.

PR #926, opened 2019-02-08
Add sehbang

regadas:
Right now the assembled script seems to not have any shebang. This works well in shells like zsh and bash but not in fish.

I believe that by adding this "will" fix it. I wasn't really able to run `mill -i amm[2.12.8].assembly` so apologies in advance.

Update:
Managed to test this and it solves the issue.


lihaoyi:
Not the right fix, unfortunately, as this breaks the windows .bat file

PR #933, opened 2019-03-07
Switch to coursier 1.1.0-M13-1

alexarchambault:
~(Waiting for it to be released and sync-ed to Maven Central.)~

### DIFF
--- a/amm/interp/src/main/scala/ammonite/interp/InterpAPI.scala
+++ b/amm/interp/src/main/scala/ammonite/interp/InterpAPI.scala
@@ -4,6 +4,7 @@ package ammonite.interp
 import ammonite.util.{Colors, Ref}
 import ammonite.runtime.APIHolder
 import ammonite.runtime.Evaluator.AmmoniteExit
+import coursier.util.Task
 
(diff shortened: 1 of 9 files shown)


alexarchambault:
`ivyCache` is removed as the new high level API of coursier now has also has such a cache (see `withFetchCache` in the changes here).

alexarchambault:
Going to test that a bit more locally, then merge I guess.

alexarchambault:
Start-up time was hit when replacing the ivymap by the fetch cache from coursier. Running
```bash
$ mill 'amm[2.12.8].assembly'
$ cat foo.sc
import $ivy.`org.apache.spark::spark-sql:2.4.0`
import org.apache.spark.sql._
println(SparkSession)
$ time java -jar out/amm/2.12.8/assembly/dest/out.jar foo.sc
```
was consistently resulting in around 1500 ms with the coursier fetch cache, and around 1400 ms without, and with Ammonite 1.6.4. I'll investigate later… (the call to coursier itself didn't seem to be the culprit… maybe it's just because there are more classes to load)

Issue #874, opened 2018-09-28
Ammonite caches dependencies too aggressively


betehess:
I am trying to get an Ammonite script to depend on *the latest version of an internal dependency*.

I have something that looks like that:

```scala
interp.repositories() = List(coursier.MavenRepository("..."), ...)
System.setProperty("coursier.ttl", "0s")
interp.load.ivy("foo.bar" %% "baz" % sys.env.getOrElse("VERSION", "0+"))

@

@main def entrypoint(args: String*): Unit = {
  foo.bar.baz.Main.main(args.toArray)
}
```

It looks like Ammonite is not able to detect that a new version was released. We currently ask the users of the script to either:
* `rm -Rf ~/.ammonite/cache/`
* use `VERSION="x.y.z"`

But it defeats the purpose of having users not to worry about knowing what the latest version is.

The `coursier.ttl` trick was suggested by @alexarchambault and he pointed me to https://github.com/coursier/coursier/blob/a2ef75abcbb2cbdbc421f22c55a4a2ab960687e2/cache/jvm/src/main/scala/coursier/Cache.scala#L1137 but it didn't fix the problem.

alexarchambault:
This is kind of the same issue as https://github.com/lihaoyi/Ammonite/issues/433 (whose fix was lost at some point, it seems).

I guess we can get rid of the `ivyCache` of Ammonite altogether… It's not that useful since the switch to coursier, and it can have unintended side effects, like this issue.

PR #940, opened 2019-03-12
Ensure we don't keep changing artifacts in ivymap

alexarchambault:
This makes `IvyThing.resolveArtifact` return a boolean along the list of artifacts, saying whether that list can be cached or not. This is then taken into account before writing things to the cache.

### DIFF
--- a/amm/interp/src/main/scala/ammonite/interp/Interpreter.scala
+++ b/amm/interp/src/main/scala/ammonite/interp/Interpreter.scala
@@ -618,11 +618,12 @@ class Interpreter(val printer: Printer,
           output = printer.errStream,
           hooks = resolutionHooks
         )match{
-          case Right(loaded) =>
+          case Right((canBeCached, loaded)) =>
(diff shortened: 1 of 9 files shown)


alexarchambault:
First commit somehow adds to https://github.com/lihaoyi/Ammonite/pull/451. It makes resolution results aren't kept in the ivymap if they contain some snapshot artifacts..

Second commit should fix https://github.com/lihaoyi/Ammonite/issues/874. Resolution results aren't kept in ivymap if an interval appears in the initial dependencies.

alexarchambault:
Note there should still be corner cases when things that shouldn't be cached are nonetheless.

In the first case, if one loads a snapshot dependency that has no JARs and that depends on non-snapshot things, then no snapshot JAR will end up in the result, and this will be cached.

In the second case, if one depends on a fixed version of a dependency, but that dependency depends on some other dependencies with an interval, things are cached too.

PR #944, opened 2019-03-20
Fix issue with class-based wrapper

alexarchambault:
Fixes https://github.com/almond-sh/almond/issues/334.

The "used earlier definitions" mechanism of the class-based wrapper was wrongfully nullifying some references to previous commands, even though some of their types were referenced. Which triggered NPEs in some cases (https://github.com/almond-sh/almond/issues/334).


alexarchambault:
Will probably go ahead and make a release with all the latest stuff.

lihaoyi:
sounds good

PR #951, opened 2019-04-09
Move Preprocessor implem to a non anonymous class 

alexarchambault:
This PR is a dummy refactoring I guess.

This moves the implem of `ammonite.interp.Preprocessor` to its own distinct class, `DefaultPreprocessor`, so that it can be tweaked / customized when using Ammonite as a library (I'm planning to do such things in [almond](https://github.com/almond-sh/almond) in particular).


alexarchambault:
Probably going to go ahead and merge that one. (And maybe do a release after that.)

PR #974, opened 2019-06-14
Do not automatically include sources with all dependencies (lihaoyi#763)

tyrius02:
Every dependency can include classifiers (tests, javadoc, sources). The line `.addClassifiers(Classifier.sources)` forces the inclusion of sources with _every_ dependency.
Removing this line prevents the automatic inclusion of sources that is causing strange issues (such as Hadoop libraries not functioning properly).
If you want sources with a dependency they'll need to be explicitly specified.


tyrius02:
Looks like I get to go fix some tests now.

PR #975, opened 2019-06-15
Injecting import hooks

Baccata:
Rather than having the import hooks hardcoded, this allows to inject them as a `Map[Seq[String], ImportHook]` from the main.

The aim at this is for ammonite's downstreams to have a mean of changing the behaviour of the import hooks. In particular, this should allow mill to inject its own version into the resolution logic, in order to avoid discrepencies of versions between mill and the loaded contrib-module / plugins

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,6 @@ project/.sbtserver
 tags
 nohup.out
 out
+.bloop
+.vscode
+.metals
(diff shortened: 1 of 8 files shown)


Baccata:
Relevant discussion to understand the rationale behind this change :

https://gitter.im/lihaoyi/mill?at=5cf51eca481ef4167bc278e8 

@lefou

lihaoyi:
looks good to me

PR #979, opened 2019-06-23
[WIP][Issue-977] Proposal for change of script semantics 

Baccata:
## The what and the why  

This is a POC branch supporting a proposal to change the semantics of ammonite scripts. 

The endgoal would be turn ammonite a minimalist build-tool for `.sc` files, and allow the implementation of a subset of the [build server protocol](https://github.com/scalacenter/bsp/blob/master/docs/bsp.md), which would enable interoperability with third party tools like Metals or Intellij.

It would also help towards [Haoyi's proposal](https://contributors.scala-lang.org/t/proposal-simplifying-the-scala-getting-started-experience/2978) to simplify the "getting started experience", 

In the current state, the main blocker (imho) is the heavy coupling between the compilation and the run of scripts. This gives great power to the user, but has the drawback of making hard for tooling to be built around ammonite scripts. In addition, the multi-stage scripts feature (enabled ammonite's current  compileRun loop) seem to be confusing to many people. 

I'd like to propose some different semantics, the implementation of which would be breaking, but a reasonable migration path would be offered. 

## Goals of the proposal

* make semantics clearer
* facilitate a path towards BSP integration. 

## Elements of the proposal : 

1. Plain removal of multi-stage scripts. No more "@" allowed.  
2. Renaming of `import $exec` to `import $predef`, which I believe is closer to what it actually does, namely making all definitions present in the imported file directly available to the script without further imports 
3. By default, avoid running scripts that are imported via `import $file` and `import $predef`. This means that mutating the compiler's settings in these imports would not have an effect on the compilation of the script importing them anymore. 
4. Addition of an `import $stage` addressing 3. Could do with a more intuitive name. The goal is to make it really clear that this code is executed when compiling its downstreams, so that users build correct expectations. 
5. Modification of the various API methods made available to scripts (configureCompiler, etc) to avoid surfacing details of the compiler, and change their implementation to aggregate information into a json-serialisable payload that could be saved to disk for caching purposes.  

## What's implemented so far : 

Points 2, 3 and 4. I have brought the call to `evalMain` (for scripts anyway) closer to the surface, to have more control over when it's called exactly. The code is ugly and it's just been mostly about bashing my head against the codebase to build some mental representation of it, whilst trying to achieve something interesting enough to justify a proposal, and minimal enough to not make a discussion too time consuming for the parties involved.


Baccata:
@lihaoyi, in the interest of failing-fast, this could do with early feedback :smile: .

lihaoyi:
Some comments:

- I recall we discussed using `predef{...}` blocks in order to allow for configuration of the compiler. You could define a predef block in the same script file for convenience, or define it in a separate file and import it using `import $file`. That would subsume the current `import $predef` setup you've proposed.

- With predef blocks, we would basically have a hard-coded two-stage script-running process:
    - All scripts are resolved first via the `import $file` statements
    - All predefs are then evaluated (in what order???)
    - All non-predef code in the scripts is then compiled
    - All non-predef code in the scripts is the run (or not, if you want to package into jars, run METALS, etc.)

- It's not clear why `configureCompiler` would need to be serializable; the predef blocks would run in the same JVM as the compiler. It's only execution that might run in a separate JVM, and those should generally not need to call `configureCompiler`

- I don't know what `import $stage` does

Baccata:
My bad, I should have addressed those points (in particular, the predef block)

> I don't know what import $stage does. 

That is my attempt at solving the same problem we had discussed with the `predef` block

#### Regarding "import $predef". 

I think `predef` stands for `predefined`, and therefore the concept of predefining values and functions is a separate one from the concept of `modifying the compiler's settings`, which is the problem that current multi-stage scripts solve. Therefore : I renamed `import $exec` to `import $predef`, as `$exec` allows to predefine values and make them available to the current scope, in a similar way that the Scala compiler uses `predef.scala` 

#### What is `import $stage` / Why no predef block  ? 

Because, after thinking about it, defining the semantics of what we had called a "predef" block was complicated, which in turn would make it confusing to the end user imho : 

* it looks like a good old method, but it isn't ? 
* is it subject to magic imports ? 
* do we allow magic imports inside it ? 
* what order are they evaluated in ? 
* is bespoke code gonna have to be implemented on editor side to handle it, considering a single file could find itself with two different classpaths to deal with, depending which part of the file you're editing ? 

Asking myself these questions, it became somewhat clear that an easier solution, both conceptually and implementation-wise, would be to use the magic import to allow for imports that are run before compilation. I called it `$stage` (maybe `$setup` would be a better name) . 

#### What's the point of making `configureCompiler` serialisable ? 

In the similar way that mill tasks relies on agressive caching to avoid redoing work un-necessarily, if a piece of code is run with the intention to mutate the compiler settings in ammonite, we might want to cache the result until that piece of code changes. 

Also, as far as I can tell, `configureCompiler` is used to mostly to tweak compiler options (I have not found another usecase in the wild). When reading documentation, most people will find compiler options in their string form. If a `def addCompilerOptions(options : Seq[String])` was added to replace the current form of `configureCompiler`, we could easily intercept it and dump it on disk for later usage. Same goes with the methods that add stuff to classpaths, add maven repositories ... 

In addition, these information (classpath / compiler options) are [relevant to the bsp](https://github.com/scalacenter/bsp/blob/master/docs/bsp.md#scalac-options-request). Making `configureCompiler` serialisable would allow for intercepting the data and communicate it to the bsp client.

lihaoyi:
I think for now, your focus on names is misguided. The important thing we discussed was changing the compilation/execution semantics. That is a difficult, significant semantic change, with large amounts of incompatibility, and hopefully significant wins such as METALS integration and pre-compilation. Renaming and reformatting a bunch of existing features gives us none of those things.

Here's how I propose we proceed: as a first pass, I'd like to see the semantic change to the compilation/execution semantics implemented with zero renamings or new syntax. No `import $predef`, no `import $stage`, no random renamings like `def map` to `def traverse`, no reformattings like `_(settings)` to `s => s(settings)`, no adding trailing commas. We stick with exactly the current syntax, `@`, and implement the new semantics on top of it, with a minimal diff from the current implementation. The semantic changes we discussed are significant and complicated enough they deserve a focused discussion/review free without the distraction of random cleanup work.

After the semantic changes are done, we can then fiddle with names and formatting as a separate pass and separate set of pull requests.

Baccata:
Fair comment. However, renaming asides, the point I'm trying to make is that, imho, a "predef" block is not a good solution to the problem of dynamically tweaking compiler settings, and that problem has to be addressed in the new semantics. 

Should that first pass leave the problem aside for later ?

lihaoyi:
Yeah, as long as it's possible to make the current syntax work at all with the new compilation/execution model, even if it's not that great, we can punt on new syntax until later.

Baccata:
Closing this to start fresh on next iteration

PR #982, opened 2019-06-24
Fix publish docs job on CI

alexarchambault:


alexarchambault:
Merging this, that should help make the other PRs green.

PR #984, opened 2019-06-30
Convert coursier bootstrap into a flat classpath + filter

lihaoyi:
We now construct a normal fat jar, but also have Mill build an `ammonite-api-whitelist.txt` file containing a list of valid classpath entries we would like to expose in the REPL. This list is plumbed through Ammonite as an opaque `classPathWhitelist: Set[Seq[String]]`, to two important places:

- A `WhiteListClasspath`, which limits what classes/resources the Scala compiler is able to see at compile time

- A `WhiteListClassLoader`, which limits what classes/resources are made available to the REPL at runtime

We perform the classpath filtering at runtime, in the Ammonite code, which lets us fall back to the original simple fat-jar-style launcher. This also gives us the flexibility for embedded versions of Ammonite to inject a different `classPathWhitelist` callback, e.g. to allow use of application classses while still hiding the Ammonite classpath. The `classPathWhiteList` is fed into the cache tags to ensure `--thin` scripts and REPLs have their own caches.

Only static classpath entries are filtered; you can still `import $ivy` things outside the filtered subset, e.g. if you want to load a different version of uPickle than the one that's filtered out by `--thin`, you can still do so. To do so, we also plumb through an `initialClassLoader`/`initialClassPath` so we can distinguish the static classpath entries from those that are loading by `import $ivy` in the course of a REPL/script execution

Having the possibility of external (application) classes overriding ammonite dependencies is not yet supported. The easiest thing to do might be to simply take the ammonite fat-jar and shading everything except the classes listed in `ammonite-api-whitelist.txt`

This should avoid the performance problems we were seeing where coursier's nested-jar arrangement was slowing things down due to repeated unzipping.

We also should no longer need the custom `test-runner` or `test-api` modules.

I also took the liberty to make use of the new `./mill` version-pinning launcher script, and consolidate the cross-version scala-compiler-related logic into a single `CompilerCompatibility.scala` file


alexarchambault:
@lihaoyi Are you sure about renaming `interp-api` and `repl-api` to `interpApi` and `replApi`? That leads to [really weird module names at the end](https://repo1.maven.org/maven2/com/lihaoyi/ammonite-replApi_2.13.0/) (and weird file paths in the repo too, like `amm/replApi/…`).

lihaoyi:
Not really sure, maybe we could try converting those subfolders into `repl/api/` folders? That's what we do in Mill, and gives nice `amm-repl-api` module names

PR #985, opened 2019-07-02
replApi/interpApi -> repl.apl/interp.api

lihaoyi:
Should help avoid capitol letters in the published module names


lihaoyi:
boo CI y u no pass :(

alexarchambault:
@lihaoyi That's basically green now (could be even greener with things like [this](https://github.com/coursier/coursier/blob/dfa8f3a43afda577721e8baf64e72a818ec519c1/appveyor.yml#L32-L34))

lihaoyi:
looks good to me!

PR #987, opened 2019-07-08
Tweak interpreter constructor

alexarchambault:
These changes should be inocuous, yet add back _some_ source compatibility with Ammonite 1.6.7 (allows downstream users to keep supporting 2.11 via 1.6.7, while still depending on the latest Ammonite for 2.12 and 2.13).


alexarchambault:
🤔

PR #997, opened 2019-08-07
Add 2.12.9 support

alexarchambault:


alexarchambault:
@lihaoyi If I cut a release right after that, should it be `1.6.10` or `1.7.0`?

PR #1001, opened 2019-08-16
Full cross scala version for runtime

alexarchambault:
It depends on interp-api, that is fully cross-versioned, so it should be fully cross-versioned itself too.

Without this, `com.lihaoy:ammonite-runtime_2.12` depends on `com.lihaoyi:ammonite-repl-api_2.12.9`, which makes `com.lihaoyi:ammonite_2.12.8` pulls both `com.lihaoy:ammonite-repl-api_2.12.8` and `com.lihaoy:ammonite-repl-api_2.12.9`.

### DIFF
--- a/build.sc
+++ b/build.sc
@@ -156,9 +156,10 @@ object amm extends Cross[MainModule](fullCrossScalaVersions:_*){
     )
   }
 
-  object runtime extends Cross[RuntimeModule](binCrossScalaVersions:_*)
+  object runtime extends Cross[RuntimeModule](fullCrossScalaVersions:_*)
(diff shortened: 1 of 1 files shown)


lihaoyi:
+1

alexarchambault:
Alternatively, `interp-api` and `repl-api` could be binary cross-versioned. They'll still depend on particular versions of scala-{library,compiler,reflect}, but these are usually forced down the line ([even in mill](https://github.com/lihaoyi/mill/blob/ca1def4d6f5c829701e04f86dacb720bb57f036f/scalalib/src/Lib.scala#L72-L83) it seems).

PR #1024, opened 2019-12-19
Fix @main method handling in scripts run via --class-based

lihaoyi:
For now we just hardcode support for the two built in `CodeWrapper` instances

Making it extensible is TBD if anyone wants to pick that up


lihaoyi:
b6ae3ffdb46dd7d8c920a656b73c9ac7c52fd49e

PR #1033, opened 2020-01-28
Fixes the link to the bootstrap script

several27:


notdryft:
`boostrap` should also be changed to `bootstrap` (missing t)

sake92:
Fixed in unknown commit..

PR #1036, opened 2020-02-26
Update scala-collection-compat version

AprilAtProtenus:
Update scala-collection-compat version so that libraries
depending on newer versions work in ammonite scripts.


lihaoyi:
looks like a flaky failure in the third party URL our test posts to, kicking off those tests again

sake92:
Outdated.

PR #1037, opened 2020-02-27
Fix Appveyor CI

alexarchambault:


alexarchambault:
For w/e reasons, `C:\projects\ammonite\amm\src\main\scala-2.12.10` ends up as `:\projects\ammonite\amm\src\main\scala-2.12.10` in `java.class.path` (no `C` at the beginning) in the launcher generated by `amm[2.12.10].launcher`.

It seems it's only this particular path (all the others have their `C`), and it seems the `C` is still there in `amm[2.12.10].sources` and `amm[2.12.10].runClasspath` (that I think `amm[2.12.10].launcher` uses).

alexarchambault:
Ok, with `2.13.1`, `C:\Users\appveyor\AppData\Local\Coursier\cache\v1\https\repo1.maven.org\maven2\com\lihaoyi\pprint_2.13\0.5.8\pprint_2.13-0.5.8-sources.jar` becomes `C:\Users\appveyor\AppData\Local\Coursier\cace\v1\https\repo1.maven.org\maven2\com\lihaoyi\pprint_2.13\0.5.8\pprint_2.13-0.5.8-sources.jar` (`cache` -> `cace`).

So it seems a character at a precise position gets stripped at some point…

PR #1047, opened 2020-03-23
Don't go through compiler to add bridge imports

alexarchambault:
This saves a few "Compiling…" steps when starting Ammonite with an empty cache.


alexarchambault:
Merging this one… I won't cut a release myself anyway.

PR #1052, opened 2020-03-24
Remove unnecessary parameter

alexarchambault:


alexarchambault:
Mmmhh… maybe ignore that one actually…

PR #1053, opened 2020-03-24
Add BSP support

alexarchambault:
So I think this is ready for review.

This PR proposes to add [BSP](https://github.com/build-server-protocol/build-server-protocol) support in Ammonite, in order to add support for Ammonite scripts in metals. It has a twin PR in metals, https://github.com/scalameta/metals/pull/1538.

The context of this work is https://github.com/scalameta/metals-feature-requests/issues/80, that I worked on for the Scala Center.

I'm going to upload demo GIFs on https://github.com/scalameta/metals/pull/1538.

### DIFF
--- a/amm/interp/src/main/scala-2.13.1/ammonite/interp/MakeReporter.scala
+++ b/amm/interp/src/main/scala-2.13.1/ammonite/interp/MakeReporter.scala
@@ -13,18 +13,21 @@ import scala.tools.nsc.reporters.FilteringReporter
 import scala.tools.nsc.typechecker.Analyzer
 
 object MakeReporter {
+
+  type Reporter = scala.tools.nsc.reporters.Reporter
+
   def makeReporter(errorLogger: => String => Unit,
                    warningLogger: => String => Unit,
                    infoLogger: => String => Unit,
-                   outerSettings: Settings) = {
+                   outerSettings: Settings): FilteringReporter = {
     new FilteringReporter {
-      def displayPrompt(): Unit = ???
+
       def doReport(pos: scala.reflect.internal.util.Position,
                    msg: String,
-                   severity: Severity): Unit = {
+                   severity: Severity): Unit =
         display(pos, msg, severity)
-      }
-      def display(pos: Position, msg: String, severity: Severity) = {
+
+      def display(pos: Position, msg: String, severity: Severity) =
         severity match{
           case ERROR =>
             Classpath.traceClasspathProblem(s"ERROR: $msg")
@@ -34,9 +37,8 @@ object MakeReporter {
           case INFO =>
             infoLogger(Position.formatMessage(pos, msg, false))
         }
-      }
 
-      val settings = outerSettings
+      def settings = outerSettings
     }
   }
 }
--- a/amm/interp/src/main/scala-2.13.1/ammonite/interp/MakeReporter.scala
+++ b/amm/interp/src/main/scala-2.13.1/ammonite/interp/MakeReporter.scala
@@ -16,10 +16,10 @@ object MakeReporter {
 
   type Reporter = scala.tools.nsc.reporters.Reporter
 
-  def makeReporter(errorLogger: => String => Unit,
-                   warningLogger: => String => Unit,
-                   infoLogger: => String => Unit,
-                   outerSettings: Settings): FilteringReporter = {
+  def makeReporter(errorLogger: (Position, String) => Unit,
+                   warningLogger: (Position, String) => Unit,
+                   infoLogger: (Position, String) => Unit,
+                   outerSettings: Settings): Reporter =
     new FilteringReporter {
 
       def doReport(pos: scala.reflect.internal.util.Position,
@@ -31,14 +31,13 @@ object MakeReporter {
         severity match{
           case ERROR =>
             Classpath.traceClasspathProblem(s"ERROR: $msg")
-            errorLogger(Position.formatMessage(pos, msg, false))
+            errorLogger(pos, msg)
           case WARNING =>
-            warningLogger(Position.formatMessage(pos, msg, false))
+            warningLogger(pos, msg)
           case INFO =>
-            infoLogger(Position.formatMessage(pos, msg, false))
+            infoLogger(pos, msg)
         }
 
       def settings = outerSettings
     }
-  }
 }
--- a/amm/interp/src/main/scala-not-2.13.1/ammonite/interp/MakeReporter.scala
+++ b/amm/interp/src/main/scala-not-2.13.1/ammonite/interp/MakeReporter.scala
@@ -14,10 +14,12 @@ import scala.tools.nsc.typechecker.Analyzer
 
 object MakeReporter{
 
+  type Reporter = AbstractReporter
+
   def makeReporter(errorLogger: => String => Unit,
                    warningLogger: => String => Unit,
                    infoLogger: => String => Unit,
-                   outerSettings: Settings) = {
+                   outerSettings: Settings): Reporter = {
     new AbstractReporter {
       def displayPrompt(): Unit = ???
 
--- a/amm/interp/src/main/scala-not-2.13.1/ammonite/interp/MakeReporter.scala
+++ b/amm/interp/src/main/scala-not-2.13.1/ammonite/interp/MakeReporter.scala
@@ -16,10 +16,10 @@ object MakeReporter{
 
   type Reporter = AbstractReporter
 
-  def makeReporter(errorLogger: => String => Unit,
-                   warningLogger: => String => Unit,
-                   infoLogger: => String => Unit,
-                   outerSettings: Settings): Reporter = {
+  def makeReporter(errorLogger: (Position, String) => Unit,
+                   warningLogger: (Position, String) => Unit,
+                   infoLogger: (Position, String) => Unit,
+                   outerSettings: Settings): Reporter =
     new AbstractReporter {
       def displayPrompt(): Unit = ???
 
@@ -27,15 +27,14 @@ object MakeReporter{
         severity match{
           case ERROR =>
             Classpath.traceClasspathProblem(s"ERROR: $msg")
-            errorLogger(Position.formatMessage(pos, msg, false))
+            errorLogger(pos, msg)
           case WARNING =>
-            warningLogger(Position.formatMessage(pos, msg, false))
+            warningLogger(pos, msg)
           case INFO =>
-            infoLogger(Position.formatMessage(pos, msg, false))
+            infoLogger(pos, msg)
         }
       }
 
       val settings = outerSettings
     }
-  }
 }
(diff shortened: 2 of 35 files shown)


alexarchambault:
(Quick update: this is almost ready, I'm in the midst of writing a more lengthy description.)

alexarchambault:
## Try it yourself

Open a sbt or mill project in vscode, with the metals extension enabled. Create `.vscode/settings.json`, and put the following in it:
```json
{
  "metals.serverVersion": "0.8.5-amm-7,
  "metals.customRepositories": [
    "central",
    "bintray:alexarchambault/metals-amm"
  ]
}
```

Then open or create an Ammonite script at the root of the project. (Scripts in sub-directories ought to work too, they're just less tested for now…) A "Ammonite script detected" dialog should open, validate with "Import". The statusbar should then display "Importing Ammonite script", then "Indexing", and lastly "Compiling [script-name].sc". Once it's done compiling, code completion / hover / code navigation should work, and you should get accurate diagnostics in the script.

alexarchambault:
### Why BSP

BSP allows to have metals to query the class path, the scalac options, the diagnostics, etc., of Ammonite scripts, and offer the full-fledged IDE experience thanks to that (with code navigation, completion, hover, accurate diagnostics, …). metals already interacts with projects through BSP, so everything's in place in it to get IDE support for tools supporting BSP.

alexarchambault:
### PR structure

The first commits of this PR (as of writing this, https://github.com/lihaoyi/Ammonite/pull/1053/commits/8990c04000d80ec5319e231b13eaf7524bd9d23d to https://github.com/lihaoyi/Ammonite/pull/1053/commits/74cab7738536e549ec83dcc43f069e3ff40aaac3) are mostly innocuous refactorings, preparing the ground for the next ones.

Then https://github.com/lihaoyi/Ammonite/pull/1053/commits/9b5dadaf36ad0ef1f4a8bdc148d03f6fbd10882a adds a new API to deal with scripts.

Later commits are about BSP proper, adding support for it and tweaking it.

#### Refactorings

https://github.com/lihaoyi/Ammonite/pull/1053/commits/8990c04000d80ec5319e231b13eaf7524bd9d23d and https://github.com/lihaoyi/Ammonite/pull/1053/commits/dc7ae56955114a840f29de3a034fe5077b10a796 slightly change how `import $ivy` are represented in memory, allowing to get the underlying dependency of `import $ivy`s. Later on, this is useful when a script depends on another one: we merge its dependencies with the current one, then run a new resolution.

https://github.com/lihaoyi/Ammonite/pull/1053/commits/56b4f5f9498d2d250f786159eed01ca71bf7abff is a small change that allows to match results more easily.

https://github.com/lihaoyi/Ammonite/pull/1053/commits/26302ef750338d60f53b2be72d483bd8b6df3146 factors some dependency-related logic from `Interpreter`. (Maybe `IvyThing` could be merged into it?)

https://github.com/lihaoyi/Ammonite/pull/1053/commits/4c4917850be3417cff0e80f4333351905bc4d40c and https://github.com/lihaoyi/Ammonite/pull/1053/commits/eba1e91202ea497b9cea4ed17e792d0dd28c4e84 allow to build and pass custom reporters to scalac. (BSP support gets diagnostics from compilation via a custom reporter.)

https://github.com/lihaoyi/Ammonite/pull/1053/commits/1695dfd22f592b350e57dd5dae555ba647c25f66 allows to disable the `LineModifierPlugin`. It seems it's flawed in some cases (in code like `2`, that gets wrapped as `val res0 = 2`, it ignores the `val res0 =` part). Using it could also lead to messed up semanticdbs. Instead, BSP support has its own logic to map positions in scala code to positions in sc files, that it uses for diagnostics, and to post-process semanticdbs.

https://github.com/lihaoyi/Ammonite/pull/1053/commits/9dd7b56ae31646310d53288091603fba0786b4ab allows to pass a file name that matches the package to scalac (like `ammonite/$file/main.scala`), which semanticdb picks up later on.

https://github.com/lihaoyi/Ammonite/pull/1053/commits/4799c940c8eca702ed060cf1e33f1ce850a8f6bc allows to setup bridge APIs in the new script API and BSP, only for compiling, without actual bridge instances.

https://github.com/lihaoyi/Ammonite/pull/1053/commits/2571755a2bdd5c263d4181ed13c9c2bc6136b265 is useful when keeping case classes with `Imports` fields in `Map`s.

https://github.com/lihaoyi/Ammonite/pull/1053/commits/f3795489d740f1c576ff67913fe10ba4a032e71f allows not to ignore empty lines at the beginning of files. It seems `leadingSpaces` misses those, which messes with position conversion in scala code to sc scripts.

https://github.com/lihaoyi/Ammonite/pull/1053/commits/72d841f36d995c4aa0e9c0c3f252a72339ec5d2d marks generated code in scala code, like `2` becomes `/*<amm>*/val res0 = /*</amm>*/2`, which is used in the logic converting positions in scala code to positions in sc scripts.

https://github.com/lihaoyi/Ammonite/pull/1053/commits/d9ca11c2dd0880c76ccf6c2519f4634a66985f0a factors a method re-used in BSP support.

https://github.com/lihaoyi/Ammonite/pull/1053/commits/74cab7738536e549ec83dcc43f069e3ff40aaac3 puts the actual script code between `/*<script>*/` and `/*</script>*/` in generated scala code, and puts the generated `def $main` methods between `/*<generated>*/` and `/*</generated>*/`. That allows some logic in metals to read those scala files, discard the `/*<generated>*/` section, and replace the `/*<script>*/` one with a possibly unsaved content of the script. That allows metals to get valid Scala code from scripts itself, that it can pass to the presentation compiler for completions, hover, etc.

#### New script API

https://github.com/lihaoyi/Ammonite/pull/1053/commits/9b5dadaf36ad0ef1f4a8bdc148d03f6fbd10882a adds a new API to handle scripts. For now, this API only parallels the one in `Interpreter`. `Interpreter` doesn't use this new API, and keeps handling scripts its own way.

Compared to the way scripts are handled in `Interpreter`, this new API
- allows one to inspect script dependencies between each others and inspect their library dependencies, without compiling or running them. This in turn allows to answers BSP requests (`workspace/buildTargets`, `buildTarget/sourceDependencies`, etc.), without compiling scripts, and
- has a slightly different semantics (see above).

It has some drawbacks compared to `Interpreter` for now:
- it has much less aggressive caching
- it doesn't allow to run scripts, it only inspects and compile them.

#### BSP support

https://github.com/lihaoyi/Ammonite/pull/1053/commits/6f0efa04262fc8fd703a30a45da022799f182d68 adds `AmmoniteBuildServer`, a `ch.epfl.scala.bsp4j.BuildServer` implementation, in the `interp` module. It relies heavily on the newly added script API.

https://github.com/lihaoyi/Ammonite/pull/1053/commits/b2e4181e8a25f75dce25737327eb662b4642862b adds tests for it.

https://github.com/lihaoyi/Ammonite/pull/1053/commits/0a12034add3046bbca78407995df54d0d12f6abb adds a `--bsp` option to the Ammonite CLI, that starts `AmmoniteBuildServer` (listening and answering JSON-RPC requests on stdin / stdout).

Lastly, https://github.com/lihaoyi/Ammonite/pull/1053/commits/ff70563d1afc0cfa1d9ef2cec9c72945038d7558 allows to generate [SemanticDBs](https://scalameta.org/docs/semanticdb/guide.html) via the script API, and enables that in `AmmoniteBuildServer`.

alexarchambault:
### Script semantics discrepancies

The script API, used by the BSP server implementation here, has slightly different semantics compared to the one of `Interpreter`.

I mapped Ammonite scripts to the BSP model the following way: Each script is a "build target". A build target is kind of like a module in mill or sbt. An `import $file` adds a dependency towards another script, just like `moduleDeps` in mill adds a dependency towards another module. An `import $ivy` adds a dependency towards an external library, just like `ivyDeps` in mill. Depending on another script pulls its dependencies, just like in mill or sbt. And like in mill or sbt, dependency resolution is run for each script / module.

So the semantics of scripts through BSP is a little different than the one Ammonite currently uses to run scripts (handled mostly in `Interpreter.scala`). In particular, the class path of scripts are more isolated from each others in BSP. A script only pulls the scripts of its dependencies, and know nothing of the one of its dependees. For example, the following scripts are valid when running them with the current Ammonite model:
```scala
// main.sc
import $ivy.`com.chuusai::shapeless:2.3.3`
import $file.lib1, lib1.hlist
println(hlist)

// lib1.sc
case class Foo(n: Int)
val gen = shapeless.Generic[Foo]
val hlist = gen.to(Foo(2))
```
whereas they're invalid when viewed through BSP (because `lib1.sc` references shapeless without importing it via an `import $ivy`, and unlike with the current Ammonite model, the class path of `main.sc` doesn't "leak" into the one of `lib1.sc`, that doesn't even import it).

lihaoyi:
Thanks @alexarchambault ! I'll try to find some time to review this

lihaoyi:
@alexarchambault I just skimmed through and this looks reasonable to me.

I think in the long term, your new inter-script dependencies semantic is the way I would like to go forward: scripts would only have access to things they directly import, rather than having access to whatever happens to have been `import $ivy`ed or `import $file`d at the moment they are run. The semantic you are proposing for script BSP handling matches much better what I would expect when I see a script import another

This would mean `SpecialClassLoader` would have to maintain a Direct Acyclic Graph of classloaders, rather than the current linear linked list. Ideally we would refactor `Interpreter` to share as much code as possible between the REPL and the new Script runner, rather than having completely parallel implementations

We do not need to do this in this PR; just something to look forward to in future. I think for now this PR looks good to me

alexarchambault:
> I think in the long term, your new inter-script dependencies semantic is the way I would like to go forward: scripts would only have access to things they directly import, rather than having access to whatever happens to have been import $ivyed or import $filed at the moment they are run. The semantic you are proposing for script BSP handling matches much better what I would expect when I see a script import another.

Agreed. I didn't take the time to do it, but I agree this is the direction Ammonite should go into in the long term.

> This would mean SpecialClassLoader would have to maintain a Direct Acyclic Graph of classloaders, rather than the current linear linked list.

We don't necessarily need a class loader DAG. To me, this is kind of like when we have several modules in mill or sbt. In Ammonite, let's say we have an entrypoint script, that may reference code from other scripts. When we want to run that, we just get the full class path of the entrypoint script, which includes all its library dependencies JARs and those of its script dependencies, and the target directories of its script dependencies and of the entrypoint script itself. And we can run whichever script we want (entrypoint one or its dependencies) from that single class path.

As long as each script is compiled in isolation, things can't really go wrong (not more wrong than when we run things from or generated by mill or sbt).

alexarchambault:
> how does your BSP script compiler handle multi stage scripts, with stages separated by `@`?

I didn't pay a close attention to them initially. My idea was to propose to add special imports for repositories and scalac options (``import $repo.`sonatype:snapshots` ``, ``import $scalac.`- feature` ``), so that multi-stage scripts are less needed. But I didn't take the time to do it.

It may be possible to make some metals features work fine with them. For code navigation, each block is compiled separately, so we get one semanticdb per block. We would need to merge all of them in the `.sc` file semanticdb. It's [not done yet](https://github.com/lihaoyi/Ammonite/pull/1053/commits/ff70563d1afc0cfa1d9ef2cec9c72945038d7558#diff-544debaa5aeb531b386070a91eb567cdR120) either. For completion or hover, metals would need to be able to extract just one block from the `.sc` file, to pass to the PC, which may not be trivial from metals…

If users do fancy things with `interp` in a block, to setup things for subsequent blocks, we could offer the possibility to run some blocks with a mock `InterpAPI`, get the additional dependencies or repositories, and use them when compiling the other blocks.

Having each block be a BSP build target may require some changes in metals too, so that it's aware that the build target of a `.sc` file depends on the position in it.

alexarchambault:
Merging this!

PR #1054, opened 2020-03-24
Simplify Interpreter creation…

alexarchambault:
…by passing bridges and predefs only to its `initializePredef` method. They're only used there.

That simplifies `Interpreter` creation by avoiding having to reference the interpreter instance when we create it, like
```scala
// before
val interp: Interpreter = // type annotation required
  new Interpreter(
    predefs = …,
    bridges = Seq(
      // reference interp to build bridges
    )
  )
interp.initializePredefs()

// after
val interp = // no type annotation needed
  new Interpreter(
    …
  )
val bridges = Seq(
  // use interp…
)
interp.initializePredefs(predefs, bridges)

```

### DIFF
--- a/amm/interp/src/main/scala/ammonite/interp/Interpreter.scala
+++ b/amm/interp/src/main/scala/ammonite/interp/Interpreter.scala
@@ -25,15 +25,6 @@ import coursierapi.{Dependency, Fetch}
  */
 class Interpreter(val printer: Printer,
                   val storage: Storage,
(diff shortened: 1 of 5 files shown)


alexarchambault:
Seeing Travis CI as running, but [it's actually green](https://travis-ci.org/github/lihaoyi/Ammonite/builds/666488386).

Going to merge this one, as these are innocuous changes…

PR #1058, opened 2020-04-03
Stop relying on flaky jsonplaceholder.typicode.com service on CI

alexarchambault:


alexarchambault:
Seems to cumbersome to setup…

PR #1060, opened 2020-04-10
Fix link to ammonite.repl.Main in docs

1015bit:
Fixes a broken link in the docs :)

### DIFF
--- a/readme/Repl.scalatex
+++ b/readme/Repl.scalatex
@@ -140,7 +140,7 @@
     You can also pass a string to the @code{Main} call containing any
     commands or imports you want executed at the start of every run, along
     with @lnk("other configuration",
-    "http://www.lihaoyi.com/Ammonite/api/repl/index.html#ammonite.Main").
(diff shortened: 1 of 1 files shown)


alexarchambault:
Thanks for opening this @pbvie!

It seems what this links to is actually stale, `ammonite.Main` [exists](https://github.com/lihaoyi/Ammonite/blob/1d5868d0dce6bbf9c11ad00c92916ee0938ba3b1/amm/src/main/scala/ammonite/Main.scala), whereas `ammonite.repl.Main` doesn't anymore.

@lihaoyi Are the [API docs](http://ammonite.io/api/repl/index.html#ammonite.repl.Main) under http://ammonite.io/api supposed to be updated?

lihaoyi:
@alexarchambault I don't remember whether the API docs were uploaded automatically or manually. Feel free to do whatever!

mpkocher:
https://ammonite.io/#Scaladoc

These perhaps also need to be resolved/fixed?

PR #1065, opened 2020-04-23
Bump Scala version 2.13.2

lihaoyi:


amm/interp/src/main/scala-2.13.2/ammonite/interp/MakeReporter.scala
```diff
@@ -0,0 +1,43 @@
+package ammonite.interp
```
alexarchambault, line 1: Seems adding that file is unnecessary.

build.sc
```diff
@@ -19,11 +19,11 @@ val commitsSinceTaggedVersion = {
 }
 
 
-val binCrossScalaVersions = Seq("2.12.11", "2.13.1")
-val isScala2_12_10OrLater = Set("2.12.10") ++ binCrossScalaVersions
+val binCrossScalaVersions = Seq("2.12.11", "2.13.2")
+val isScala2_12_10OrLater = Set("2.12.10", "2.12.11", "2.13.0", "2.13.1", "2.13.2")
```
alexarchambault, line 23: We didn't want `2.13.0` in that list IIRC, don't remember why…

alexarchambault:
The semantic db version in [`AmmoniteBuildServer`](https://github.com/lihaoyi/Ammonite/blob/d723a938b7041267a5ae257ead7d15414fb0d096/amm/interp/src/main/scala/ammonite/interp/script/AmmoniteBuildServer.scala#L67) needs to be bumped to `4.3.9` I think.

amm/interp/src/main/scala-2.13.2/ammonite/interp/MakeReporter.scala
lihaoyi: Pretty sure we need it, or we need to reconfigure the build to have a `2.13.1+/` folder

lihaoyi:
@alexarchambault travis is green, this should be ready for review

Issue #1068, opened 2020-04-25
Ammonite embed in spring boot application very slow problem.


zheh12:
I embed ammonite in a spring application use such code to start repl.
```java
@SpringBootApplication
public class DemoApplication implements CommandLineRunner {

	public static void main(String[] args) {
		SpringApplication.run(DemoApplication.class, args);
	}

	@Override
	public void run(String... args) throws Exception {
		ammonite.Main$.MODULE$.main(new String[]{"-w"});
	}
}
```
And package a jar to start. The repl function well. But It is really slow, for only `1 + 1`, it use 5 seconds 
to get Ans.

But If not use a fat jar to run, it work well.

I found the problem why it is so slow. The most slow part is compile the class.

For the spring boot fat jar, there is a custom jar package method, so there is a lot jar in jar. So for ammonite, there will be a lot classpath like this to use CustomZipAndJarFileLookupFactory to do the job.
```
    jar:file:/Users/my/code/tests/demo/target/demo-0.0.1-SNAPSHOT.jar!/BOOT-INF/classes!/,
  jar:file:/Users/my/code/tests/demo/target/demo-0.0.1-SNAPSHOT.jar!/BOOT-INF/lib/spring-boot-starter-2.2.6.RELEASE.jar!/,
  jar:file:/Users/my/code/tests/demo/target/demo-0.0.1-SNAPSHOT.jar!/BOOT-INF/lib/spring-boot-2.2.6.RELEASE.jar!/,
  jar:file:/Users/my/code/tests/demo/target/demo-0.0.1-SNAPSHOT.jar!/BOOT-INF/lib/spring-context-5.2.5.RELEASE.jar!/,
  jar:file:/Users/my/code/tests/demo/target/demo-0.0.1-SNAPSHOT.jar!/BOOT-INF/lib/spring-boot-autoconfigure-2.2.6.RELEASE.jar!/,
  jar:file:/Users/my/code/tests/demo/target/demo-0.0.1-SNAPSHOT.jar!/BOOT-INF/lib/spring-boot-starter-logging-2.2.6.RELEASE.jar!/,
  jar:file:/Users/my/code/tests/demo/target/demo-0.0.1-SNAPSHOT.jar!/BOOT-INF/lib/logback-classic-1.2.3.jar!/,
  jar:file:/Users/my/code/tests/demo/target/demo-0.0.1-SNAPSHOT.jar!/BOOT-INF/lib/logback-core-1.2.3.jar!/,
[...]

PR #1069, opened 2020-04-25
Fix Spring embed slow repl problem

zheh12:
The issue about the patch try to fix is #1068

### DIFF
--- a/amm/interp/src/main/scala/ammonite/interp/internal/CustomURLZipArchive.scala
+++ b/amm/interp/src/main/scala/ammonite/interp/internal/CustomURLZipArchive.scala
@@ -169,11 +169,22 @@ final class CustomURLZipArchive(val url: java.net.URL) extends AbstractFile with
   }
 
 
-  def allDirsByDottedName: collection.Map[String, DirEntry] = {
-    dirs.map {
-      case (k, v) =>
-        k.mkString(".") -> v
-    }
+  lazy val allDirsByDottedName: collection.Map[String, DirEntry] = {
+    val springBootInfoPath = "BOOT-INF.classes"
+    val isSpringBootClassFile =
+      url.getFile != null && url.getFile.contains(springBootInfoPath.replace('.', '/'))
+
+    dirs
+      .map { case (k, v) => k.mkString(".") -> v }
+      .collect {
+        case (path, v) if isSpringBootClassFile &&
+          path.startsWith(springBootInfoPath) && path != springBootInfoPath => {
+          val newPath = path.substring(springBootInfoPath.length + 1)
+          newPath -> v
+        }
+        case (path, v) if !isSpringBootClassFile =>
+          path -> v
+      }
   }
 
 }
--- a/amm/interp/src/main/scala/scala/tools/nsc/CustomZipAndJarFileLookupFactory.scala
+++ b/amm/interp/src/main/scala/scala/tools/nsc/CustomZipAndJarFileLookupFactory.scala
@@ -18,6 +18,8 @@ import scala.tools.nsc.util.{ClassPath, ClassRepresentation}
 
 object CustomZipAndJarFileLookupFactory {
 
+  private val cache = collection.mutable.HashMap[URL, ClassPath]()
+
   private final class ZipArchiveClassPath(val zipUrl: URL)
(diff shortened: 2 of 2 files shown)


zheh12:
@alexarchambault @lihaoyi  When you have time, please give a look at the patch, and give some suggestion.

zheh12:
@alexarchambault  how can I let the ci rerun, I think there is some problem of the ci.

amm/interp/src/main/scala/ammonite/interp/internal/CustomURLZipArchive.scala
```diff
@@ -169,11 +175,22 @@ final class CustomURLZipArchive(val url: java.net.URL) extends AbstractFile with
   }
 
 
-  def allDirsByDottedName: collection.Map[String, DirEntry] = {
-    dirs.map {
-      case (k, v) =>
-        k.mkString(".") -> v
+  lazy val allDirsByDottedName: collection.Map[String, DirEntry] = {
```
alexarchambault, line 178: I believe this would be slightly more principled:
```scala
lazy val allDirsByDottedName: collection.Map[String, DirEntry] = {

  val springBootInfoPath = "BOOT-INF.classes"
  val isSpringBootClassFile =
    url.getFile != null && url.getFile.contains(springBootInfoPath.replace('.', '/'))

  dirs
    .map { case (k, v) => k.mkString(".") -> v }
    .collect {
      case (path, v) if isSpringBootClassFile && path != springBootInfoPath =>
        val newPath = path.substring(springBootInfoPath.length + 1)
        newPath -> v
      case (path, v) if !isSpringBootClassFile =>
        path -> v
    }
}
```

alexarchambault:
To give more context, my understanding is that this is necessary when Ammonite is depended on and started from an app packaged as a [spring boot executable JAR](https://docs.spring.io/spring-boot/docs/current/reference/html/appendix-executable-jar-format.html#executable-jar-jar-file-structure). https://github.com/lihaoyi/Ammonite/pull/850 must have made the JARs under `BOOT-INF/lib` work fine. Some of the classes of this file live under `BOOT-INF/classes`, so they need some workarounds to be found by scalac.

zheh12:
> To give more context, my understanding is that this is necessary when Ammonite is depended on and started from an app packaged as a [spring boot executable JAR](https://docs.spring.io/spring-boot/docs/current/reference/html/appendix-executable-jar-format.html#executable-jar-jar-file-structure). #850 must have made the JARs under `BOOT-INF/lib` work fine. Some of the classes of this file live under `BOOT-INF/classes`, so they need some workarounds to be found by scalac.

There is a long story with spring boot custom class loader. A source code for the classloader is 
on the https://github.com/spring-projects/spring-boot/tree/master/spring-boot-project/spring-boot-tools/spring-boot-loader/src/main/java/org/springframework/boot/loader.  

I will try to explain what i know. Spring boot loader use a special custom jar structure to package a application. It use jar in jar structure which is not a standard jar file. So spring boot do a lot hack thing to make it work. Which most import is A special custom classLoader `LaunchedURLClassLoader`, a special jar url handler `jar.loader.Handler` and a special jar file format `loader.jar.JarFile`.

So spring boot can process the special jar url such as

```
  jar:file:/Users/my/code/tests/demo/target/demo-0.0.1-SNAPSHOT.jar!/BOOT-INF/classes!/,
  jar:file:/Users/my/code/tests/demo/target/demo-0.0.1-SNAPSHOT.jar!/BOOT-INF/lib/spring-boot-starter-2.2.6.RELEASE.jar!/,
  jar:file:/Users/my/code/tests/demo/target/demo-0.0.1-SNAPSHOT.jar!/BOOT-INF/lib/spring-boot-2.2.6.RELEASE.jar!/,
```

So I think a lot way to work it out. But finally only the simple way to work it out is use the hard code. I know the code for the spring boot process such as use `BOOT-INF/classes` is not very elegant. But It is the simple way just can work for the special case of spring-boot.

amm/interp/src/main/scala/ammonite/interp/internal/CustomURLZipArchive.scala
zheh12: goog point, thanks

Issue #1070, opened 2020-04-25
Can we have a better api for other jvm language, such as java/kotlin/clojure?


zheh12:
I think now the api is Main class, but it is a scala code. We can simple use the scala api in other jvm language. So I want to give a java api to wrap the scala code, so other jvm language can have a better experience to embed the ammonite repl in the application. How do you think about the idea 
@alexarchambault  @lihaoy ? If you think this will be a good idea, I will give a try.

zheh12:
I write a basic java wrapper for self use https://github.com/lihaoyi/Ammonite/pull/1071. The usage is 
```
@SpringBootApplication
public class DemoApplication implements CommandLineRunner {
	@Autowired
	private ApplicationContext context;

	@Autowired
	private TestController testController;

	public static void main(String[] args) {
		SpringApplication.run(DemoApplication.class, args);
	}

	@Override
	public void run(String... args) {
		System.out.println("run ammonite repl");
		JavaMain repl = new Builder().build();
		repl.runWithWatch(new JavaBind<?>[]{
				new JavaBind<>("context", context),
				new JavaBind<>("controller", testController)
		});
	}
}
```

lihaoyi:
This is out of scope for the Ammonite codebase, but feel free to publish it as a separate module

PR #1071, opened 2020-04-26
better jvm language support

zheh12:
for issue  #1070


lihaoyi:
This is out of scope for the Ammonite codebase, but feel free to publish it as a separate module

PR #1072, opened 2020-05-01
Tweak BSP support

alexarchambault:
Cache compilation results in memory, so that we don't spend time recompiling everything all the time, in particular.


alexarchambault:
Going to merge that one if the CI is green. This basically only touches BSP-related stuff, mostly to fix issues raised in https://github.com/scalameta/metals/pull/1538.

PR #1076, opened 2020-05-05
Inline hard-coded predef imports

alexarchambault:
This PR replaces the hardcoded predef scripts (like [these](https://github.com/lihaoyi/Ammonite/blob/6e2245ec489eef51e7971e89205aa3dc8d1dbbfa/amm/repl/src/main/scala/ammonite/main/Defaults.scala#L28-L53) or [this one](https://github.com/lihaoyi/Ammonite/blob/d8b22aa47bcc0f26f8d435f73c8e98b401621932/amm/src/main/scala/ammonite/Main.scala#L358-L361)), by hardcoded `ImportData`. This is made possible by the hardcoded predef scripts consisting only of imports.

This allows to scrap the frequent upfront predef compilations, like
```
Compiling (synthetic)/ammonite/predef/DefaultPredef.sc
```
(which happen when starting a new Ammonite version, or when starting Ammonite for the first time in a given directory)

### DIFF
--- a/amm/repl/src/main/scala/ammonite/main/Defaults.scala
+++ b/amm/repl/src/main/scala/ammonite/main/Defaults.scala
@@ -31,19 +31,9 @@ object Defaults{
       "ammonite.interp.api.IvyConstructor.{ArtifactIdExt, GroupIdExt}",
       importType = ImportData.Type
     ),
-    ImportData("""ammonite.ops.{
-      PipeableImplicit,
-      FilterMapExtImplicit,
-      FilterMapArraysImplicit,
-      FilterMapIteratorsImplicit,
-      FilterMapGeneratorsImplicit,
-      SeqFactoryFunc,
-      RegexContextMaker,
-      Callable1Implicit
-    }"""),
     ImportData("ammonite.runtime.tools.{browse, grep, time}"),
     ImportData("ammonite.runtime.tools.tail", importType = ImportData.TermType),
-    ImportData("ammonite.repl.tools.{desugar, Desugared, source}")
+    ImportData("ammonite.repl.tools.{desugar, source}")
   )
 
 
--- a/amm/repl/src/main/scala/ammonite/main/Defaults.scala
+++ b/amm/repl/src/main/scala/ammonite/main/Defaults.scala
@@ -3,7 +3,7 @@ package ammonite.main
 
 import java.io.InputStream
 
-import ammonite.util.Util
+import ammonite.util.{ImportData, Imports, Util}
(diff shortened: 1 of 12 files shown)


amm/repl/src/main/scala/ammonite/main/Defaults.scala
```diff
@@ -25,32 +25,37 @@ object Defaults{
...
+      FilterMapExtImplicit,
+      FilterMapArraysImplicit,
+      FilterMapIteratorsImplicit,
+      FilterMapGeneratorsImplicit,
+      SeqFactoryFunc,
+      RegexContextMaker,
+      Callable1Implicit
+    }"""),
+    ImportData("ammonite.runtime.tools.{browse, grep, time}"),
+    ImportData("ammonite.runtime.tools.{GrepResult, Grepper, tail}", importType = ImportData.TermType),
```
alexarchambault, line 42: Maybe `GrepResult` and `Grepper` aren't actually needed here…

lihaoyi: Let's remove them!

lihaoyi:
This looks cool. If it works, then ship it

amm/repl/src/main/scala/ammonite/main/Defaults.scala
alexarchambault: Yes, confirmed locally they aren't needed.

amm/repl/src/main/scala/ammonite/main/Defaults.scala
```diff
@@ -25,32 +25,34 @@ object Defaults{
...
+    ImportData("ammonite.interp.api.IvyConstructor.{ArtifactIdExt, GroupIdExt}", importType = ImportData.Type),
+    ImportData("""ammonite.ops.{
+      PipeableImplicit,
+      FilterMapExtImplicit,
+      FilterMapArraysImplicit,
+      FilterMapIteratorsImplicit,
+      FilterMapGeneratorsImplicit,
+      SeqFactoryFunc,
+      RegexContextMaker,
+      Callable1Implicit
```
lihaoyi, line 39: IMO We should probably kill most of these as well. They were a cute experiment, but I've been moving away from the DSL-y operator-y style of ammonite.ops for most of my code, and this is a natural time to drop these imports. Anyone who wants them can get them by `import ammonite.ops._` anyway, which is no great hardship

alexarchambault: I'm all for that, let's try!

PR #1078, opened 2020-05-06
Tweak predef imports

alexarchambault:
This mainly merges `Default.predefImports` and `Main.extraPredefImports`. They were used together almost all the time. Except in some tests, where `Default.predefImports` was used alone, but adding `extraPredefImports` isn't a problem.

This also moves them to `Interpreter`, where they make a sensible default for `Interpreter.initializePredef` (and the BSP support will be able to re-use them from there).


alexarchambault:
Merging this, this will help addressing some issues seen in the BSP support.

Issue #729, opened 2017-12-21
ctrl-c to cancel tab-completion


kevincairo:
Sometimes tab-completion runs quite slowly, stalling for a few seconds or more. For example, on a Raspberry Pi 3, `new Fi-<tab><tab>` takes ~10 seconds to finish suggestions because it find 379 possible matches.

If there's an easy fix, it would be nice to be able to ctrl-c (or similar) to cancel the completion since it can be quicker to just type the symbol if you know it (and muscle memories leads to repeatedly initiating a tab-completion even though you know it'll take a while). 

Alternatively, if there are any low-hanging fruit for simply optimizing the tab completion, that would benefit even more people.

I'd be happy to take a crack if either approach seems doable and anyone can point me in the right direction. (Otherwise, no worries if this is a won't-fix.)

PR #1082, opened 2020-05-09
Handle SIGINT When Reading User Input

mgyucht:
This closes issue #729.

There are several issues with the interaction between the terminal, SIGINT, and signal handling:
* `Signaller` actually does not restore the previous signal handler in its finally block; it reinstalls the current handler (unless someone modifies `Signaller.handlers` directly). This seems to be simply a bug.
* Users cannot send SIGINT to Ammonite while the input line is being read in, including during auto completion. Instead, Ammonite interprets the control character literally. However, this behavior is still in place during autocompletion, so users cannot Ctrl-C to cancel autocompletion, as this doesn't send SIGINT.
* Even if users could send SIGINT to Ammonite when typing, the Signaller is only invoked after having read in the users' input.

This PR addresses all three issues: we correct the behavior of Signaller, modify the TTY configuration so Ctrl-C results in SIGINT during autocomplete, and move the Signaller invocation before the frontend call so that the SIGINT handler is registered during autocomplete.

A screenshot of the new behavior:
Screenshot 2020-05-10 at 01 26 34
(This was made by typing tab then Ctrl-C in quick succession.)

### DIFF
--- a/amm/repl/src/main/scala/ammonite/repl/AmmoniteFrontEnd.scala
+++ b/amm/repl/src/main/scala/ammonite/repl/AmmoniteFrontEnd.scala
@@ -50,7 +50,9 @@ case class AmmoniteFrontEnd(extraFilters: Filter = Filter.empty) extends FrontEn
 
     val autocompleteFilter: Filter = Filter.action(SpecialKeys.Tab){
       case TermState(rest, b, c, _) =>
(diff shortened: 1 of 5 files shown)


lihaoyi:
@kevinkyyro can you try building Ammonite from this branch and play around with it to see if it does what you expect?

PR #1090, opened 2020-06-04
Adds magic import to statically add maven/ivy repositories

Baccata:
This adds support for statically adding repositories for artifact resolution, via additional import hooks.   

Usage looks like this :

```scala 
import $repo.`https://some.repo/.../` // parses maven repo pattern
import $repo.`ivy:https://some.repo/.../` // parses ivy repo pattern 
```

Tested locally using a private artifactory 

Checklist : 

- [x] tests 
- [x] documentation


Baccata:
With the recent addition of ammonite support to metals, which relies on static dependency resolution, it is important to give users a way to statically add repositories used to resolve their artefacts.

See https://github.com/scalameta/metals/issues/1766#issuecomment-631519887

@alexarchambault, will anything be required in the BSP side of things ?

alexarchambault:
For the BSP aspects, you should ensure that the repositories from the imports are added to `Dependencies` in [this method](https://github.com/lihaoyi/Ammonite/blob/521452fdf67b2f7e71ba855c30b9c925bf14ad4c/amm/interp/src/main/scala/ammonite/interp/script/Script.scala#L102). Don't hesitate to add some tests in [`AmmoniteBuildServerTests`](https://github.com/lihaoyi/Ammonite/blob/master/amm/src/test/scala/ammonite/interp/script/AmmoniteBuildServerTests.scala), checking for example that if you depend on a script, all its repositories are used in the current one too.

amm/runtime/src/main/scala/ammonite/runtime/ImportHook.scala
```diff
@@ -39,6 +39,8 @@ object ImportHook{
     Seq("url") -> URL,
     Seq("file") -> File,
     Seq("exec") -> Exec,
+    Seq("repo", "maven") -> MavenRepo,
```
alexarchambault, line 42: I would have gone for just `$repo`, like
```scala
import $repo.`https://jcenter.bintray.com`
import $repo.`ivy:…`
```

In a first time, you could go for Maven or Ivy by checking if the repository string starts with `ivy:`.

coursier has [a method](https://github.com/coursier/coursier/blob/3f310cad337ff44cae1490618a717904b3b9a429/modules/coursier/jvm/src/main/scala/coursier/internal/PlatformRepositoryParser.scala#L14) (accessible via `coursier.parse.RepositoryParser.repository`) to parse both Maven and Ivy repositories at the same time. It needs to be made accessible in [coursier/interface](https://github.com/coursier/interface) so that it could later be used here too.

Baccata:
> you probably need to add a Seq[Repository] field to Result.ClassPath for that, I guess.

Actually I've started adding another ADT member (`Repo`) to `Result` to achieve the same thing

amm/runtime/src/main/scala/ammonite/runtime/ImportHook.scala
Baccata: Happy to change it :+1:

Baccata: Changed in https://github.com/lihaoyi/Ammonite/pull/1090/commits/40b0256bf7d4f193b482a751ff6f4b208ae7fa93

amm/interp/src/main/scala/ammonite/interp/Interpreter.scala
```diff
@@ -195,6 +194,10 @@ class Interpreter(val printer: Printer,
           else headFrame.addClasspath(res.files.map(_.toNIO.toUri.toURL))
 
           Res.Success(Imports())
+
+        case res: ImportHook.Result.Repo =>
+          Res.Success(Imports())
```
alexarchambault, line 199: Wouldn't it work to just call `repositories.update` here, instead of adding `addRepository`?

Baccata: Hah, good point, I'll change it

Baccata: https://github.com/lihaoyi/Ammonite/pull/1090/commits/e532f58e3e73812713a936b91b67b2c34fe3e97a

amm/src/test/resources/bsp/multi/lib3.sc
```diff
@@ -0,0 +1,5 @@
+// See https://github.com/Baccata/gradle-simple/blob/master/src/main/java/Hello.java
+import $repo.`https://jitpack.io`
+import $ivy.`com.github.baccata:gradle-simple:63f75f7133`
```
Baccata, line 3: needed a java library to avoid having to cross-compile to various Scala versions.

The original https://github.com/jitpack/gradle-simple only has not-package class containing a main method. I'm not sure why, but ammonite didn't seem to be happy with it. Forked it to add package + remove main. 

I'm happy to transfer the repo's ownership to whatever relevant github organisation (coursier or haoyi's) if you guys prefer

Baccata:
I don't think the CI errors are related to anything added by this PR.

alexarchambault:
The `noLongLines` check fails on the CI. Locally, you can run
```bash
$ ./mill 'ops[2.13.2].test' test.ammonite.ops.ExampleTests.noLongLines
```
then shorten the lines it points at.

amm/src/test/resources/scripts/ResolversStatic.sc
```diff
@@ -0,0 +1,4 @@
+import $repo.`ivy:https://ambiata-oss.s3-ap-southeast-2.amazonaws.com/[defaultPattern]`
```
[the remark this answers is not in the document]
Baccata: Replaced in 68d71e0

amm/src/test/resources/bsp/multi/lib3.sc
Baccata: Replaced in 68d71e0

amm/src/test/scala/ammonite/session/ScriptTests.scala
```diff
@@ -336,6 +336,17 @@ object ScriptTests extends TestSuite{
           """)
         }
       }
+      test("resolverStatic"){
+        test("pass"){
+          if (scala2_11) check.session(s"""
```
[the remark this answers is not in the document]
Baccata: Removed the check in 68d71e0

Baccata:
I think this is ready for another review pass 😸

PR #1104, opened 2020-07-14
Add 2.12.12 support

ckipp01:
I can't add reviewers, so I'm just going to ping you @alexarchambault. We're looking to release Metals today if possible with 2.12.12 support, so we figured it'd also be nice to have the default Ammonite version be the same for users. Let me know if I missed anything.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,12 +12,12 @@ matrix:
   - stage: build
     env: CI_SCRIPT="./mill -i unitTest"
     jdk: openjdk11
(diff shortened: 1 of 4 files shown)


ckipp01:
I have some legit issues it seems, looking into them.

ckipp01:
Assumed that this was going to be a quick upgrade, but there is some interop stuff that I'm not familiar with at all in Ammonite so I'm just going to close this. 😬

PR #1105, opened 2020-07-14
Add scala 2.12.12 support

alexarchambault:
Includes https://github.com/lihaoyi/Ammonite/pull/1104.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,12 +12,12 @@ matrix:
   - stage: build
     env: CI_SCRIPT="./mill -i unitTest"
     jdk: openjdk11
-    scala: 2.12.11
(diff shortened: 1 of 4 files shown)


alexarchambault:
I think the extra commit compared to https://github.com/lihaoyi/Ammonite/pull/1104 should make the tests pass (or at least should fix the compilation errors seen there).

cc @ckipp01

ckipp01:
> I think the extra commit compared to #1104 should make the tests pass (or at least should fix the compilation errors seen there).
> 
> cc @ckipp01

ah jeez, it was seriously something that simple... Of course. Thanks for pointing this out to me though!

PR #1114, opened 2020-08-12
Fix Ammonite samples in the readme

### LINKED ISSUE #1102: Scala examples missing from docs(?)
Example: https://ammonite.io/#Subprocesses

Here's a close-up: The text suggests that just below the `bash` example, there is going to be an ammonite example. However, the example doesn't show up...
image

Full screenshot:
image

steinybot:
Fixes #1102 

It would be nice to have a regression test which would catch if this breaks again but I'm not sure where to put it.


sake92:
Fixed in #1205

PR #1123, opened 2020-11-18
Deprecate --no-remote-logging option

jackkoenig:
Remote logging functionality has already been removed

The functionality seems to have been removed in https://github.com/lihaoyi/Ammonite/commit/2fdd2dfe772d07cf59613a563167218936ae5013, released in version `1.6.4`.

Deprecating the field in the `Main` case class does cause some deprecation warnings in the codebase itself so perhaps that change is more annoying than it is useful. I mainly thought it prudent to remove the unused option from `--help`.


jackkoenig:
I didn't add any tests only because I wasn't sure where to start. I would appreciate a pointer!

alexarchambault (approved):
Thanks!

PR #1125, opened 2020-11-20
Add 2.13.4 support

ckipp01:
Just pr adds in support for 2.13.4 which we would like to include in our Metals release that supports 2.13.4.

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -28,7 +28,7 @@ jobs:
           java-version: '8'
       - run: ./mill -i unitTest 2.12.12
 
-  unit-8-2_13_3:
+  unit-8-2_13_4:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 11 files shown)


alexarchambault:
Just pushed an extra commit that will (hopefully) address the compilation errors. (That's because some conversions were marked as non explicit in `2.13.4`, look for `CharSequence` in the [`2.13.4` release notes](https://github.com/scala/scala/releases/tag/v2.13.4).)

ckipp01:
> Just pushed an extra commit that will (hopefully) address the compilation errors. (That's because some conversions were marked as non implicit in `2.13.4`, look for `CharSequence` in the [`2.13.4` release notes](https://github.com/scala/scala/releases/tag/v2.13.4).)

Hey! Thanks for taking a peak! I just threw this up here this morning before work hoping for the best 😬 , but I had planned on taking a look back after work. It looks like it's still failing, so if you do want to keep working on it, by all means please do! If not I'll try to take a look tonight at what's going on.

alexarchambault:
@ckipp01 Feel free to have a look later on. I'm not sure I'll be able to have a look today.

ckipp01:
So I don't really get what's going on here and why the changes are affecting those tests...

PR #1127, opened 2020-11-25
Rebootstrap Ammonite on latest Mill master

lihaoyi:


alexarchambault:
@lihaoyi You may want to move the `release*` jobs to separate workflows (so separate `.yml` files). For now, the GitHub actions UI doesn't allow to re-run individual jobs, if ever one of them fails. You can only re-run whole workflows. That could be a problem if one of the release jobs fails. We would have to re-run even the ones that succeeded, and they would fail too (attempting to release again artifacts that were already published). Whereas if they're in separate workflows, they could be re-run individually.

A drawback of that would be that the release jobs wouldn't wait for the tests to have run fine. But if we only merge green PRs, that shouldn't be a problem.

Issue #1130, opened 2020-12-01
Cannot build mill assembly from a release source archive


rolisv:
When building assembly from the source tarball:
```bash
$ curl -LO https://github.com/lihaoyi/Ammonite/archive/2.3.8.tar.gz
$ tar -xf 2.3.8.tar.gz
$ cd Ammonite-2.3.8
$ mill 'amm[2.13.3].assembly'
Compiling /home/ro/workspace/github/Ammonite-2.3.8/build.sc
fatal: not a git repository (or any parent up to mount point /)
Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).
os.SubprocessException: CommandResult 128

  os.proc.call(ProcessOps.scala:85)
  ammonite.$file.build.$init$(build.sc:11)
  ammonite.$file.build$.<clinit>(build.sc:40)
```
It looks like `build.sc` expects the source to be in git repo.

Is building from the release archive supported? I could not find any docs on that.

sake92:
What is the use case for this?  
You can just clone this repo and then run that command.

NatPRoach:
Just jumping in here to say one use case for adding support for this would be putting tools on to conda / conda-forge `https://conda-forge.org/` and other similar package / environment managers. As is attempting to build from a release tar ball for a package that uses `mill` to build it will not work since release tar balls are not git repos. And since `conda-forge` requires using release tarballs such that they can be checked vs a `sha256` / `md5` checksum'd this means that tools that use the mill build script to compile themselves at present cannot be distributed on conda-forge.

PR #1131, opened 2020-12-02
Allow building from release archive

rolisv:
Add support for `RELEASE_ARCHIVE_VERSION` environment variable to `build.sc`.
If `RELEASE_ARCHIVE_VERSION` is set, it assumes, that build is performed from the
release source archive and does not invoke `git` command, which otherwise would result
in `fatal: not a git repository` error.
`RELEASE_ARCHIVE_VERSION` should be set to the source archive version, e.g:
```
$ RELEASE_ARCHIVE_VERSION=2.3.8 mill -i amm[2.13.3].assembly
```

Addresses issue #1130

### DIFF
--- a/build.sc
+++ b/build.sc
@@ -8,12 +8,19 @@ val isMasterCommit =
   sys.env.get("GITHUB_REPOSITORY") == Some("lihaoyi/Ammonite") &&
   sys.env.get("GITHUB_REF").exists(x => x.endsWith("/master"))
 
-val latestTaggedVersion = os.proc('git, 'describe, "--abbrev=0", "--tags").call().out.trim
+val releaseArchiveVersion = sys.env.get("RELEASE_ARCHIVE_VERSION")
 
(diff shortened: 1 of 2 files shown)


alexarchambault:
@rolisv Would it work for you if the version would default to a literal `X.Y.Z` instead of requiring users to set `RELEASE_ARCHIVE_VERSION`? This shouldn't be much of a problem, unless you're actually packaging Ammonite for your own needs from the source archive?

rolisv:
@alexarchambault It would work, if that literal is set only in release archive tarball. We need some kind of marker in `build.sc` to know that we are building from release archive and not invoke git commands.

Arch Linux builds Ammonite from release archive sources and runs a bunch of sed scripts to "butcher" `build.sc` into not running any git commands. But if `build.sc` changes, those sed scripts need to be updated. The goal here is to be able to build Ammonite from a release archive without needing to patch it first.

PR #1133, opened 2020-12-03
Groundwork for Scala version decoupling

alexarchambault:
This PR prepares the ground for Scala version decoupling.

It mainly
- removes dead code that makes it easier to later evolve things
- uses simpler types in methods later meant to be part of a Java interface
- moves things to places where it's easier to evolve them later on
- try to depend less on os-lib and the ops module, especially in user-facing modules
- tweak some `--thin` related things, that helps for the upcoming `--scala …` option

About the second last point, I've been realizing that os-lib and ops module stuff may be added back later on, in the upcoming PR decoupling Scala versions. I'm finishing cleaning up the PR decoupling Scala versions, and that may allow to drop some of these os-lib and ops module related changes from here. I'm opening this PR to check what the CI says and for you to see, but in any case, let's wait until the upcoming Scala version decoupling PR is opened to merge this.


alexarchambault:
Closing this. Going to open new PRs with less changes.

PR #1135, opened 2020-12-04
Decouple the Scala version used by Ammonite from the one exposed to users

alexarchambault:
This currently includes https://github.com/lihaoyi/Ammonite/pull/1133, so it actually looks slightly lengthier than it is.

~Checking what the CI says in a first time, I'll describe it in more details right after.~


alexarchambault:
The changes introduced by this PR are quite large… There are some breaking changes in the early commits (those changing `os.Path` to `java.nio.file.Path` in the APIs), but I'm thinking of reverting those, so that I could break this PR in smaller chunks, that could be discussed and merged one after the other.

The overall goal is to make Ammonite interact with the user session and scalac only via Java APIs, and this is achieved by adding Java interfaces, in new pure Java modules:
- `ammonite.interp.api.InterAPI` becomes [a Java interface](https://github.com/alexarchambault/Ammonite/blob/3ade62ab870c60a751a4aaa2127abac1d94cbdc6/amm/interp/api/src/main/java/ammonite/interp/api/InterpAPI.java), and its module `interp-api` is now pure Java,
- `ammonite.repl.api.ReplAPI` becomes [a Java interface](https://github.com/alexarchambault/Ammonite/blob/3ade62ab870c60a751a4aaa2127abac1d94cbdc6/amm/repl/api/src/main/java/ammonite/repl/api/ReplAPI.java), and its module `repl-api` is now pure Java,
- new interfaces `ammonite.compiler.iface.{` [`CompilerLifeCycleManager`](https://github.com/alexarchambault/Ammonite/blob/3ade62ab870c60a751a4aaa2127abac1d94cbdc6/amm/compiler/interface/src/main/java/ammonite/compiler/iface/CompilerLifecycleManager.java), [`CompilerBuilder`](https://github.com/alexarchambault/Ammonite/blob/3ade62ab870c60a751a4aaa2127abac1d94cbdc6/amm/compiler/interface/src/main/java/ammonite/compiler/iface/CompilerBuilder.java), [`Parser`](https://github.com/alexarchambault/Ammonite/blob/3ade62ab870c60a751a4aaa2127abac1d94cbdc6/amm/compiler/interface/src/main/java/ammonite/compiler/iface/Parser.java) `}` are introduced in a new `compiler-interface` module.

Extension methods, such as [these](https://github.com/alexarchambault/Ammonite/blob/3ade62ab870c60a751a4aaa2127abac1d94cbdc6/amm/interp/api/full/src/main/scala/ammonite/interp/api/InterpBridge.scala#L12-L38) or [these](https://github.com/alexarchambault/Ammonite/blob/3ade62ab870c60a751a4aaa2127abac1d94cbdc6/amm/repl/api/full/src/main/scala/ammonite/repl/ReplBridge.scala#L188-L312), help make these interfaces more usable from Scala code.

## APIs

The original `interp-api` and `repl-api` modules actually get split in two: `interp-api` / `interp-api-full`, and `repl-api` / `repl-api-full`. `interp-api` and `repl-api` are pure Java. `interp-api-full` and `repl-api-full` are Scala modules. The `APIHolder`s and extension methods for the Java APIs live in the latter modules.

## `--scala`

The actual decoupling of Scala versions is enabled when `--scala X.Y.Z` is passed to the Ammonite launcher. This option takes over the former `--thin` option. `--scala X.Y.Z` re-uses the whitelisting from `--thin`, but only leaves pure Java classes in the whitelist (those of `interp-api` and `repl-api` in particular). This way, the initial interpreter ClassLoader only contains pure Java classes. To load a Scala version, and recover the former Scala methods of the APIs, we [fetch](https://github.com/alexarchambault/Ammonite/blob/3ade62ab870c60a751a4aaa2127abac1d94cbdc6/amm/src/main/scala/ammonite/Main.scala#L114-L133) `interp-api-full` and `repl-api-full` for Scala `X.Y.Z` upon startup, and add them to the interpreter ClassLoader just like libraries users would have added themselves. This way, these additional libraries are effectively isolated from Ammonite itself.

## Compiler

For the compiler, a new `compiler` Scala module has implementations for the interfaces of the pure Java `compiler-interface`. Just like for `interp-api-full` and `repl-api-full`, upon startup, we fetch `compiler` for Scala `X.Y.Z`, and add it to the interpreter ClassLoader. We then create instances of some of its classes [by reflection](https://github.com/alexarchambault/Ammonite/blob/3ade62ab870c60a751a4aaa2127abac1d94cbdc6/amm/interp/src/main/scala/ammonite/interp/Interpreter.scala#L826-L842), and cast them as one of `ammonite.compiler.iface.{CompilerLifeCycleManager, CompilerBuilder, Parser}`. We then [pass those instances](https://github.com/alexarchambault/Ammonite/blob/3ade62ab870c60a751a4aaa2127abac1d94cbdc6/amm/interp/src/main/scala/ammonite/interp/Interpreter.scala#L35-L36) to `Interpreter` in particular, which uses them to parse and compile things.

## Tests (only REPL tests for now)

About the tests, I only enabled this feature in REPL tests for now (but manual tests for scripts work).

To test that, a new [`amm.repl.cross-tests`](https://github.com/alexarchambault/Ammonite/blob/3ade62ab870c60a751a4aaa2127abac1d94cbdc6/build.sc#L304) module is added. Its sources are those of the tests of `amm.repl` as is (but for a boolean variable that enables cross testing). This module is defined as a `CrossModule`, but doesn't use the "cross-variable" as its Scala version. Instead, it assumes it's the Scala version that should be exposed to users in the tests. So it [sticks](https://github.com/alexarchambault/Ammonite/blob/3ade62ab870c60a751a4aaa2127abac1d94cbdc6/build.sc#L307-L308) to the main Scala version (currently, `2.13.3`), but [pulls](https://github.com/alexarchambault/Ammonite/blob/3ade62ab870c60a751a4aaa2127abac1d94cbdc6/build.sc#L338-L340) the class path of `interp-api-full` / `repl-api-full` / `compiler` for the cross Scala version, and passes it to the tests via [a Java property](https://github.com/alexarchambault/Ammonite/blob/3ade62ab870c60a751a4aaa2127abac1d94cbdc6/build.sc#L347). The tests then just [add those](https://github.com/alexarchambault/Ammonite/blob/3ade62ab870c60a751a4aaa2127abac1d94cbdc6/amm/repl/src/test/scala/ammonite/TestRepl.scala#L59-L66) to the interpreter class loader, just like how `--scala X.Y.Z` does. So
```text
$ ./mill amm.repl.cross-tests[2.12.12].test
```
runs the test sessions code in `2.12.12`, but the logic running the sessions runs itself in `2.13.13`.

alexarchambault:
To test `--scala X.Y.Z` locally, the following command should do:
```text
$ ./mill '__[2.12.12].publishLocal' &&\
    ./mill 'amm.interp.api.publishLocal' &&\
    ./mill 'amm.compiler.interface.publishLocal' &&\
    ./mill 'amm.repl.api.publishLocal' &&\
    ./mill -i 'amm[2.13.3].run' --scala 2.12.12
```

This publishes the modules that are loaded upon startup first (`interp-api-full`, `repl-api-full`, `compiler`, and their dependencies), along with the pure Java modules. `--scala X.Y.Z` can then pick those from the local Ivy2 repository.

alexarchambault:
To enable `--scala 3` support later on, only the modules that get fetched upon startup need to be cross-compiled for Scala 3. These are
- `interp-api-full`
- `repl-api-full`
- `compiler`
- `util`

Their dependencies are:
```text
$ cs resolve com.lihaoyi:ammonite-repl-api-full_2.13.3:2.3.8-71-f113c5050 | grep '2\.13' | grep -v '^com\.lihaoyi:ammonite' | grep -v '^org.scala-lang:'
com.lihaoyi:fansi_2.13:0.2.9:default
com.lihaoyi:fastparse_2.13:2.3.0:default
com.lihaoyi:geny_2.13:0.6.2:default
com.lihaoyi:mainargs_2.13:0.1.4:default
com.lihaoyi:os-lib_2.13:0.7.1:default
com.lihaoyi:pprint_2.13:0.5.9:default
com.lihaoyi:scalaparse_2.13:2.3.0:default
com.lihaoyi:sourcecode_2.13:0.2.1:default
org.scala-lang.modules:scala-collection-compat_2.13:2.1.4:default
org.scala-lang.modules:scala-xml_2.13:1.2.0:default
```

All of them but fastparse / scalaparse / mainargs are already cross-published for `3.0.0-M2`. fastparse / scalaparse are pulled by the `compiler` module, we should work around them by using the dotty parser.

mainargs for `3.0.0-M2` is missing. It's only needed for `@main` entrypoints, so I'm thinking about disabling those for Scala 3 in a first time, until we can get mainargs for it.

alexarchambault:
I'm hesitating discussing the various commits of this PR right now… @lihaoyi Just tell me if you're ok with the general direction, and feel like reviewing this (+4000 -2000 / 74 commits) PR, or you prefer me to split it in smaller PRs.

lihaoyi:
Thanks Alex! I won't be able to get to this in the next few days, but hopefully have time coming up next week to give this a proper review

alexarchambault:
I've come to realize that this PR actually deals with several successive concerns, even though things are not done in this order in the commit history:
- extracting compiler- and parsing-related code (mostly from `interp`) to a new distinct `compiler` module
- "Java-ize" the interface between the `interp` and `compiler` modules (no more direct dependency between both, instead `interp` expects an instance of a Java interface, that users can find an implementation of in `compiler`)
- "Java-ize" the interface between `interp` and `interp-api` / `repl-api` (no direct dependency to / from `interp`, instead `interp` creates instances of Java interfaces, that `interp-api` / `repl-api` pick at runtime later on)
- wire things so that e.g. Ammonite running in 2.13 can run 2.12 code (from the CLI, and in the tests)

It might be possible to add Scala 3 support with only the first step, even though I'd be in favor of keeping the others too (they would allow to stop cross-compiling most modules to `2.12` while retaining `2.12` support for some time, and would even allow later on to compile most Ammonite modules with Scala 3 only while retaining Scala 2 support).

So to help reviewing this, I could rebase things, and open a first PR only adding the `compiler` module, that should be the less controversial change here. And we can wait before deciding to Java-ize  things. I've been working on the core of Scala 3 support lately, and it should be possible to rebase it on an Ammonite code base without the "Java-ize" stuff.

alexarchambault:
Also, after the changes here, I feel the graph of modules in Ammonite becomes somehow messy… (`ops`, `util`, `runtime`, `compiler-interface`, `compiler`, `interp-api-full`, `interp-api`, `interp`, `repl-api-full`, `repl-api`, `repl`, `main`, …). I think it could be simplified, by merging some modules:
- `ops`, `interp-api-full`, `repl-api-full` into a single `api` module
- `runtime` merged into `interp`
- `repl` merged into `main`
- `interp-api` and `repl-api` into a single `api-interface` module.

So that we would end up with just:
- `util`: shared utility between `interp` and `compiler`
- `compiler`: code interfacing with scalac / dotc
- `compiler-interface`: interfaces implemented in `compiler`
- `api`: user-facing code
- `api-interface`: interfaces passed to `api` to interact with the interpreter
- `interp`: module allowing to create / embed a full-fledged interpreter
- `main`: the CLI app of Ammonnite, wrapping an interpreter from `interp` in a CLI application.

alexarchambault:
@lihaoyi If needed, I can be up for a call, if there are things you think would be more convenient to discuss this way.

alexarchambault:
FYI, I pushed the early Scala 3 support on [this branch](https://github.com/alexarchambault/Ammonite/tree/wip-scala-3). Its README gives more details about what works / what doesn't work yet, and how to run Ammonite with Scala 3 support from the branch sources.

lihaoyi:
Hey @alexarchambault! Sorry I just got to this.

I think at a high level, I want to preserve the `runtime`/`interp` and `repl`/`main` splits. I think they have meaningful semantics distinctions (e.g. `runtime` doesn't need compiler, `interp` does) and helps us keep things organized and e.g. avoid accidentally classloading the scala compiler on the cached-script code path.

Regarding your earlier question splitting up the PR, I think limiting the step to extracting a `compiler` module that differs between Scala2 and Scala3 would be my preference. AFAIK you should only need to Javaize the interface between `compiler` and the rest of the code, and wouldn't need to Javaize the other inter-module interfaces since those will remain on Scala2. This seems like it would also help avoid the sprawl of `-api` and `-api-full` modules.

In general, let's aim to get Scala 3 in with minimal churn, and then we can consider further cleanups like dropping cross-compilation across Scala 2 versions later

smarter:
I think there's some overlap between the Java compiler interface you're trying to define and the one that metals uses: https://github.com/scalameta/metals/tree/main/mtags-interfaces/src/main/java/scala/meta/pc, maybe there's an opportunity for defining a common standard here? In Dotty we also have https://github.com/lampepfl/dotty/tree/master/interfaces/src/dotty/tools/dotc/interfaces which is very minimal but could be extended.

alexarchambault:
> I think at a high level, I want to preserve the `runtime`/`interp` and `repl`/`main` splits. I think they have meaningful semantics distinctions (e.g. `runtime` doesn't need compiler, `interp` does) and helps us keep things organized and e.g. avoid accidentally classloading the scala compiler on the cached-script code path.

Right. About `runtime`/`interp` though, with the changes here, `interp` doesn't depend anymore on scala-compiler, only on a new pure Java `compiler-interface` module. So `runtime` makes less sense in that case I guess.

> Regarding your earlier question splitting up the PR, I think limiting the first step to extracting a compiler module that differs between Scala2 and Scala3 would be my preference.

Yeah, that's what I'm going to do. We may even not need the Java API for `compiler` either. Scala 2.13 and Scala 3 code should be able to work alongside each other, both at compile-time and runtime. My plan now would be to try to keep `interp`/`repl`/`main`/`interp-api`/`repl-api` in Scala 2.13, and only `compiler` in Scala 3, if that can work.

alexarchambault:
@smarter Thanks for those links! It might be possible to extend those for Ammonite… The interfaces from metals seem quite specific to it though (with many lsp4j classes involved). But some definitions from the dotty interfaces could have been used here (like `Diagnostic`).

That said, for now, the level at which we abstract things in this PR involves things quite specific to Ammonite, like [class whitelist stuff](https://github.com/alexarchambault/Ammonite/blob/3ade62ab870c60a751a4aaa2127abac1d94cbdc6/amm/compiler/interface/src/main/java/ammonite/compiler/iface/CompilerLifecycleManager.java#L12), the [dependency completer](https://github.com/alexarchambault/Ammonite/blob/3ade62ab870c60a751a4aaa2127abac1d94cbdc6/amm/compiler/interface/src/main/java/ammonite/compiler/iface/CompilerLifecycleManager.java#L11), or even the [handling of "`Frame`s"](https://github.com/alexarchambault/Ammonite/blob/3ade62ab870c60a751a4aaa2127abac1d94cbdc6/amm/compiler/interface/src/main/java/ammonite/compiler/iface/CompilerLifecycleManager.java#L10). But maybe with slightly lower level abstractions, we could get interfaces that could be useful in other contexts, and having them in dotty itself would be interesting then.

alexarchambault:
Just so you know, I'll probably open a new PR adding just the new compiler module during the first week of January I think.

alexarchambault:
Closing this. Going to open new PRs with less changes.

PR #1139, opened 2020-12-16
Update Scala to 2.13.4

alexarchambault:


ckipp01:
@alexarchambault thanks a ton for taking care of this. While updating this in metals https://github.com/scalameta/metals/pull/2331 it looks like the 2.13.4 is now there:

```
❯ cs complete com.lihaoyi:ammonite-repl-api_2.13.4:
2.3.8-3-05aed517
```

But the 2.12.12 version isn't

```
❯ cs complete com.lihaoyi:ammonite-repl-api_2.12.12:2.3.8-3-05aed517

```

I do see a failure in the logs, but also see that it says that it published. So I think the job may have actually failed to published: https://github.com/lihaoyi/Ammonite/runs/1570558987#step:5:4696

Is it possible to maybe get this re-ran?

alexarchambault:
> Is it possible to maybe get this re-ran?

We can't re-run individual jobs for now, as they're all in the same workflow, and I don't think the release jobs are idempotent (can be run several times safely). We should be able to re-run individual jobs with https://github.com/lihaoyi/Ammonite/pull/1140.

PR #1140, opened 2020-12-18
Run release jobs in separate workflows

alexarchambault:
So that they can be re-run independently of each other. For example, in [this run](https://github.com/lihaoyi/Ammonite/actions/runs/428176544), we could re-run just release2 that failed.


alexarchambault:
@lihaoyi You added the GitHub workflow recently. Are you ok with these changes?

lihaoyi (approved):
shipit

PR #1143, opened 2021-01-06
Update scala-collection-compat to 2.3.1

alexarchambault:
Only version also published for Scala `3.0.0-M2`.


alexarchambault:
Merging this, so that we can re-run the release jobs.

PR #1144, opened 2021-01-14
Move compiler-related stuff to a new compiler module

alexarchambault:
This subsumes https://github.com/lihaoyi/Ammonite/pull/1135. The core idea of this PR consists in moving any compiler- or parsing-related logic to a new `compiler` module, that we can rewrite for Scala 3 later on.

Thanks to the 2.13 / 3 compatibility, we can just rewrite this `compiler` module in Scala 3 in a first time, and run it alongside the other modules compiled for 2.13, to get Scala 3 support. This is done on [this branch](https://github.com/alexarchambault/Ammonite/tree/scala-3-with-dotty-compat) (I'm about to add details about what it can and cannot do in its README).

This PR also adds a new `compiler-interface` module, defining a few compiler- and parsing-related traits, and has `interp` and `repl` depend on those traits rather than on the classes of `compiler`. This ensures we don't inadvertently depend on concrete scalac or fastparse definitions or `compiler` internals from `interp` and `repl`, which would hamper substituting `compiler` targetting Scala 3 to `compiler` targetting Scala 2.13 at runtime.

In more detail, the commits in this PR successively:
- do minimal refactorings that help subsequent developments,
- add a new `compiler` module and move code to it,
- add a new `compiler-interface` module, with mostly just traits,
- implement the traits of `compiler-interface` in `compiler`,
- rely on `compiler-interface` rather than `compiler` in `interp`, then in `repl`.

Commit-by-commit:
- Minimal refactorings
  - https://github.com/lihaoyi/Ammonite/commit/8615e3f5009c7bdb8ff839f34848f538df100543, https://github.com/lihaoyi/Ammonite/commit/593cff2283fad47ae1066bea09cb5b28044953bf: unused / dead code
  - https://github.com/lihaoyi/Ammonite/commit/00b3664894c7711355442ff3cbcab755ddfc013a, https://github.com/lihaoyi/Ammonite/commit/d6c3530b20db610894dc7fdbb8fd184165744a3f, https://github.com/lihaoyi/Ammonite/commit/e742bc87aa719e8aedc32f1101a8446f771341d4, https://github.com/lihaoyi/Ammonite/commit/95243f32318a2946ae1f1aa9bfad4c001891e753: tweak `CompilerLifecycleManager` API
[...]

PR #1150, opened 2021-02-03
Add Scala 3 support via dotty compatibility

alexarchambault:
So this PR adds Scala 3 support, as a follow up of https://github.com/lihaoyi/Ammonite/pull/1144.

Its changes successively consist in:
- minor refactorings, easing subsequent developments (from https://github.com/lihaoyi/Ammonite/pull/1150/commits/0dd150b9c01fcba57d9350281dff8e6fe9849485 up to https://github.com/lihaoyi/Ammonite/pull/1150/commits/6208f809f4208f223a520aee8640e617329c7b5c, 12 commits as of now),
- actually adding Scala 3 support (https://github.com/lihaoyi/Ammonite/pull/1150/commits/8c22bfa162fabbc0088bf63a2e235be30f0b1664 up to https://github.com/lihaoyi/Ammonite/pull/1150/commits/6a815b0006b0b95a9a2921fe6ad6badfb5752ee3, 10 commits)
- updating the tests, so that they run fine for Scala 3 (https://github.com/lihaoyi/Ammonite/pull/1150/commits/a4e2db92404186158c27ca27cc9e1bf12f5b391c up to the end, 28 commits).

All commits but one don't touch more than approx. 100 lines each, which should make it easy to review those. The only sizable commit is the one adding the Scala 3 implementation of the `compiler` module, under `amm/compiler/src/main/scala-3` (https://github.com/lihaoyi/Ammonite/pull/1150/commits/8c22bfa162fabbc0088bf63a2e235be30f0b1664). This commit mainly:
- adds new sources under `amm/compiler/src/main/scala-3`,
- updates `build.sc` so that using just `3` as cross version builds and runs Ammonite as a mix of Scala 2.13 / 3 modules (that run fine together thanks to dotty compatibility)

Many of the sources this commit adds under `amm/compiler/src/main/scala-3` are adapted either from dotty sources, or from Ammonite Scala 2 things.

Some features don't work yet in Scala 3, or don't work as fine as in Scala 2 yet. The commits updating the tests should make clear what has been disabled for Scala 3. Most notably:
- entrypoints: this would require mainargs to be ported over to Scala 3, its use of Scala 2 macros complicates that,
- line numbers in diagnostics and stack traces: an equivalent of the `LineNumberModifier` plugin for Scala 3 would fix both at the same time, else `ammonite.interp.script.PositionOffsetConversion` could be used to fix at least the former
- deep completions: there's currently a candidate implementation for these (in `amm/compiler/src/main/scala-3/dotty/ammonite/compiler/Completion.scala`), but it's awfully slow, so it's disabled by default (look for the commented out line calling `addDeepCompletions`)
- import $ivy completion: there's a candidate implementation too (in `Completion.scala`), but it has issues with backticks. This required tinkering in Scala 2, which seemed less straightforward to do in Scala 3 (IIRC, dotty doesn't easily hand us over backticks around what's being completed)
- `source` and `desugar` macros: they need to be ported over to Scala 3 (using Tasty reflection-based macros I guess)

There are also limitations in the Scala 3 specific code, most notably:
- the script blocks splitter (splitting around `@`) is regex based for now (I didn't try to have the dotty parser handle those yet)


alexarchambault:
@lihaoyi This is ready for review, if you wish. Don't fear the number of commits or the number of changed lines… All commits but one touch less than or around 100 lines. The only sizeable commit is the one adding the Scala 3 source of the `compiler` module - it mostly only adds code, but for its changes in `build.sc`.

That said, I can always split that to make it easier to review it:
- all the commits until the one adding Scala 3 code in `compiler` in a first PR, then
- adding Scala 3 code per se, then
- adding CI jobs and adjusting the tests for Scala 3.

The tests now pass in https://github.com/lihaoyi/Ammonite/pull/1151, that switches to `3.0.0-M3`. Its developments could also be folded here.

Lastly, some things could be cleaned-up in the biggest commit (Scala 3 code in `compiler`), I could update that, but that shouldn't hamper review… (or even merging).

lihaoyi (approved):
@alexarchambault I skimmed this, I think it looks good to me. I trust both that you know your way around the codebase, and also our test suite to catch any regressions in the old code paths.

Since the bulk of the code here is in Scala 3, could it be worth getting someone more familiar with Scala3 to review those sections? Even if you need to explain what the code is meant to do, a second perspective from the Scala3 side of things may be valuable to make sure what we're doing sounds reasonable to them

alexarchambault:
(Last push force: folded https://github.com/lihaoyi/Ammonite/pull/1151 here.)

alexarchambault:
> Since the bulk of the code here is in Scala 3, could it be worth getting someone more familiar with Scala3 to review those sections? Even if you need to explain what the code is meant to do, a second perspective from the Scala3 side of things may be valuable to make sure what we're doing sounds reasonable to them

I agree it would be nice to have more Scala 3 expert eyes have a look, yes.

@smarter @anatoliykmetyuk Don't know if you guys would be interested in having a look at the commit named `Add Scala 3 stuff`, currently at https://github.com/lihaoyi/Ammonite/pull/1150/commits/ecf292847abb7c4753a84760cbe8e55b26f14bb8. This commit adds Scala 3 implementations for parsing and compiler related stuff.

In more detail, it adds the following classes:
- `AmmonitePhase`: a Scala 3 equivalent to [`AmmonitePlugin`](https://github.com/lihaoyi/Ammonite/blob/c4989f6c5d3bdb991541d81cbfe544e6bfcc7d0a/amm/compiler/src/main/scala/ammonite/compiler/AmmonitePlugin.scala), which tracks new variables, but also new imports, during compilation
- `Compiler`: as the name says, going from source to byte code, but also some completion-related things
- `DottyParser` and `Parsers`: parsing-related stuff
- `Preprocessor`; generating additional code (printing the type and value of new `val`s, etc.) from an AST of the input code (most of it adapted from its [Scala 2 counterpart](https://github.com/lihaoyi/Ammonite/blob/c4989f6c5d3bdb991541d81cbfe544e6bfcc7d0a/amm/compiler/src/main/scala/ammonite/compiler/DefaultPreprocessor.scala))
- `SyntaxHighlighting`: adapted from [dotty sources](https://github.com/lampepfl/dotty/blob/3.0.0-M3/compiler/src/dotty/tools/dotc/printing/SyntaxHighlighting.scala)
- `Completion`: adapted from [dotty sources](https://github.com/lampepfl/dotty/blob/3.0.0-M1/compiler/src/dotty/tools/dotc/interactive/Completion.scala) too, so that we can add support for import $ivy completions, deep completion (committed but disabled for now, as my current implem is too slow), but also for a better handling of back-quoted identifiers (that could be interesting for the dotty REPL too)

smarter:
Looks like there's quite a lot of code adapted from existing code in dotty, my advice would be to instead change dotty to make the existing code reusable for your purposes when possible.

alexarchambault:
> Looks like there's quite a lot of code adapted from existing code in dotty, my advice would be to instead change dotty to make the existing code reusable for your purposes when possible.

I agree. That's mostly syntax-highlighting and completion-related code.

mallman:
Hi @alexarchambault. I was just working with Spark SQL in Jupyter thanks to Almond, and I'm curious how you're handling the removal of `scala.Symbol` and its related ' literal syntax from Scala 3 in this patch. Have you provided some kind of substitute? Unfortunately I don't understand this patch well enough to answer this question for myself. FWIW, my inquiry is partly motivated by the extensive use of the symbol literal syntax in the documentation for Ammonite at ammonite.io. However, I must admit I'm also rather partial to that syntax for use in Spark SQL, so I'm hopeful you have found a pleasant alternative.

Cheers.

alexarchambault:
@mallman In the Ammonite codebase, most symbol literals (maybe even all of them, I think) are related to [os-lib](https://github.com/com-lihaoyi/os-lib) paths, in code like `os.pwd / 'foo / 'something`. Luckily, os-lib also accepts string literals, so I just changed such code to `os.pwd / "foo" / "something"`.

I'm not sure how spark-sql will handle the end of symbol literals, but IIRC, it also accepts string literals in most places, right? Using string literals would be the way to go then.

(By the way, this kind of question might be more suited to the [Gitter channel](https://gitter.im/lihaoyi/Ammonite), although I have to admit I should check my Gitter notifications more often 😬)

mallman:
FYI, motivated by a little nudging from the Spark community, symbol literals may make a comeback in Scala 3 after all.

https://github.com/lampepfl/dotty/pull/11588

alexarchambault:
For anyone wondering what's the status of this, bits of this PR (mostly code originally adapted from dotty) have been moved to dotty itself in https://github.com/lampepfl/dotty/pull/11794, which is part of Scala `3.0.0-RC2`. I can't update this PR to `3.0.0-RC2` right now though: it relies on dotty compatibility to have Scala 2.13 and 3.0 modules go along, and no 2.13.x release (or even nightly) can read 3.0.0-RC2 modules for now. Because tasty reader version was bumped in 3.0.0-RC2, and work is still ongoing to bump 2.13.x to the same tasty reader version.

As soon as a 2.13.x nightly can read 3.0.0-RC2 modules, I'll update this PR, and it should be good to merge.

alexarchambault:
Note that the limitations discussed in [this comment](https://contributors.scala-lang.org/t/porting-ammonite-to-dotty/4469/29?u=alexarchambault), and below it, still apply.

Support for new types of statements (extensions, …) need to be added (which hopefully should only be a matter of adding a few cases in the preprocessor around [here](https://github.com/alexarchambault/Ammonite/blob/abac990a0606b1415b2e4784b0f81c03a7f41f65/amm/compiler/src/main/scala-3/ammonite/compiler/Preprocessor.scala#L244)).

Line numbers in the stack traces are still wrong (we need an equivalent of the [`LineModifierPlugin`](https://github.com/com-lihaoyi/Ammonite/blob/1cce53f30d8208b87173c2976a5751ebdc1ba45b/amm/compiler/src/main/scala/ammonite/compiler/AmmonitePlugin.scala#L297) in Scala 3 - I initially thought just postprocessing the warning and error messages, as done in the [BSP support](https://github.com/com-lihaoyi/Ammonite/blob/0f0d597f04e62e86cbf76d3bd16deb6965331470/amm/interp/src/main/scala/ammonite/interp/script/PositionOffsetConversion.scala#L116), would be fine, but it can't work for line numbers in stack traces, so we still need a compiler plugin…)

Entrypoints in scripts don't work yet. [mainargs](https://github.com/com-lihaoyi/mainargs) needs to be ported to Scala 3.

alexarchambault:
It seems we now have nightlies with https://github.com/scala/scala/pull/9617 ([`4fa6da5`](https://scala-ci.typesafe.com/artifactory/scala-integration/org/scala-lang/scala-compiler/2.13.6-bin-4fa6da5/), [`cefcd53`](https://scala-ci.typesafe.com/artifactory/scala-integration/org/scala-lang/scala-compiler/2.13.6-bin-cefcd53/)), going to try to update this.

alexarchambault:
Now only waiting for https://github.com/com-lihaoyi/Ammonite/pull/1180, so that we can use `2.13.6` as the `3.0.0`-compatible 2.13 version rather than a 2.13 nightly, and so that we can scrap all the scala nightly version stuff.

PR #1159, opened 2021-03-11
Added Scala 2.13.5 to cross matrix

lefou:


lefou:
@alexarchambault This is ready for review/merge.

alexarchambault:
I wanted to jump in, my bad :| Adding the version here is not enough. It's not tested on the CI in particular.

https://github.com/com-lihaoyi/Ammonite/pull/1139 illustrates the kind of changes needed to bump the Scala version.

lefou:
Sorry for my impatience. Looks like there is still something to do, to release a version for 2.13.5. 

I find it quite hard to distinguish the equal labelled GH workflows. My take in other projects to tackle a large test matrix is to depend with the release job on the success of the tests, so that the actual release only happens if all is good, but still allow a re-run of the CI in case of issues. It's best it to not mirror the matrix defined in the `build.sc` in the CI, but just let mill do the work.

Anyway, I can try to provide the necessary changes required for a 2.13.5 release in another PR.

alexarchambault:
> My take in other projects to tackle a large test matrix is to depend with the release job on the success of the tests, so that the actual release only happens if all is good

I agree. For now, I think it's better to keep them separate (which incidentally runs test and release jobs in parallel), so that we can re-run individual release jobs if they fail. If they were all defined in the same `.yml` file, we wouldn't be able to re-run only the jobs that failed.

PR #1161, opened 2021-03-12
Use latest Scala 2.13 for build and use a test matrix in CI

lefou:
This is my attempt to finish the Scala 2.13.5 bump and also properly test it. See #1159 

I switched to a test matrix, to make it more obvious where to add/change the scala version in case of a bump. IMHO, we should add _all_ scala versions, to which we have a release for, to the test matrix.

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -7,7 +7,12 @@ on:
     branches:
       - master
 jobs:
-  unit-11-2_12_13:
+  test:
+    strategy:
+      fail-fast: false
+      matrix:
+        java-version: [8, 11]
+        scala-version: [2.12.13, 2.13.5]
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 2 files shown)


lefou:
@alexarchambault I have no idea what the test failures mean.

amm/interp/src/main/scala/ammonite/interp/script/AmmoniteBuildServer.scala
```diff
@@ -69,7 +69,7 @@ class AmmoniteBuildServer(
           Dependency.of(
             "org.scalameta",
             "semanticdb-scalac_" + scala.util.Properties.versionNumberString,
-            "4.4.6"
```
lefou, line 72: I think this version should come from a the `build.sc`, e.g. by generating a `Versions` object.

Pros:
* easier discoverable dependency
* runtime version in sync with compile time version
* enables potential automatic update via. scala-steward

alexarchambault: I agree... Feel free to fix that if you'd like.

lefou: Ok, I was going to create a `Versions` in `amm.interp`, but I also found `ammonite.Constants` in scope. Should I rather reuse this?

lefou: I will open a new PR for this, combined with some more `build.sc` cleanup.

lefou: FTR, here is the PR: #1163

alexarchambault (approved):

alexarchambault:
(Let's squash-and-merge as soon as the CI is green, to keep the git history clean.)

PR #1163, opened 2021-03-14
Consolidate dependencies

lefou:
Also fixes some deprecated API use in `build.sc` and use correct GitHub organization in pom metadata.

A new `ammonte.interp.script.Constants` file is generated to use the semanticdb version defined in `build.sc` at runtime.


alexarchambault (approved):
Thanks @lefou!

PR #1164, opened 2021-03-14
Show instructions per OS.

sake92:
Ammonite works fine on Windows.  
Even if it doesn't, we should encourage users at least to try it! :)

Maybe we should also mention https://get-coursier.io/docs/cli-installation ?  
With `cs setup` you get all the necessary tools for working with scala, including Ammonite.

### DIFF
--- a/readme/Ops.scalatex
+++ b/readme/Ops.scalatex
@@ -35,9 +35,6 @@
   @p
     That handles the common case for you: recursively deleting folders, not-failing if the file doesn't exist, etc.
 
-  @p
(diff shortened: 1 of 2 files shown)


lefou (commented):
@sake92 Looks ok to me, but I'm really not very deep into Ammonite to have a final word.

alexarchambault (approved):
Thanks @sake92!

alexarchambault:
I never looked too closely at the doc, I think it has other outdated sections now... But all these changes look legit, let's merge!

PR #1165, opened 2021-03-15
Fix github urls for upload and pom metadata

lefou:
This should fix the release process.

I think a review is essential here. The CI will not help much until we merge into master and see a release happen.

### DIFF
--- a/build.sc
+++ b/build.sc
@@ -7,7 +7,7 @@ import $ivy.`io.get-coursier::coursier-launcher:2.0.0-RC6-10`
 val ghOrg = "com-lihaoyi"
 val ghRepo = "Ammonite"
 val masterBranch = "master"
-val homePage = "http://ammonite.io"
+val homePage = "https://ammonite.io"
 
 val isMasterCommit =
   sys.env.get("GITHUB_REPOSITORY") == Some(s"${ghOrg}/${ghRepo}") &&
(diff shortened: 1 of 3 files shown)


build.sc
```diff
@@ -662,12 +667,12 @@ def publishExecutable() = {
...
     if (!unstable){
       requests.post(
-        "https://api.github.com/repos/lihaoyi/Ammonite/releases",
+        s"https://api.github.com/repos/${ghOrg}/${ghRepo}/releases",
         data = ujson.write(
           ujson.Obj(
             "tag_name" -> buildVersion,
             "name" -> buildVersion,
-            "body" -> s"http://www.lihaoyi.com/Ammonite/#$buildVersion"
+            "body" -> s"${homePage}/#${buildVersion}"
```
sake92, line 675: Will this have double slash? Because `homePage` has a trailing slash.  
I prefer to have base URLs without a trailing slash..

lefou: Yeah, makes sense. Me too.

build.sc
```diff
@@ -4,9 +4,14 @@ import $file.ci.upload
 
 import $ivy.`io.get-coursier::coursier-launcher:2.0.0-RC6-10`
 
+val ghOrg = "com-lihaoyi"
+val ghRepo = "Ammonite"
+val masterBranch = "master"
+val homePage = "http://ammonite.io/"
```
sake92: https? And without trailing slash please. 😄

sake92 (approved):
LGTM

alexarchambault (approved):
Let's see how it goes!

PR #1170, opened 2021-03-23
Lazy load images in docs

sake92:
Improves #751  
This uses `loading="lazy"` native attribute as described here: https://web.dev/browser-level-image-lazy-loading/


alexarchambault (approved):
Thanks!

PR #1172, opened 2021-03-24
Fix Scripts docs

sake92:
Fixes:
- replaced `Ivy artifacts` with `Ivy/Maven artifacts`, just so users don't get confused whether they can use Maven repos...
- removed `scopt` refs
- removed `predefShared`


alexarchambault (approved):
👍

PR #1173, opened 2021-03-24
Fix compare in docs

sake92:
Umm.. was failing with `sh: 1: cannot open /dev/tty: No such device or address`.   
GH Actions probably doesn't have a tty/console when runs these, obviously... 😄 

I found this hack: https://github.com/actions/runner/issues/241#issuecomment-745902718

### DIFF
--- a/.github/workflows/publishDocs.yml
+++ b/.github/workflows/publishDocs.yml
@@ -18,4 +18,5 @@ jobs:
       - uses: actions/setup-java@v1
         with:
           java-version: '8'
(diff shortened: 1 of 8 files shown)


alexarchambault (approved):
Thanks @sake92!

I'm not that familiar with the website-related stuff, but I guess that's all fine.

sake92:
Not sure it works 100% but we'll see..  
Shouldn't be worse than the current situation where snippets are totally blank. 😄

PR #1180, opened 2021-05-17
Add Scala 2.13.6 support

alexarchambault:


alexarchambault:
It seems we need https://github.com/scalameta/scalameta/pull/2344.

PR #1183, opened 2021-05-27
Use Process.isAlive() in Shellout

Flowdalic:


Flowdalic:
Friendly ping. It would be great if someone could at least approve the pending workflows for this PR, so that I can see if there is anything to fix. :)

alexarchambault (approved):
Thanks!

PR #1187, opened 2021-06-18
Fix scala version in bootstrap script

lilac:
Fix the issue https://github.com/com-lihaoyi/Ammonite/issues/1018


alexarchambault (approved):
Thanks!

Issue #1204, opened 2021-08-14
FlatMap (||) does not work on Map collection type


sharon-wang:
### Error Output
Error seen in `publishDocs` github actions job:
`Ammonite Repl 2.4.0-34-2586f978 (Scala 2.13.1 Java 1.8.0_302)`
```
[info] ====================EXECUTING====================
[info] List(/home/runner/work/Ammonite/Ammonite/out/amm/2.13.1/assembly/dest/out.jar, --color, true, --no-remote-logging, --no-home-predef, --predef, shell/src/main/resources/ammonite/shell/example-predef-bare.sc)
[info] // I'm only interested in OS-related properties, show them to me!
[info] sys.props || grep! "os|OS".r
[info] exit
[info] 
[info] Map(JAVA_OPTS -> -Xmx600m)
[error] Loading...
[error] cmd0.sc:1: value || is not a member of scala.sys.SystemProperties
[error] val res0 = sys.props || grep! "os|OS".r
[error]                      ^
[error] Compilation Failed
[error] Bye!
```

A bit more context running locally:
`Ammonite Repl 2.4.0 (Scala 3.0.0 Java 16.0.2)`
```
    |  value || is not a member of scala.sys.SystemProperties.
    |  An extension method was tried, but could not be fully constructed:
    |
    |      ammonite.ops.FilterMapExtImplicit[scala.sys.SystemProperties](
    |        ammonite.$sess.cmd17.res17
    |      )(
    |        scala.collection.generic.IsIterable.isSeqLikeIsIterable[Repr](
    |          scala.collection.generic.IsSeq.rangeIsSeq[Nothing]
    |        )
    |      )
```
Note: [`class SystemProperties extends AbstractMap[String, String]`](https://www.scala-lang.org/api/3.0.0/scala/sys/SystemProperties.html)

### Reproduction
As seen above, run `sys.props || grep! "os|OS".r` or something equivalent in the form `<MAP> || <COMMAND>`.

sharon-wang:
Oops, just realized this has already been brought up in https://github.com/com-lihaoyi/Ammonite/issues/1103. Closing.

PR #1205, opened 2021-08-14
Fix documentation rendering (`@compare`,` ammSample` bugs)

sharon-wang:
There were some bugs in `Sample.scala` which were truncating the script results, adding `$ bash` to every line of bash output and a typo (wrong directory name) in `Shell.scalatex` which caused a bash command to fail. These errors have made it challenging for those learning Ammonite via the documentation at https://ammonite.io/, as numerous inline examples were missing (not rendered).

This PR addresses the rendering problems for `@compare` and `ammSample`, which were reported in the following issues:
Fixes: https://github.com/com-lihaoyi/Ammonite/issues/1203, fixes: https://github.com/com-lihaoyi/Ammonite/issues/1102, fixes: https://github.com/com-lihaoyi/Ammonite/issues/1086, fixes: https://github.com/com-lihaoyi/Ammonite/issues/1132.

I am new to Ammonite and found it difficult to follow the documentation due to the rendering issues, so I hope this alleviates that problem for the issue openers and myself!

Additionally, I opened #1204, which I discovered while trying to fix the rendering issues. Looking through the `publishDocs` github action output will show other errors.

---

I went about my testing via the `publishDocs` github action, by printing out the raw html generated in `Sample.scala`. I wasn't sure how to build the docs locally, so this was my workaround 😅 

Here's the diff showing that some of the content was not getting rendered at all, and now we do have divs with the expected content: [html_diff.txt](https://github.com/com-lihaoyi/Ammonite/files/6985933/html_diff.txt) (rename to `html_diff.diff`)
And the html output that the diff was generated from: 
[current.txt](https://github.com/com-lihaoyi/Ammonite/files/6985934/current.txt) (rename to `current.html`), [fixes.txt](https://github.com/com-lihaoyi/Ammonite/files/6985935/fixes.txt) (rename to `fixes.html`).

### DIFF
--- a/readme/Sample.scala
+++ b/readme/Sample.scala
@@ -20,12 +20,12 @@ object Sample{
   val cygwinSed = """$ sed -i '0,/"\$0"/{s/"\$0"/`cygpath -w "\$0"`/}' /usr/local/bin/amm"""
   val filesystemCurl =
     "$ mkdir -p ~/.ammonite && curl -L -o ~/.ammonite/predef.sc https://git.io/vHaKQ"
-  
+
   val cacheVersion = 6
   def cached(key: Any)(calc: => String) = {
     val path = cwd/'target/'cache/(key.hashCode + cacheVersion).toString
     try read! path
-    catch{ case e =>
+    catch { case e : Throwable =>
       val newValue = calc
       write.over(path, newValue)
       newValue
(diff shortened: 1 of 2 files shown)


sharon-wang:
This also appears to be related: https://github.com/com-lihaoyi/Ammonite/pull/1114

sake92:
@sharon-wang welcome and thank you very much for doing this!  😃

PR #1209, opened 2021-08-23
Add option to enable ScalaPy in the REPL

kiendang:
Add `--scalapy` flag that enables ScalaPy in the REPL by setting up the correct system properties required by ScalaPy, pre-importing ScalaPy Ivy dependency and the `py` object. The `--python-executable` allows optionally specifying the path to a specific Python interpreter executable to be used with ScalaPy (defaults to `python3` if not provided).

```sh
amm --scalapy
```

or

```sh
amm --scalapy --python-executable /usr/local/bin/python3.8
```

```scala
@ py"'Hello!'"
res0: py.Dynamic = Hello!

@ py"[x for x in ${List(1, 2, 3).toPythonProxy}]"
res1: py.Dynamic = [1, 2, 3]
```

Specifying `--amm --python-executable <path>` is equivalent to running

```scala
@
import $ivy.`ai.kien::python-native-libs:0.2.1`
import $ivy.`me.shadaj::scalapy-core:0.5.0`

ai.kien.python.Python("<path>").scalapyProperties.get.foreach {
  case (k, v) => System.setProperty(k, v)
}

import me.shadaj.scalapy.py
import me.shadaj.scalapy.py.{PyQuote, SeqConverters}
```

This adds no dependency to ammonite (not even scalapy) other than [python-native-libs](https://github.com/kiendang/python-native-libs), a tiny helper for calculating the correct system properties for ScalaPy to load a certain Python interpreter.

@alexarchambault

### DIFF
--- a/amm/repl/src/main/scala/ammonite/main/Defaults.scala
+++ b/amm/repl/src/main/scala/ammonite/main/Defaults.scala
@@ -29,6 +29,17 @@ object Defaults{
       typeOf
     }""")
   )
+
+  val scalapyImports = Imports(
+    ImportData("""me.shadaj.scalapy.{
+      py
+    }"""),
+    ImportData("""me.shadaj.scalapy.py.{
+      PyQuote,
+      SeqConverters
+    }""")
+  )
+
   def ammoniteHome = os.Path(System.getProperty("user.home"))/".ammonite"
 
   def alreadyLoadedDependencies(
--- a/amm/src/main/scala/ammonite/Main.scala
+++ b/amm/src/main/scala/ammonite/Main.scala
@@ -7,6 +7,7 @@ import java.nio.file.NoSuchFileException
 import ammonite.compiler.{CodeClassWrapper, DefaultCodeWrapper}
 import ammonite.compiler.iface.{CodeWrapper, CompilerBuilder, Parser}
 import ammonite.interp.{Watchable, Interpreter, PredefInitialization}
+import ammonite.interp.api.IvyConstructor
 import ammonite.interp.script.AmmoniteBuildServer
 import ammonite.runtime.{Frame, Storage}
 import ammonite.main._
@@ -19,6 +20,9 @@ import ammonite.runtime.ImportHook
 import coursierapi.Dependency
 import scala.concurrent.Await
 import scala.concurrent.duration.Duration
+import scala.util.{Success, Failure}
+
+import ai.kien.python.Python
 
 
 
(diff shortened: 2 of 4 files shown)


lihaoyi:
This seems like something that should be implementable without needing to hardcode support in Ammonite. Ammonite does not and should not know about even popular Scala libraries, let alone niche ones like ScalaPY

alexarchambault:
On purely technical grounds, adding that support directly in Ammonite allows to start a ScalaPy-included session without the overhead of compiling a predef. `--py` directly adds dependencies and some imports and runs a bit of code, without compiling anything upfront. That's the main motivation for adding it here IMO.

But I think this could be worked around with more powerful predefs. For example, `--predef` could accept files ending in `.json`, looking like
```json
{
  "dependencies": [
    "me.shadaj::scalapy-core:X.Y.Z",
    "ai.kien::python-native-libs:0.2.1"
  ],
  "imports": [
    "me.shadaj.scalapy.py",
    "me.shadaj.scalapy.py.{PyQuote, SeqConverters}"
  ],
  "init": "ai.kien.python.Python.setProperties",
  "code": ""
}
```

All fields would be optional. Here `init` points to a static method that would be called by reflection (by-passing Scala compilation). That way, users would still benefit from fast startups if the `code` field is empty or absent.

alexarchambault:
On non-technical grounds, adding `--py` in Ammonite could have encouraged the use of both ScalaPy - hopefully making it less "niche" - and of Python libraries from Scala, which I think could help new Scala users having a Python background. But I can understand it might feel a bit early at this point… So the more powerful predefs proposed above could work too IMO.

lihaoyi:
In general, none of the things above are specific to Ammonite, nor is there a clear problem that this PR is addressing. I think we can safely close this

alexarchambault:
@lihaoyi Are you fine with the JSON predefs proposal above? It could be useful in many contexts, not only this one.

lihaoyi:
The json predefs seem like a neat idea, but nobody has actually asked for it, and you have to compile scripts anyway since we don't support precompilation. I think if we want to achieve that, we should work on enabling general-purpose precompilation instead rather than adding this niche feature that doesn't really solve the problem fully

alexarchambault:
> The json predefs seem like a neat idea, but nobody has actually asked for it

Nobody asked for it because nobody knew it was possible until https://github.com/com-lihaoyi/Ammonite/pull/1076, and even since, I guess we're the only ones to know.

I'm not sure what you mean by "general-purpose precompilation", but I know in other contexts that using the already warmed-up scalac instances of Bloop just mostly makes the need of caching compilation outputs unnecessary.

lihaoyi:
general purpose pre compilation is https://github.com/com-lihaoyi/Ammonite/issues/437

People ask for impossible things all the time, if it's something important to them. The above issue is an example of such an impossible ask. 

We could easily add JSON predefs; it's just a bunch of plumbing. But we shouldn't, unless we have a good reason, and I haven't seen one given yet.

Every feature adds to the complexity of the codebase and user experience, and a feature that isn't sufficiently useful could easily be negative value even if trivial to implement. Both ScalaPy and JSON predefs seem to me like they would fall under that category

PR #1211, opened 2021-08-24
fix ImportData.ImportType use of sealed

nbauernfeind:
The compiler says not all types are covered for this piece of code:
```scala
private[this] def onImportData(id: ImportData): Unit = {
  id.importType match {
    case ImportData.Type => onImportDataType(id)
    case ImportData.Term => onImportDataTerm(id)
    case ImportData.TermType =>
      onImportDataType(id)
      onImportDataTerm(id)
  }
}
```

I believe, that this is just simply from an incorrect usage of Scala's `sealed`.


nbauernfeind:
@alexarchambault I would be curious what you think of this PR as well =). Thank you in advance!

Issue #1148, opened 2021-01-26
Feature request: trap sys.exit calls


soronpo:
If I run `amm --watch` on a script that calls a function that does `sys.exit` ammonite exits and I need to rerun it to watch the files. 
It would be nice if ammonite can still return to watch the files in this case.

PR #1219, opened 2021-10-20
Issue #1148 trap exit when running script with --watch option

smthelusive:
When user runs the script with `--watch` option and if the script calls `sys.exit`, ammonite exits and doesn't stay in the watch loop. 
This PR adds a `TrapExitSecurityManager` that would be used when the `--watch` option is activated. When there is a `sys.exit` in the scrips, the special `TrapExitException` will be thrown giving a helpful output for the user:
Screenshot 2021-10-20 at 14 45 53

Please let me know what you think 🙂

### DIFF
--- a/amm/src/main/scala/ammonite/main/TrapExitSecurityManager.scala
+++ b/amm/src/main/scala/ammonite/main/TrapExitSecurityManager.scala
@@ -1,0 +1,18 @@
+package ammonite.main
+import java.security.Permission
+
+/**
+ * This security manager is assigned as default one when we run user scripts with the --watch option
+ * This allows to trap sys.exit by throwing a special TrapExitException
+ */
+object TrapExitSecurityManager extends SecurityManager {
+
+  override def checkExit(status: Int): Unit = throw new TrapExitException(status)
+
+  override def checkPermission(perm: Permission): Unit = {}
+
+  private class TrapExitException(status: Int) extends RuntimeException {
+    override def toString: String = s"script exited with status $status"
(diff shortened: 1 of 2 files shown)


amm/src/main/scala/ammonite/main/TrapExitSecurityManager.scala
```diff
@@ -0,0 +1,18 @@
+package ammonite.main
+import java.security.Permission
+
+/**
+ * This security manager is assigned as default one when we run user scripts with the --watch option
+ * This allows to trap sys.exit by throwing a special TrapExitException
+ */
+object TrapExitSecurityManager extends SecurityManager {
```
lolgab, line 8: Can this object be `private [ammonite]`?

lolgab:
sbt [recently dropped](https://github.com/sbt/sbt/commit/0413727796adebdb25d9fdf70ea038723f354ee6) the TrapExit feature since the SecurityManager was deprecated in Java 17. Do we want to add it now to Mill?

alexarchambault (approved):
Merging, thanks @smthelusive! (and @lolgab for his review)

amm/src/main/scala/ammonite/main/TrapExitSecurityManager.scala
alexarchambault: AFAIK, little effort has been made to clean up the public API of this module, which is not really meant to be used as a library anyway. So it shouldn't be a problem to have this be public.

PR #1225, opened 2021-11-19
Use more Scala 3 dependencies and bump Mill

lolgab:


lolgab:
I noticed that since the modules have `support3 = false`, setting the dependencies is not enough to use their Scala 3 counterparts (sadly).
Still not having to declare them with `withDottyCompat` seems an improvement.
I'm checking what's needed to make more modules use `supports3 = true`

alexarchambault (approved):
Thanks @lolgab!

PR #1226, opened 2021-11-19
Make runtime and repl support3

lolgab:
Depends on #1225 

This allows to use upickle macros in the repl and has a test to check that this is true.


lolgab:
Superseded by #1383

PR #1227, opened 2021-11-26
EOL Ammonite-Ops and Ammonite-Shell

lihaoyi:
These two modules were an interesting experiment almost a decade ago, but overall usage has not panned out. Ammonite-Ops is largely replaced by https://github.com/com-lihaoyi/os-lib, even if it's not an exact match. Ammonite-Shell never really saw widespread usage: the vast majority of Ammonite users use Ammonite as a Scala REPL and script runner in addition to (rather than instead of) their default Bash/Zsh/etc. shells.

This PR deletes everything related to these modules: their implementation, build config, and docs. This should help shrink the codebase, simplify the codebase, and let us focus more clearly on what's important going forward


alexarchambault:
It would be great to remove those 👍.

I didn't review in detail yet, but, "the CI errors seem legit" (although the ones in repl tests don't make a lot of sense at first…).

lihaoyi:
@alexarchambault I got all tests passing on Scala 2.x, but somehow there are Scala 3 failures I don't understand. Do you think you could take a look see if you can make sense of it?

alexarchambault:
I'll try to have a look. IIRC I already ran into such errors, and fixed them by ensuring pprint_2.13 isn't in the class path with Scala 3 (and no pprint_3 in Scala 2.13 either).

lihaoyi:
@alexarchambault looks like I managed to shake the build a bit and get it to compile and pass tests. I don't actually understand what the stuff in the build.sc is doing, but I guess if it passes tests we're probably good?

build.sc
```diff
@@ -316,19 +305,24 @@ class TerminalModule(val crossScalaVersion: String) extends AmmModule{
...
+        if (scala3Versions.contains(crossScalaVersion))
+          Agg(
+            ivy"com.lihaoyi:pprint_3:${Deps.pprint.dep.version}",
+            ivy"org.scala-lang.modules::scala-collection-compat:${Deps.scalaCollectionCompat.dep.version}",
+            ivy"org.scala-lang:scala3-library_3:$crossScalaVersion"
+          )
+        else Agg(
+          Deps.pprint
         )
+      )
```
lihaoyi, line 325: @alexarchambault In particular i have no idea what this part is doing haha

alexarchambault: IIRC, `crossScalaVersion` should be like `3.x`, but as `supports3` is false (the default), we compile this module with Scala 2.13. But `crossScalaVersion` like `3.x` implies it's meant to be consumed by Scala 3 modules nonetheless, so we force some dependencies to have a `_3` suffix (and also pull `scala3-library`, that Mill-built libraries don't pull automatically). It's required for dependencies having macros, like pprint (as Scala 2 and 3 macros are different, and pprint doesn't rely on the [Scala 2 & 3 macro mixing stuff](https://docs.scala-lang.org/scala3/guides/migration/tutorial-macro-mixing.html)).

alexarchambault: The `ivy"org.scala-lang.modules::scala-collection-compat:…"` seems superfluous here, as it doesn't force any suffix, and there's already `Deps.scalaCollectionCompat` a few lines before.

alexarchambault (approved):
LGTM I guess!

alexarchambault:
Seems the (somewhat annoying, but somewhat useful too :sweat_smile:) `noLongLines` test went away with this...

PR #1229, opened 2021-11-29
Split up Main.scala into Main.scala/MainRunner.scala/AmmoniteMain.scala

lihaoyi:
This should help make things easier to find.

The dependency order is `AmmoniteMain.scala` depends on `MainRunner.scala` depends on `Main.scala`

The new main method is in `ammonite.AmmoniteMain`. This was chosen to help improve the UX of tools like `jps` which only list the classname and not the package path as part of its output. Previously it would just show a mysterious `Main` JVM process running, with this change it'll show `AmmoniteMain` which is a lot more useful.

A few minor helper functions needed to be moved around to preserve acyclicity of the three files, I moved them all to `ammonite.util`


alexarchambault (approved):

amm/src/main/scala/ammonite/Main.scala
```diff
@@ -266,256 +266,3 @@ case class Main(predefCode: String = "",
   }
 }
 
-object Main{
-
-  /**
-    * The command-line entry point, which does all the argument parsing before
-    * delegating to [[Main.run]]
-    */
-  def main(args0: Array[String]): Unit = {
```
lefou, line 275: Is it possible, to keep a `@deprecated` forwarder here?

PR #1231, opened 2021-12-02
Fix handling of imports involving synthetic case class objects

### LINKED ISSUE #474: implicit def should override all previous definitions
```
@ import $ivy.`com.chuusai::shapeless:2.3.2` 
import $ivy.$                             
@ import shapeless._ 
import shapeless._
@ implicit def :: = 1  
defined function ::
@ implicitly[Int]  
cmd5.sc:1: could not find implicit value for parameter e: Int
val res5 = implicitly[Int] 
                     ^
Compilation Failed
```

I expect  `::` defined in REPL should be taken over the imported `shapeless.::`, and the `implicitly[Int]` should work.

lihaoyi:
Fixes #474

Turns out that even though the companion object for a `case class Foo` is marked as synthetic, it can still be imported and used. Thus when determining imports, we should not filter out synthetic symbols

### DIFF
--- a/amm/repl/src/test/scala/ammonite/session/ImportTests.scala
+++ b/amm/repl/src/test/scala/ammonite/session/ImportTests.scala
@@ -330,5 +330,25 @@ object ImportTests extends TestSuite{
         res3: Int = 1
       """)
     }
+    test("shapelessBugMinimized"){
+      check.session("""
+        @ object f{ case class Foo()}
+
+        @ import f.Foo
+
+        @ val Foo = 1
+
+        @ Foo
+      """)
+    }
+    test("shapelessBugFull"){
+      check.session("""
+        @ import $ivy.`com.chuusai::shapeless:2.3.2`, shapeless.::
+
+        @ implicit def :: = 1
+
+        @ ::
+      """)
+    }
   }
 }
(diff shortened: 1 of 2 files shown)


amm/repl/src/test/scala/ammonite/session/ImportTests.scala
```diff
@@ -330,5 +330,25 @@ object ImportTests extends TestSuite{
...
+        @ import f.Foo
+
+        @ val Foo = 1
+
+        @ Foo
+      """)
+    }
+    test("shapelessBugFull"){
+      check.session("""
+        @ import $ivy.`com.chuusai::shapeless:2.3.2`, shapeless.::
```
alexarchambault, line 346: For this to work with Scala 2.13:
```suggestion
        @ import $ivy.`com.chuusai::shapeless:2.3.7`, shapeless.::
```

amm/repl/src/test/scala/ammonite/session/ImportTests.scala
```diff
@@ -330,5 +330,25 @@ object ImportTests extends TestSuite{
...
+        @ object f{ case class Foo()}
+
+        @ import f.Foo
+
+        @ val Foo = 1
+
+        @ Foo
+      """)
+    }
+    test("shapelessBugFull"){
```
alexarchambault, line 344: I guess this test should be guarded by a `if (check.scala2)`, as shapeless basically [isn't the same project](https://github.com/typelevel/shapeless-3) in Scala 3.

alexarchambault (commented):
Apart from what's mentioned in the comments, LGTM!

lihaoyi:
@alexarchambault thanks for the pointers, test suite is green now

PR #1232, opened 2021-12-03
Allow users to re-run watched scripts by pressing enter

lihaoyi:
Now a user can do two things during `--watch`: either `Ctrl-C` to exit, or `Enter` to re-run. This is the first step in supporting the same functionality in Mill https://github.com/com-lihaoyi/mill/issues/490.

I also incorporated some of the upstream improvements in the Mill implementation of `def watchAndWait` https://github.com/com-lihaoyi/mill/blob/8c2ee3801f6471d4bf526f638c3668e4337c0fa6/main/src/mill/main/MainRunner.scala#L51-L67, which we should be able to largely replace with the original except for an override wrapping the original in the two `setIdle` calls.

The nested functions were getting messy, so moved some of them out into separate helpers

Tested manually by running `./mill -i amm[2.12.10].run -w foo.sc x 1 true` on a dummy script, verified that pressing `Enter` causes it to re-run while pressing other things doesn't

Review by @lefou @alexarchambault

### DIFF
--- a/amm/src/main/scala/ammonite/MainRunner.scala
+++ b/amm/src/main/scala/ammonite/MainRunner.scala
@@ -70,12 +70,19 @@ class MainRunner(cliConfig: Config,
   def runRepl(): Unit = watchLoop(isRepl = true, printing = false, _.run())
 
   def watchAndWait(watched: Seq[(Watchable, Long)]) = {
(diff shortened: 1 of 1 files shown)


lefou (commented):
Great. Looks good to me!

PR #1233, opened 2021-12-04
Use full file path for parse error messages and compiler metadata

lihaoyi:
Fixes https://github.com/com-lihaoyi/Ammonite/issues/646
Fixes https://github.com/com-lihaoyi/Ammonite/issues/493

Previously we constructed a `new io.VirtualFile` with an identical name and path, both short, to ensure stack traces included only the file name. The correct thing to do is to provide separate names and paths, so stack traces can use the file name while compiler errors can use the full file path. This is done differently in Scala2 and Scala3 due to differences in the compiler APIs, but the end result is the same

This makes the parse/compile errors line up more with how the normal Scala compiler reports them, and fixes `sourcecode.File` to properly report the full path of the script file.

Updated the existing LineNumberTests to verify the fixed behavior, and added a new `sourceCodeMetadata.sc` test case to verify the fix applies to the `sourcecode.*` macros. For now `sourcecode.*` macros do not work in the REPL or scripts in Scala3, but that is an unrelated problem we can deal with later

Review by @alexarchambault  and @anatoliykmetyuk

### DIFF
--- a/amm/compiler/src/main/scala-2/ammonite/compiler/Compiler.scala
+++ b/amm/compiler/src/main/scala-2/ammonite/compiler/Compiler.scala
@@ -70,22 +70,11 @@ object Compiler{
 
   /**
     * Converts a bunch of bytes into Scalac's weird VirtualFile class
-    *
-    * Can only take in a relative path for the VirtualFile! Otherwise
-    * fails with a weird assert in the presentation compiler:
-    *
-    * [Current.sc]: exception during background compile:
-    * java.lang.AssertionError:
-    * assertion failed: (
-    *   ammonite/predef//Users/lihaoyi/Dropbox/Github/<-wrapped to next line->
-    *   Ammonite/target/tempAmmoniteHome/predef.sc,false,16,16)
-    *
-    * It appears that passing in the filename *only* results in the correct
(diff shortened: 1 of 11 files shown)


lihaoyi:
@alexarchambault looks like the only things failing are some Scala3 tests. I think I need to modify this Scala3 code:

- https://github.com/com-lihaoyi/Ammonite/blob/master/amm/compiler/src/main/scala-3/ammonite/compiler/Compiler.scala#L210


The same way as the corresponding Scala2 code:

- https://github.com/com-lihaoyi/Ammonite/pull/1233/files#diff-3ab64d7cc381aecb54556f006a9a4ae207c16b14fd565a6e1b51ece614e3046eL88-L89

I'm not familiar with the Scala3 APIs, do you have any idea how we might accomplish that?

lihaoyi:
@alexarchambault Also, it seems the `sourcecode.*.apply()` functions don't work? Any idea why that might be the case?

lihaoyi:
@alexarchambault have you had a chance to look at this? If you're busy I'll probably just disable the respective Scala3 tests for now and move forward with the fix for Scala2

lihaoyi:
The sourcecode problem appears to be due to having the scala2 version of sourcecode on the scala3 ammonite test classpath

PR #1236, opened 2021-12-12
Release 2.5.0

lihaoyi:


lolgab (approved):

alexarchambault (commented):
The CI errors look legit (we might want to drop 2.12.0 - 2.12.7 support now, though).

lihaoyi:
Yeah I see that. Given that 2.12 is now on 2.12.15, and upgrading between point releases is meant to be "easy", I think dropping 2.12.0-2.12.7 sounds good to me

lefou (approved):

PR #1244, opened 2022-01-22
fix amm-template.sh: AMM_VERSION gets ignored in AMM_EXEC_PATH

funemy:
The `AMM_VERSION` gets ignored in `AMM_EXEC_PATH`

Changing Amm version wouldn't work if you already had a different Amm version installed for the same Scala version.


alexarchambault (approved):

alexarchambault:
(Pushed a merge commit to trigger CI.)

alexarchambault:
Thanks for this!

PR #1256, opened 2022-02-27
Linux curl is scala 2.13

ykmiu:
Was figuring out a local issue and noticed that this suggests the main Linux install is scala 2.12 when in fact it's a 2.13.

2.12 is referred to in the older scala versions section,


alexarchambault (approved):
Thanks!

PR #1264, opened 2022-04-13
Add back positions in Scala 3 diagnostics

alexarchambault:
I didn't take the time to update positions in diagnostics returned by dotty, when adding Scala 3 support, so positions were stripped in Scala 3 diagnostics. This PR fixes that.

Before:
```text
   |  n azzd 3
   |  ^^^^^^
   |value azzd is not a member of Int, but could be made available as an extension method.
   |
   |The following import might make progress towards fixing the problem:
   |
   |  import sourcecode.Text.generate
   |
Compilation Failed
```

After:
```text
-- [E008] Not Found Error: cmd0.sc:3:4 -----------------------------------------
3 |  n azzd 3
  |  ^^^^^^
  |value azzd is not a member of Int, but could be made available as an extension method.
  |
  |The following import might make progress towards fixing the problem:
  |
  |  import sourcecode.Text.generate
  |
Compilation Failed
```

### DIFF
--- a/amm/compiler/src/main/scala-3/ammonite/compiler/Compiler.scala
+++ b/amm/compiler/src/main/scala-3/ammonite/compiler/Compiler.scala
@@ -13,7 +13,7 @@ import ammonite.compiler.iface.{
   _
 }
 import ammonite.compiler.internal.CompilerHelper
-import ammonite.util.{ImportData, Imports, Printer}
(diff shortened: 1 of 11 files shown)


alexarchambault:
(I think the line numbers are still wrong for scripts with multiple `@` sections though.)

PR #1268, opened 2022-04-21
Add script argument control flags

### LINKED ISSUE #1267: Script argument handling can be unintuitive
Consider `MyScript.sc`:
```scala
def doTheThing(
  @arg(short='p') parallelism: Int = 1,
  @arg dry: Flag,
  @arg ids: Leftover[Int]
) = {
  println(s"parallelism: $parallelism")
  println(s"dry: ${dry.value}")
  println(s"ids: ${ids.value.mkString("[", ", ", "]")}")
}
```

Currently I can invoke this like:
```
$ amm MyScript.sc 4 5 6 7
parallelism: 4
dry: true
ids: [6, 7]
```
I think it's clear how the arguments are being misconstrued. What I'd like to have is:
```
$ amm MyScript.sc 4 5 6 7
parallelism: 1
dry: false
ids: [4, 5, 6, 7]
```

As far as I can tell, this is because the main method is invoked with `allowPositional = true` (`Scripts.scala:109`).

I believe forwarding `allowPositional` to `ammonite.Main` should allow this to be configured at least in an embedded environment. Adding a command line argument shouldn't be that much more effort.

If I'm mistaken, this may end up being an improvement for `mainargs` instead.

abernardi597:
Closes #1267

### DIFF
--- a/amm/src/main/scala/ammonite/AmmoniteMain.scala
+++ b/amm/src/main/scala/ammonite/AmmoniteMain.scala
@@ -109,7 +109,12 @@ object AmmoniteMain{
               false
 
             case (None, head :: rest) =>
-              val success = runner.runScript(os.Path(head, os.pwd), rest)
+              val success = runner.runScript(
+                os.Path(head, os.pwd),
+                allowPositional = !cliConfig.scripts.noPositionalArgs.value,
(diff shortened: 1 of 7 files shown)


alexarchambault:
I never really messed with mainargs stuff. I'll let @lihaoyi (or anyone else?) review this.

PR #1278, opened 2022-07-27
bugfix: Set proper Scala version in ScalaBuildTarget data

tgodzik:
Previously, we would use the Scala version from the Scala library, which will be 2.13.8 even for Scala 3. Now we take the same version as the compiler.


alexarchambault (approved):
Thanks, good catch!

PR #1279, opened 2022-08-12
Fix empty printer hygiene

alexarchambault:
Fixes https://github.com/com-lihaoyi/Ammonite/issues/1277.


alexarchambault:
This needs an integration test with a `scala.sc` file, like in https://github.com/com-lihaoyi/Ammonite/issues/1277.

lefou (approved):
This looks good to me. Regardless of the existence of an test, this is best practice for Scala 2.x macro writing. A test would be nice though.

PR #1284, opened 2022-08-25
Fix source and browse utilities to be able to use out of the box

tanishiking:
fix https://github.com/com-lihaoyi/Ammonite/issues/1254

This PR enables Ammonite-REPL to use `browse` and `source` built-in utilities out of the box.

ammonite-browse

- `browse`
  - As @ches mentioned in https://github.com/com-lihaoyi/Ammonite/issues/1254, `browse` takes implicit `pprint.PPrinter` that is not in the REPL scope by default.
  - fixed by giving a default argument https://github.com/com-lihaoyi/Ammonite/commit/05e295855b79a2cd3e861a5b7c81553caa30678a
- `source`
  - define `browseSourceCommand` that is called from macro (looks like forgotten to port in https://github.com/com-lihaoyi/Ammonite/commit/ba57f10ab8ca1c5b8c1e2041c9c338d44d1cb885)
  - Added `repl.codeColorsImplicit` to `predefImport`, (is this ok?)
  - Add `pprint.PPrinter` as default parameters (in a wired way)
    - default argument doesn't work for macro definition
    - import implicit `pprint.PPrinter` to REPL scope may change the behavior of other commands (like `grep`, `browse`) because they also receive implicit `pprint.PPrinter` (but it would be better to use different printer instance?)

### DIFF
--- a/amm/compiler/src/main/scala-2/ammonite/compiler/tools/SourceRuntime.scala
+++ b/amm/compiler/src/main/scala-2/ammonite/compiler/tools/SourceRuntime.scala
@@ -12,6 +12,10 @@ import scala.language.experimental.macros
 
 object SourceRuntime{
 
+  def defaultPPrinter = pprint.PPrinter.Color.copy(defaultHeight = Int.MaxValue)
+
+  def browseSourceCommand(targetLine: Int) = Seq("less", "+" + targetLine,"-RMN")
+
   /**
     * Pull the height from the pretty-printer as a heuristic to shift the
     * desired line towards the middle of the screen. Typically, the
--- a/amm/compiler/src/main/scala-2/ammonite/compiler/tools/source.scala
+++ b/amm/compiler/src/main/scala-2/ammonite/compiler/tools/source.scala
@@ -40,29 +40,49 @@ object source{
   }
 
   def apply(f: => Any)
-           (implicit pprinter: pprint.PPrinter,
-            colors: CodeColors): Unit = macro applyMacro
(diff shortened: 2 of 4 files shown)


alexarchambault (approved):
Nice!

I wonder if we don't miss some implicit PPrint somewhere… These used to work fine IIRC. I might have messed with the implicit PPrinter instance that should have been implicitly available in the session. But that can always be fixed later I guess.

scalway:
None of this version works in my system:

```
$ scala-cli repl --ammonite-version 2.4.0 --ammonite --scala 2.13.5
...
$ scala-cli repl --ammonite-version 2.5.1 --ammonite --scala 2.13.7
...
$ scala-cli repl --ammonite-version 2.5.4 --ammonite --scala 2.13.8
Welcome to the Ammonite Repl 2.5.4 (Scala 2.13.8 Java 11.0.16)
@ source(scala.collection.mutable.Buffer) 
cmd0.sc:1: could not find implicit value for parameter pprinter: pprint.PPrinter
val res0 = source(scala.collection.mutable.Buffer)
                 ^
Compilation Failed
```

This works like expected:
```
~$ scala-cli repl --ammonite-version 2.3.8 --ammonite --scala 2.13.5
Welcome to the Ammonite Repl 2.3.8 (Scala 2.13.3 Java 11.0.16)
@ source(scala.collection.mutable.Buffer)
//shows it as expected
```

tanishiking:
This change isn't included in 2.5.4, will be available in the next release (or the latest build as you tested :+1:)

PR #1285, opened 2022-09-04
Autocomplete (some) ivy Imports when there is no Closing Backtick

LaurenceWarne:
Hi, this PR is a (partial, and Scala 2 only - there appears to be other issues with completions for Scala 3?) fix for https://github.com/com-lihaoyi/Ammonite/issues/965:

https://user-images.githubusercontent.com/17688577/188308492-22450d23-2e45-4c18-9337-dcc2244d5995.mp4

Essentially, if it turns out there are no completion candidates for a snippet, and the the snippet starts with 
```
$ivy.`
```
We just retry by appending a backtick if the string only contains only one backtick... more of a workaround than an in depth fix  described in https://github.com/com-lihaoyi/Ammonite/issues/965#issuecomment-495857542 (there are many valid cases this wouldn't touch), but I hope it would cover the most common ways people would use `import $ivy`, the case described in the issue for example.

Let me know what you think anyway, thanks :slightly_smiling_face:

### DIFF
--- a/amm/compiler/src/main/scala-2/ammonite/compiler/Pressy.scala
+++ b/amm/compiler/src/main/scala-2/ammonite/compiler/Pressy.scala
@@ -335,7 +335,13 @@ object Pressy {
 
       val signatures = all.collect { case (name, Some(defn)) => defn }.sorted.distinct
 
-      (i - prefix.length, allNames, signatures)
+      val isPartialIvyImport = allNames.isEmpty &&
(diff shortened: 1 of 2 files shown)


lolgab (commented):
Very nice! :)

alexarchambault (approved):
Thanks a lot @LaurenceWarne!

LaurenceWarne:
Awesome, thanks :slightly_smiling_face:

PR #1288, opened 2022-09-07
Remove pre Scala 3.0 final hacks, future proof some Scala 3 stuff

alexarchambault:
This is helpful for the upcoming `3.2.0` support.


alexarchambault:
(CI is actually green.)

PR #1298, opened 2022-10-07
add @main entrypoint support for Scala 3

mpollmeier:
mainargs 0.2.3 [added support for scala3](https://github.com/com-lihaoyi/mainargs/pull/18)

To test (manually):

test.sc:
```scala
@main def main(name: String) = {
  println(s"Hello, $name!")
}
```

```bash
mill -i "amm[3.2.0].assembly" && out/amm/3.2.0/assembly.dest/out.jar test.sc --name Michael
mill -i "amm[2.13.9].assembly" && out/amm/2.13.9/assembly.dest/out.jar test.sc --name Michael
```

### DIFF
--- a/amm/src/test/scala/ammonite/main/MainTests.scala
+++ b/amm/src/test/scala/ammonite/main/MainTests.scala
@@ -292,11 +292,5 @@ class MainTests extends TestSuite{
 }
 
 object MainTests extends MainTests {
-  val isScala2 = ammonite.compiler.CompilerBuilder.scalaVersion.startsWith("2.")
-  override def tests =
-    if (isScala2) super.tests
-    else
-      Tests {
-        test("Disabled in Scala 3") - "Disabled in Scala 3"
-      }
+  super.tests
 }
--- a/amm/src/test/scala/ammonite/main/MainTests.scala
+++ b/amm/src/test/scala/ammonite/main/MainTests.scala
@@ -9,7 +9,7 @@ import utest._
   * Tests around Ammonite's CLI handling of main methods, argument parsing,
   * and the associated error behavior if the caller messes up.
  */
-class MainTests extends TestSuite{
+object MainTests extends TestSuite{
   def exec(p: String, args: String*) =
     new InProcessMainMethodRunner(InProcessMainMethodRunner.base / 'mains / p, Nil, args)
 
@@ -289,8 +289,4 @@ class MainTests extends TestSuite{
       }
     }
   }
-}
-
-object MainTests extends MainTests {
-  super.tests
 }
--- a/amm/src/test/scala/ammonite/main/MainTests.scala
+++ b/amm/src/test/scala/ammonite/main/MainTests.scala
@@ -97,9 +97,15 @@ object MainTests extends TestSuite{
           )
           assert(out.contains(expected.trim))
         }
+
         test("emptyArg"){
-          val evaled = exec("ArgList.sc", "")
-          assert(evaled.success)
+          val isScala2 = ammonite.compiler.CompilerBuilder.scalaVersion.startsWith("2.")
+          if (isScala2) {
+            val evaled = exec("ArgList.sc", "")
+            assert(evaled.success)
+          } else {
+            "Disabled in Scala 3"
+          }
         }
       }
     }
--- a/build.sc
+++ b/build.sc
@@ -185,6 +185,9 @@ trait AmmInternalModule extends CrossSbtModule with Bloop.Module{
       else Agg[Dep]()
     )
   }
+
+  // We compile with mainargs_2.13, but for the Scala 3 target we actually ship mainargs_3,
+  // because it contains Scala 3 macros which are required at runtime to find the main class entrypoint. 
   def runIvyDeps =
     if (scala3Versions.contains(crossScalaVersion))
       Agg(ivy"com.lihaoyi:mainargs_3:${Deps.mainargs.dep.version}")
--- a/build.sc
+++ b/build.sc
@@ -689,6 +689,12 @@ class MainModule(val crossScalaVersion: String)
       Deps.scalaJava8Compat
     )
 
+    def runIvyDeps =
+      if (scala3Versions.contains(crossScalaVersion))
+        Agg(ivy"com.lihaoyi:mainargs_3:${Deps.mainargs.dep.version}")
+      else
+        Agg(Deps.mainargs)
+
 
     def thinWhitelist = T{
       generateApiWhitelist(
@@ -715,7 +721,6 @@ class MainModule(val crossScalaVersion: String)
       amm.repl().sources() ++
       sources() ++
       externalSources()
-
   }
 }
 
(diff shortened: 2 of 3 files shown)


build.sc
```diff
@@ -180,9 +180,17 @@ trait AmmInternalModule extends CrossSbtModule with Bloop.Module{
...
+    Agg(Deps.mainargs) ++ (
+      if (isScala2()) Agg(Deps.acyclic)
+      else Agg[Dep]()
+    )
   }
+  def runIvyDeps =
+    if (scala3Versions.contains(crossScalaVersion))
+      Agg(ivy"com.lihaoyi:mainargs_3:${Deps.mainargs.dep.version}")
+    else
+      Agg(Deps.mainargs)
```
lolgab, line 192: Interesting approach!
So you compile with the `_2.13` artifact and then at runtime you change it with the `_3` so Scala 3 macros are available for users to compile their code.
If this works here, we could use the same approach to fix upickle which currently works in Scala 3 only in `--thin` mode + `$ivy` import but not normally since we have the `_2.13` artifact at runtime.

lolgab: I think you might wat to add a comment before `runIvyDeps` to explain what you are doing. It is not trivial to understand and people might get confused in the future.

lolgab:
Can you add/reuse a test for this functionality to run on Scala 3?

build.sc
mpollmeier: > So you compile with the _2.13 artifact and then at runtime you change it with the _3 so Scala 3 macros are available for users to compile their code.

Yes exactly - and yes you're right that totally deserves a comment - added just now. 
As far as I'm concerned this approach works, and it may also be the way to fix the messed-up dependency tree - see e.g. https://github.com/com-lihaoyi/Ammonite/issues/1241.

mpollmeier:
I reenabled the test - at least that looked like the right one...

Unfortunately it looks like my changes messed up the classpath for `test`. E.g. if I run `mill -i "amm[3.2.0].test"` it doesn't find `mainargs` any longer. I'm not really a mill expert - can you help me with that please?

lefou:
> I reenabled the test - at least that looked like the right one...
> 
> Unfortunately it looks like my changes messed up the classpath for `test`. E.g. if I run `mill -i "amm[3.2.0].test"` it doesn't find `mainargs` any longer. I'm not really a mill expert - can you help me with that please?

That's because `compileIvyDeps` no longer propagate to the `test` module (as it is no longer in `ivyDeps`). You need to add `mainargs` dependency to the dependencies of the `test` explicitly.

amm/src/test/scala/ammonite/main/MainTests.scala
```diff
@@ -292,11 +292,5 @@ class MainTests extends TestSuite{
 }
 
 object MainTests extends MainTests {
-  val isScala2 = ammonite.compiler.CompilerBuilder.scalaVersion.startsWith("2.")
-  override def tests =
-    if (isScala2) super.tests
-    else
-      Tests {
-        test("Disabled in Scala 3") - "Disabled in Scala 3"
-      }
```
lolgab, line 301: You can change `MainTests` to be an `object` directly now instead and delete this `object MainTests extends MainTests`

mpollmeier: thanks - done

mpollmeier:
@lefou thanks, yes that fixed the test classpath issue. 

I had to disable one specific case from the `MainTests` again - something related to a type in `mainargs.ParserForMethods`. I didn't dig in too deep though, maybe someone else knows what's going on. I pasted the scalac output into [8e519af](https://github.com/com-lihaoyi/Ammonite/pull/1298/commits/8e519aff4fd482584c424ad34d1c4e8250e50f43).

Either way - @main entrypoints work now for Scala 3, (almost all) tests are reenabled, so I'd say we're much better off than before.

mpollmeier:
I tried to refactor the build a bit (remove the duplication in `runIvyDeps`), but didn't see a great way forward... in the end, we do want to be explicit that we're fiddling around here...

lolgab:
This thing is hard :(
For example, now the `mainargs_2.13` is set as `provided` in the pom and there is no `mainargs_3` in it.
This means that people using the Ammonite jar will not get the mainargs dependency and it will fail with a ClassNotFound.
I tried to fix this by manually editing the dependencies entering the pom by overriding `def publishXmlDeps: Task[Agg[Dependency]]` but it gets messier and messier. Probably a better long term strategy would be to help finish the [port of fastparse to Scala 3](https://github.com/com-lihaoyi/fastparse/pull/262) to we can switch to Scala 3 completely..

lolgab:
@mpollmeier I'm trying another approach. Instead of adding the `_3` deps to `runIvyDeps` and the `_2.13` deps to `compileIvyDeps` which create a weird pom, I'm adding the `_3` deps to the pom but then I tweak the method that resolves the dependencies **for the compileClasspath only** to use the `_2.13` deps.
Also I did some cleanups in the build and now it seems we don't have any duplicated deps like `fansi_3` + `fansi_2.13`
https://github.com/com-lihaoyi/Ammonite/pull/1301
If it end up working fine, then we can rebase this PR on top and enjoy main methods on Scala 3.
About upickle it still doesn't work. I think there is some binary incompatibility between Scala 3 and Scala 2.13 which I'm not sure it comes from the upickle codebase or some difference between the two compilers.

lefou:
> About upickle it still doesn't work. I think there is some binary incompatibility between Scala 3 and Scala 2.13 which I'm not sure it comes from the upickle codebase or some difference between the two compilers.

There is a slight change, that the Scala version specific code in upickle is not interchangeable in a binary compatible sense. At least, there is no tooling ensures it.

lolgab:
> > About upickle it still doesn't work. I think there is some binary incompatibility between Scala 3 and Scala 2.13 which I'm not sure it comes from the upickle codebase or some difference between the two compilers.
> 
> There is a slight chance, that the Scala version specific code in upickle is not interchangeable in a binary compatible sense. At least, there is no tooling ensuring it.

It might be but it's not only that. Swapping `fansi_2.13` with `fansi_3` (the library doesn't have any version specific code) fails with:
```
X ammonite.terminal.HistoryTests.noHistory 23ms 
  java.lang.NoClassDefFoundError: scala/runtime/function/JProcedure1
    fansi.Attrs$.apply(Fansi.scala:580)
    fansi.Attrs$.<clinit>(Fansi.scala:514)
    ammonite.terminal.HistoryTests$.$anonfun$tests$1(HistoryTests.scala:18)
```
I tried to bump the fansi version and now the test pass just fine :) .
Probably some Scala 3 compatibility issue was fixed in the Scala 2 compiler going from 2.13.4 -> 2.13.8.

lolgab:
I think I sorted it out 🎉
@mpollmeier @lefou would you mind reviewing #1301 which should solve the issue?
I think the solution can be extended to uPickle after we release a new version based on Scala 2.13.8+ (current version was compiled with 2.13.4 which created binary incompatibilities in fansi, so I expect it can be the same issue we have in upickle.

mpollmeier:
🎉 nice one, yes will do

PR #1304, opened 2022-10-18
Use_3 for scala-collection-compat

lolgab:
scala-collection-compat suffers from the same issue seen in the other deps
In sbt we see:
```
Modules were resolved with conflicting cross-version suffixes in ProjectRef(uri("file:/..../"), "core"):
[error]    org.scala-lang.modules:scala-collection-compat _3, _2.13
```


lefou (approved):
Looks good to me. 

Should we add some auto-register to the use_3_deps list? Who knows, which dep comes next?

lolgab:
> Looks good to me.
> 
> Should we add some auto-register to the use_3_deps list? Who knows, which dep comes next?

I hope we can migrate to Scala 3 before this becomes a real problem! :)

PR #1306, opened 2022-10-19
Exclude scala-collection-compat_2.13 in Scala 3

lolgab:


build.sc
```diff
@@ -137,7 +137,7 @@ object Deps {
     override def dep(scalaVersion: String) =  ivy"com.lihaoyi::pprint:0.7.3"
   }
   object scalaCollectionCompat extends Use3Dep {
-    override def dep(scalaVersion: String) = ivy"org.scala-lang.modules::scala-collection-compat:2.6.0"
+    override def dep(scalaVersion: String) = ivy"org.scala-lang.modules::scala-collection-compat:2.8.1"
```
lolgab, line 140: This version was already used in ammonite as a transitive dependency.

build.sc
```diff
@@ -504,11 +504,15 @@ object amm extends Cross[MainModule](fullCrossScalaVersions:_*){
...
+           // we remove transitive _2.13 dependencies from Scala 3 and
            // then we add it back with _3
            .exclude("com.lihaoyi" -> s"sourcecode_${scalaBinaryVersion(scalaVersion())}")
+           .exclude("org.scala-lang.modules" -> s"scala-collection-compat_${scalaBinaryVersion(scalaVersion())}")
       else dep
-    ) ++ Agg(Deps.sourcecode.use_3(crossScalaVersion))
+    ) ++ Agg(
+      Deps.sourcecode.use_3(crossScalaVersion),
+      Deps.scalaCollectionCompat.use_3(crossScalaVersion)
+    )
```
lefou, line 515: I wonder if this can be also/better solved with `mapDependencies` in general?

lefou (approved):
Looks good to me, I guess. But I think, maybe we can ditch all the `Use3Deps` cruft and instead implement `mapDependencies` so that it replaces all known/wanted `_2.13` artifacts with `_3`. Nothing urgent, but if it turns out we need to tweak even more, it might result in a lighter and easier to understand design.

PR #1316, opened 2023-01-12
Ensure and refine lowest supported Scala versions

lefou:
I added the lowest supported Scala versions for Scala 2 and Scala 3 to the CI matrix to ensure we not accidentally use API which was introduced later. (Specifically, I wanted to validate that #1315 still works with Scala 3.0.0). That way I ruled some older versions out. Lowest supported Scala 3 version is now 3.1.1.


lefou:
Looks like the lowest Scala 3.x version Ammonite supports is `3.1.1`.

lefou:
After rebasing this to latest `main`, it looks like even more Scala 3 versions are no longer supported. It could be just because we use a too new Scala compiler to produce artifacts meant to be used by older Scala versions and some TASTy incompatibility issues. 

To be honest, I lost the interest in pushing this forward. It looks to me, nobody is interested in keeping support for older Scala versions or in proper integration testing to ensure we really support what we claim. But to be fair, we don't even claim anywhere any support for whatever version, so it might just be me and my own expectations. I don't use Ammonite myself and Mill, which I really care about, no longer depends on it, so I think best is to just let it go and relax. Feel free to take this work over.

PR #1320, opened 2023-02-06
Update upickle to 3.0.0-M2

lefou:
This is a binary incompatible change! The next release version after merge is probably `3.0.0`.

I dropped support for Scala 3.0.x due to the fact, that upickle_3 3.0.0-M2 is build against Scala 3.1.

Ammonite depends also on geny but never made that dependency explicit. I changed that and added a direct dependency where appropriate. With the bump of upickle, we implicitly also bump geny from 0.7.0 to 1.0.0. As a consequence, I also had to update os-lib from 0.8.0 to 0.9.0 as it is also depending on geny, and only version 0.9.0 is compatible to geny 1.0.0.

### DIFF
--- a/build.sc
+++ b/build.sc
@@ -79,6 +79,7 @@ object Deps {
   val cask = ivy"com.lihaoyi::cask:0.6.0"
   val coursierInterface = ivy"io.get-coursier:interface:1.0.11"
   val fastparse = ivy"com.lihaoyi::fastparse:$fastparseVersion"
+  val geny = ivy"com.lihaoyi::geny:1.0.0"
   val javaparserCore = ivy"com.github.javaparser:javaparser-core:3.2.5"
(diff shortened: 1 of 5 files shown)


lihaoyi:
@lefou from the error in the last run, it seems that the new uPickle is pulling in a version of Geny that is incompatible with the version of Geny that os-lib is using.

 Not sure why it would only fail on Scala3, but maybe we can try bumping the os-lib version?

lihaoyi:
Seems like bumping os-lib from 0.8.0 to 0.9.0 should be enough to ensure that everyone is using geny 1.0.0

lefou:
> Seems like bumping os-lib from 0.8.0 to 0.9.0 should be enough to ensure that everyone is using geny 1.0.0

Makes totally sense. Thanks for reviewing.

PR #1334, opened 2023-04-06
Allow to write REPL byte code to a directory

alexarchambault:
This adds an `--output-directory` option equivalent to `-Yrepl-outdir` in the default Scala REPL. The latter is [used by the Spark REPL](https://github.com/apache/spark/blob/156a12ec0abba8362658a58e00179a0b80f663f2/repl/src/main/scala-2.12/org/apache/spark/repl/Main.scala#L70) in particular, and I'm planning to use that new option in Spark-related developments (in [ammonite-spark](https://github.com/alexarchambault/ammonite-spark), itself used by [Almond](https://github.com/almond-sh/almond)).

This option basically writes in the passed directory the byte code generated by compiling the user code.

### DIFF
--- a/amm/compiler/interface/src/main/scala/ammonite/compiler/iface/CompilerLifecycleManager.scala
+++ b/amm/compiler/interface/src/main/scala/ammonite/compiler/iface/CompilerLifecycleManager.scala
@@ -13,6 +13,8 @@ abstract class CompilerLifecycleManager {
 
   def scalaVersion: String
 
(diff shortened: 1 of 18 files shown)


alexarchambault:
@lefou @lihaoyi I'll probably merge this tomorrow, alongside https://github.com/com-lihaoyi/Ammonite/pull/1333, unless one of you want to ponder things a bit.

lihaoyi:
@alexarchambault go for it

PR #1339, opened 2023-04-21
doc: Fix incorrect mention of Scala 2.12 being latest

TJC:
So, I noticed that the documentation contains a strange phrase, saying that the latest version of Scala is 2.12.
This is a few versions out of date! I've fixed it here.

However I think it's also worth you taking a look at the older versions listed, as currently it's just showing up a 2.12 version, and I think you have a 2.13 version around too?


alexarchambault (approved):
Nice catch, thanks!

PR #1340, opened 2023-04-21
Accept files paths and more refined class paths in 'import $cp'

alexarchambault:
See the added documentation. This makes the following work, while not breaking today's `import $cp` syntax:
```scala
@ import $cp.`/foo/a/b.jar`
@ import $cp.`/foo/a/b.jar;/foo/a/c.jar;/foo/a/d.jar`
@ import $cp.`/foo/a/*`
@ import $cp.`../foo/a/*`
@ import $cp.`${the.foo.dir}/a/*`
```

That syntax makes it easier to add elements to the class path using `import $cp`, when one gets these paths via another tool: one doesn't have to convert it to `import $cp.^.^.^.foo.a.b`. And it also:
- substitutes Java properties
- accepts multiple elements separated by `;` (Linux / macOS) or `:` (Windows)
- accepts some blobs to add all JARs in a directory

This relies on [a library](https://github.com/coursier/class-path-util) recently added in then spin off from coursier. The `--extra-jar` option of `cs launch` uses it too. I'm planning to propose to have Scala CLI accept it at various places too.

### DIFF
--- a/amm/repl/src/test/scala/ammonite/DualTestRepl.scala
+++ b/amm/repl/src/test/scala/ammonite/DualTestRepl.scala
@@ -22,6 +22,9 @@ class DualTestRepl { dual =>
   )
 
   def scalaVersion = compilerBuilder.scalaVersion
+  lazy val scalaBinaryVersion =
+    if (scalaVersion.startsWith("2.")) scalaVersion.split('.').take(2).mkString(".")
+    else scalaVersion.takeWhile(_ != '.')
   def scala2 = scalaVersion.startsWith("2.")
   def scala2_12 = scalaVersion.startsWith("2.12.")
 
--- a/amm/repl/src/test/scala/ammonite/session/BuiltinTests.scala
+++ b/amm/repl/src/test/scala/ammonite/session/BuiltinTests.scala
@@ -3,6 +3,7 @@ package ammonite.session
 import ammonite.{DualTestRepl, TestUtils}
 import utest._
 
+import scala.collection.JavaConverters._
 import scala.collection.{immutable => imm}
 object BuiltinTests extends TestSuite{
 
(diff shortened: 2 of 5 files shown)


alexarchambault:
(Didn't mean to push that to the upstream repo… 😬)

lefou:
> @ import $cp.`/foo/a/b.jar;/foo/a/c.jar;/foo/a/d.jar`

>     * accepts multiple elements separated by `;` (Linux / macOS) or `:` (Windows)

Does that mean, scripts get platform specific, or does it mean that you accept both characters (`;` and `:`) as separator, on any platform?

>     * accepts some blobs to add all JARs in a directory

Do you mean "globs", e.g. the `*` char? Please specify explicitly how they expand, as the expansion rules of globs in the `-classpath` option of the JVM are a bit strange and unexpected, IMHO.

> This relies on [a library](https://github.com/coursier/class-path-util) recently added in then spin off from coursier. The `--extra-jar` option of `cs launch` uses it too. I'm planning to propose to have Scala CLI accept it at various places too.

Please add a license and some version policy to that library, before merge.

alexarchambault:
> Does that mean, scripts get platform specific, or does it mean that you accept both characters (; and :) as separator, on any platform?

This gets platform-specific, yes. On Windows, we can't use `:` as a path-separator, as it's in every absolute paths (those that looks like `C:\…` at least). We might accept `;` on Linux and macOS… although, I haven't seen that being done elsewhere.

I guess the recommendation for OS-independent scripts would be:
- don't rely on the class path separator, use a separate `import $cp` per class path entry
- use `/` as a separator, as it's also accepted on Windows
- absolute paths can't be really OS-independent, so users should use a Java property if they want to refer to an external directory, like ``import $cp.`${spark.home}/jars/*` ``

Maybe I should put that in the doc 🤔

alexarchambault:
> Do you mean "globs", e.g. the `*` char?

Yes.

alexarchambault:
@lefou That should be better now

PR #1344, opened 2023-05-17
Assume import $cp to be a class path if it contains '${'

alexarchambault:


alexarchambault:
I'd like to add a non-regression test for that

PR #1345, opened 2023-05-17
Update class-path-util to 0.1.4

alexarchambault:


alexarchambault:
Handled from https://github.com/com-lihaoyi/Ammonite/pull/1344

PR #1348, opened 2023-05-29
Highlight MIT in LICENSE

akirataguchi115:
I also changed the copyright year from 2014 to 2015. This is because the year in copyright notice refers to the year of first publication as italicised here: https://opensource.stackexchange.com/questions/6389/how-do-the-years-specified-in-a-copyright-statement-work . According to the GitHub API this repository was not created until 2015: https://api.github.com/repos/com-lihaoyi/Ammonite :octocat:

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,10 +1,6 @@
-License
-=======
+MIT License
 
-
-The MIT License (MIT)
-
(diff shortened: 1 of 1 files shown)


alexarchambault (approved):
Thanks!

PR #1355, opened 2023-06-09
Use coursier-dependency to parse dependencies

alexarchambault:
It lives in [this repository](https://github.com/coursier/dependency). Scala CLI uses that parser too. It allows to add support for things like 'org:name:ver,intransitive' for example


alexarchambault:
Unless anyone opposes or wants to ponder things, I think I'll merge that tomorrow (in about 24h say…)

lefou:
What is the difference between coursier-dependency and coursier-dependency-interface?

alexarchambault:
coursier-dependency is the main / core library for dependency handling (data types, parsing, string interpolators, etc.). coursier-dependency-interface has helpers to convert the coursier-dependency data types to the ones of [coursier-interface](https://github.com/coursier/interface). It's briefly described [here](https://github.com/coursier/dependency/tree/4749550de4f09fc92f8c1a704d87832abff830bb#to-coursier-interface-dependency).

PR #1358, opened 2023-06-13
bsp4j: 2.0.0-M6 → 2.1.0-M5

fabianhjr-dealengine:
bsp4j:2.0.0-M6 depends on com.google.code.gson:gson:2.8.2 that has a known vulnerability:

- https://nvd.nist.gov/vuln/detail/CVE-2022-25647 (High: 7.5 CVSS)

bsp4j:2.1.0-M4 depends on com.google.code.gson:gson:2.9.1 that is not affected by CVE-2022-25647

### DIFF
--- a/amm/interp/src/main/scala/ammonite/interp/script/AmmoniteBuildServer.scala
+++ b/amm/interp/src/main/scala/ammonite/interp/script/AmmoniteBuildServer.scala
@@ -126,7 +126,6 @@ class AmmoniteBuildServer(
     clientOpt = Some(client)
   }
 
-
   def buildInitialize(params: InitializeBuildParams): CompletableFuture[InitializeBuildResult] =
     nonBlocking {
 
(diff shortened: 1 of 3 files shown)


amm/interp/src/main/scala/ammonite/interp/script/AmmoniteBuildServer.scala
```diff
@@ -126,6 +126,11 @@ class AmmoniteBuildServer(
     clientOpt = Some(client)
   }
 
+  // Stubs
+  override def buildTargetDependencyModules(dmp: DependencyModulesParams): CompletableFuture[DependencyModulesResult] = ???
```
alexarchambault, line 130: I'd rather add those to [`DummyBuildServerImplems`](https://github.com/com-lihaoyi/Ammonite/blob/e445eb40cee0b6548852fb521e48795c292f029f/amm/interp/src/main/scala/ammonite/interp/script/DummyBuildServerImplems.scala), and return empty results rather than throwing when possible.

alexarchambault (commented):
Thanks! See the added comment, but looks good otherwise.

amm/interp/src/main/scala/ammonite/interp/script/AmmoniteBuildServer.scala
fabianhjr-dealengine: Adjusting the stubs to use DummyBuildServerImplems

alexarchambault (approved):

PR #1359, opened 2023-06-19
Fix code completion breaking after wildcard import

vsevolodstep-db:
This PR fixes an issue when code completion breaks after a wildcard import (see https://github.com/com-lihaoyi/Ammonite/issues/1009 for more details). It's solved by removing ```package```  import item from imports string passed to completer.

To test this, I added a new `ReplTests` suite which instantiates a Repl and feeds it a user input string which may consist of a several commands. The existing unit tests (`AutocompleteTests` in particular) cannot be used to verify this fix, as they use `TestRepl` instead of `Repl` which duplicates code completion logics.


lihaoyi (approved):
Looks good to me. Will leave it up for a few days in case anyone else wants to review it, before merging

PR #1366, opened 2023-07-16
Enable warnings by default

alexarchambault:
Still a draft, as I'd like to add tests, and maybe put that behind a CLI option.

What do you think about enabling warnings by default?


lefou (approved):
Looks good to me!

PR #1375, opened 2023-09-13
Update Scala to 2.13.12

alexarchambault:


3XC1T3D:
Hi, will there be a 2.x release with the scala update? or do we have to wait till 3.x ist out?

lefou:
I guess, we can tag/release a `3.0.0-M1`.

vbabenkoru:
@lefou Does that mean that one would have to use a pre-release Ammonite for the latest 2.13 Scala version?

PR #1383, opened 2023-11-13
Scala 3 without Scala 2.13 toolchain

lolgab:
## Motivation

Until now Ammonite for Scala 3 was compiled using a mix of the Scala 2.13 and the Scala 3 compiler.
This was great since it allowed us to have Ammonite on Scala 3 before `fastparse`, `mainargs` and other libraries were even supported on Scala 3, or they were not on par with their Scala 2.13 counterpart.
Now that the ecosystem has caught up with Scala 3, we can avoid using Scala 2.13 and do everything with Scala 3.
This has the benefit that now all the classpath contains only Scala 3 libraries and we can finally invoke Scala 3 macros in the included libraries like `upickle`.
Also, the build gets much simpler since we don't have to do tricks to swap dependencies at runtime and manually remove transitive dependencies.
The only dependency that still is on Scala 2.13 is scalameta `common`, for which we still force the Scala 3 version of its `sourcecode` dependency to avoid double dependencies at runtime.

## uPickle macros on Scala 3

upickle

## Breaking changes

- Drop support for Scala 3.1
- Drop support for Scala 2.13.0
- Scala 3 artifacts don't have the `-cross` naming convention anymore but are just like Scala 2 artifacts except for the `_3` or `_3.x.y` suffix
- `ammonite.util.Bind` now depends on [typename](https://github.com/tpolecat/typename) instead of scala-reflect (which doesn't have a Scala 3 counterpart
- `scala-java8-compat` updated to `1.0.2`
- `mainargs` updated to `0.5.4`
- `requests` updated to `0.8.0`
- `interp.api.APIHolder.value` is now `final`
- `typeOf` removed from Scala 3 API (was not supported already)
- Vararg command line arguments now use `mainargs.Leftover`

Pull request: https://github.com/com-lihaoyi/Ammonite/pull/1383


amm/runtime/src/main/scala/ammonite/runtime/Storage.scala
```diff
@@ -188,23 +188,23 @@ object Storage{
...
+    val cacheDir = dir/"cache"/ammonite.Constants.version
+    val compileCacheDir = cacheDir/"compile"
     val classFilesOrder = "classFilesOrder.json"
     val ivyCacheFile = cacheDir/"ivycache.json"
     val coursierFetchCacheDir = cacheDir/"coursier-fetch-cache"
     val metadataFile = "metadata.json"
     val sessionFile  = dir/"session"
 
-    def getSessionId() = {
+    def getSessionId: Long = {
```
lefou, line 199: The now removed `()` was probably there to indicate the side effect of reading the file. I'm neutral to this change, though.

.mill-version
```diff
@@ -1 +1 @@
-0.10.11
+0.10.12
```
lefou, line 1: We now have a `0.11.13`

lefou (commented):
I'm not that intimate with the Ammonite code base, so I just  bikeshedded a bit below. Otherwise, this looks reasonable to me.

amm/runtime/src/main/scala/ammonite/runtime/Storage.scala
lolgab: This method overrides an abstract method that is defined without the `()` argument lists.
In Scala 3 overrides need to have the same signature, so I preferred to maintain the interface one.

lefou (approved):

lolgab:
@alexarchambault @lihaoyi Can I have another review?

.github/workflows/actions.yml
```diff
@@ -24,12 +24,13 @@ jobs:
         with:
           java-version: ${{ matrix.java-version }}
       - run: ./mill -i unitTest "${{ matrix.scala-version }}"
+      - run: ./mill -i unitTest "${{ matrix.scala-version }}"
```
lefou, line 27: Why this duplicate?

lihaoyi (approved):

lihaoyi:
Looks reasonable to me. If the test suite passes that gives us confidence that nothing broke

lihaoyi:
-350 lines of code is also very nice, given how gnarly a lot of the deleted code is

lefou (approved):

PR #1394, opened 2024-02-15
Support for Scala 3.4.0

Gedochao:


Gedochao:
```
Loading Scala 3 binary from /Users/pchabelski/IdeaProjects/Ammonite2/out/amm/interp/api/3.4.0/compile.dest/classes/ammonite/$ivy/$.class. It should have been loaded from `.tasty` file. This `.tasty` file is missing. Try cleaning the project to fix this issue.
```
This seems `3.4.0`-specific and pops up all over the place... 😕 
Replicable for example with `ammonite.integration.BasicTests.testIvySnapshotNoCache` (and some other tests)

Gedochao:
After some investigation, the failed tests seem to be caused by this:
https://github.com/lampepfl/dotty/issues/19702
As it seems identifiers containing `$` will not be allowed long term, it may require fixes on `Ammonite`'s side.

Gedochao:
I don't know the codebase enough to attempt this, but I suppose we'd need to refactor Ammonite to get rid of identifiers containing the `$` special character.
Otherwise, https://github.com/lampepfl/dotty/issues/19702 will keep haunting us.
@lefou @lolgab can you take a look and check if I'm maybe missing some simple workaround?

lihaoyi:
I think the basic issue here is we allow `import $ivy.blah`, which desugars at runtime into `import $ivy.$` and some other stuff. The "other stuff" is what's important, the `import $ivy.$` is just a placeholder stub that does nothing

I think it should be possible to desugar `import $ivy.blah` into just whitespace ` `. The `import $ivy.$` (and similar) stubs don't actually do anything, so removing them should be ok

Gedochao:
Corresponding Ammonite issue:
- https://github.com/com-lihaoyi/Ammonite/issues/1395

lefou:
Yeah, I think splitting this up into a Scala 3.3.x update and an separate 3.4.x is the best way to make progress.

Ichoran:
It's a little hard to tell what is still needed to get this working.  All of my code depends on 3.4 features, and Jupyter needs Almond, and Almond depends on Ammonite, so...basically I can't use Jupyter notebooks until this gets fixed.  I _also_ depend on Java 21 (for Loom threads), and I notice that this PR stops testing even Java 17.  Is this because the matrix is too big?

Anyway, I have time to help resolve the issue myself, but I can't tell right now what still needs doing.  I can't see failing logs with the test, and I don't really want to dive in without knowing what I'm getting myself into.

lihaoyi:
It seems the latest scala 3 added bakc support for `$`. Maybe it means we dont need any code changes and can just skip the problematic 3.4.x versions?

Ichoran:
That might be the thing to do for now, but we can't rely on anything with a _trailing_ `$` for the reason Nicolas mentions in the ad-hoc patch (which is supposed to be temporary): https://github.com/scala/scala3/pull/19705

But I think you're right that it's in.  3.4.2, I think, so maybe the thing to do is just go with that and skip 3.4.0 and 3.4.1.

PR #1503, opened 2024-06-07
Added 3.4.2 support in a fairly formulaic fashion

Ichoran:
Added 3.4.2 to targets in a fairly straightforward, formulaic fashion.  3.4.0 and 3.4.1, for good reason, fail to parse `$` as a class file, but a [temporary ad-hoc workaround was added in 3.4.2](https://github.com/scala/scala3/pull/19705), so this needn't block 3.4 support for now.  Reliance on `object $` still needs to be [fixed in Ammonite](https://github.com/com-lihaoyi/Ammonite/issues/1395).

Several internals changes required fixes also; I used the most obvious solution (e.g. if Y extends X and Y is now private, use X instead), but someone who knows more might want to read it over.

Tests pass on my machine for 3.3 and 3.4.2 (under Java 21).

If accepted, this patch would supersede https://github.com/com-lihaoyi/Ammonite/pull/1394


Ichoran:
If it works, this should complete https://github.com/com-lihaoyi/Ammonite/issues/1425

lihaoyi:
@Ichoran how much of the stuff in the `3.4.2+/` folder really needs to be duplicated? I wonder if it actually involves a widespread change to the relevant code, or whether we just need to duplicate a few helper functions and the rest of the code can remain shared for easier maintainability

Ichoran:
@lihaoyi - Yeah, I wasn't entirely happy with the code duplication, but that's what was already done for 3.0 vs 3.3.2+ (i.e. if there is a change in a file, duplicate the file in subversion-specific folders and remove it from scala-3), so I just continued the pattern.  The changes are not extensive, and could be refactored out, but that should be done for 3.0 vs 3.3.2+ also if so.

I'm happy to take a look at that, but I'm not sure that should block 3.4 support given that it's already on the late side.

I can only plead, for this PR: I may have dug a hole, but it isn't a _new_ hole, I'm just digging the existing one deeper.

lihaoyi:
sgtm

lihaoyi:
This is available as `//github.com/com-lihaoyi/Ammonite/releases/download/3.0.0-M2/3.4-3.0.0-M2-9-88291dd8`

PR #1505, opened 2024-06-14
bugfix: Use fullCrossScalaVersions for terminal and util

tgodzik:
These are used from REPL so, they can't just use any binary Scala 3 version and it seems the latest is picked by default.

Should fix https://github.com/com-lihaoyi/Ammonite/issues/1504


tgodzik:
Looks like it doesn't compile now, the other solution is a fix in Mill, but that's harder I think

PR #1507, opened 2024-06-14
Avoid publishing binCrossScalaVersions cross-built modules for 3.4.2

lihaoyi:
We only want to publish for 3.3.x, because it is compatible with all future versions, whereas 3.4.x is not compatible with 3.3.x and causes conflicts when they both have the same `_3` suffix

Fixes https://github.com/com-lihaoyi/Ammonite/issues/1504


tgodzik:
Thanks!

PR #1580, opened 2024-11-01
ci: harmonize amm-template.sh script with mill script

### LINKED ISSUE #1579: Harmonize AMM_DOWNLOAD_PATH from amm-template.sh with MILL_DOWNLOAD_PATH from mill script
The suggestion is to refactor the download path used in https://github.com/com-lihaoyi/Ammonite/blob/main/amm-template.sh#L17

```shell
AMM_DOWNLOAD_PATH="$HOME/.ammonite/download"
```

with what's in the `mill` script in https://github.com/com-lihaoyi/Ammonite/blob/main/mill#L25-L29:

```shell
if [ "x${XDG_CACHE_HOME}" != "x" ] ; then
  MILL_DOWNLOAD_PATH="${XDG_CACHE_HOME}/mill/download"
else
  MILL_DOWNLOAD_PATH="${HOME}/.cache/mill/download"
fi
```

cmontemuino:
PROBLEM: `amm-template.sh` is downloading the ammnite executable into `${HOME}/.ammnite/download`. It is not necessarily wrong, but does not follow the same pattern as in `mill` script. Having an harmonized way to keep downloaded data is always preferred.

SOLUTION: adopt the mechanism from `mill` script to download files, which also takes into account the existence of `${XDG_CACHE_HOME}` variable.

closes #1579


lefou (commented):
Looks good to me.

PR #1616, opened 2025-02-16
Use JDK 11 in CI jobs that were still on JDK 8

alexarchambault:
Seems upcoming Mill versions require JDK >= 11, see the CI failures of https://github.com/com-lihaoyi/Ammonite/pull/1608


alexarchambault:
Don't know if these were left on JDK 8 on purpose 🤔

PR #1631, opened 2025-03-27
3.7.0-RC1 ?

Quafadas:


Quafadas:
Umm... I opened this because I was curious, and a desire to help, if I can. what I realised as I began to look into it, was that I was in waaaay over my head. I believe it needs more than a superficial understanding of compiler changes for example. 

I'm going to close the PR, as I don't think it is going to save anyone time. I apologise for the confusion as I should have made that step earlier.

